PR #10, opened 2018-02-12
RethinkDB, TagView and APIView updates.

lemonsaurus:
- RethinkDB integration seperated out into a class of its own.
- APIView updated with API key validation and generation methods
- Created a TagView which is supposed to be used by the bot to submit or retrieve tags from the database. This view was created mostly to provide an example of how to use the new database integration.

### DIFF
--- a/pysite/base_route.py
+++ b/pysite/base_route.py
@@ -35,9 +35,9 @@ class APIView(RouteView):
         """ Placeholder! """
         return api_key == os.environ("API_KEY")
 
-    def generate_api_key(n: int):
+    def generate_api_key(self):
         """ Generate a random string of n characters. """
-        pool = random.choices(string.ascii_letters + string.digits, k=n)
+        pool = random.choices(string.ascii_letters + string.digits, k=32)
         return "".join(pool)
 
     def error(self, error_code: ErrorCodes):
--- a/pysite/base_route.py
+++ b/pysite/base_route.py
@@ -41,6 +41,7 @@ class APIView(RouteView):
         return "".join(pool)
 
     def error(self, error_code: ErrorCodes):
+
         data = {
             "error_code": error_code.value,
             "error_message": "Unknown error"
@@ -53,7 +54,12 @@ class APIView(RouteView):
             http_code = 404
         elif error_code is ErrorCodes.unauthorized:
             data["error_message"] = "Unauthorized"
-            http_code = 403
+            http_code = 401
+        elif error_code is ErrorCodes.invalid_api_key:
+            data["error_message"] = "Invalid API-key"
+            http_code = 401
+        elif error_code is ErrorCodes.missing_parameters:
+            data["error_message"] = "Not all required parameters were provided"
 
         response = jsonify(data)
         response.status_code = http_code
--- a/pysite/base_route.py
+++ b/pysite/base_route.py
@@ -1,4 +1,8 @@
 # coding=utf-8
+import os
+import random
+import string
+
 from flask import Blueprint, jsonify, render_template
 from flask.views import MethodView
 
@@ -27,6 +31,15 @@ class RouteView(BaseView):
 
 
 class APIView(RouteView):
+    def validate_key(self, api_key: str):
+        """ Placeholder! """
+        return api_key == os.environ("API_KEY")
+
+    def generate_api_key(n: int):
+        """ Generate a random string of n characters. """
+        pool = random.choices(string.ascii_letters + string.digits, k=n)
+        return "".join(pool)
+
     def error(self, error_code: ErrorCodes):
         data = {
             "error_code": error_code.value,
--- a/pysite/constants.py
+++ b/pysite/constants.py
@@ -1,5 +1,4 @@
 # coding=utf-8
-__author__ = "Careth Goles"
 
 from enum import IntEnum
 
--- a/pysite/constants.py
+++ b/pysite/constants.py
@@ -1,5 +1,5 @@
 # coding=utf-8
-__author__ = "Gareth Coles"
+__author__ = "Careth Goles"
 
 from enum import IntEnum
 
--- a/pysite/constants.py
+++ b/pysite/constants.py
@@ -6,3 +6,5 @@ from enum import IntEnum
 class ErrorCodes(IntEnum):
     unknown_route = 0
     unauthorized = 1
+    invalid_api_key = 2
+    missing_parameters = 3
--- a/pysite/database.py
+++ b/pysite/database.py
@@ -1,5 +1,4 @@
 # coding=utf-8
-__author__ = 'Joreth Cales'
 
 import os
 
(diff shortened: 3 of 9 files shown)


pysite/base_route.py
```diff
@@ -27,6 +31,15 @@ def setup(cls: "RouteView", blueprint: Blueprint):
 
 
 class APIView(RouteView):
+    def validate_key(self, api_key: str):
+        """ Placeholder! """
+        return api_key == os.environ("API_KEY")
+
+    def generate_api_key(n: int):
```
gdude2002: `n` should probably be defaulted to something, or maybe not present at all depending on how things go

pysite/constants.py
```diff
@@ -1,5 +1,5 @@
 # coding=utf-8
-__author__ = "Gareth Coles"
+__author__ = "Careth Goles"
```
gdude2002: Can you not just remove the thing? I fixed my IDE.. I think..

pysite/database.py
```diff
@@ -0,0 +1,44 @@
+# coding=utf-8
+__author__ = 'Joreth Cales'
```
gdude2002: Cmon dude :P

pysite/database.py
```diff
@@ -0,0 +1,44 @@
...
+import os
+
+from flask import abort
+
+import rethinkdb
+
+
+class RethinkDB:
+
+    def __init__(self):
```
gdude2002: Can we have a `loop_type` argument defaulted to `"gevent"`, or maybe a constant or something? We're gonna need to `rethinkdb.set_loop_type()` with it.

pysite/route_manager.py
```diff
@@ -3,42 +3,48 @@
...
         )
+        self.db = RethinkDB()
         self.app.secret_key = os.environ.get("WEBPAGE_SECRET_KEY")
         self.app.config["SERVER_NAME"] = os.environ.get("SERVER_NAME", "localhost")
+        self.app.before_request(self.db.before_request)
+        self.app.teardown_request(self.db.teardown_request)
 
-        self.main_blueprint = Blueprint("main", __name__)
+        # Store the database in the Flask global context
+        g.db = self.db
```
gdude2002: If you could type-hint this with a comment, our IDEs would be thankful.

`# type: RethinkDB`

pysite/views/api/bot/tag.py
```diff
@@ -0,0 +1,71 @@
+# coding=utf-8
+__author__ = 'Ferret Moles'
```
gdude2002: pls

pysite/views/api/bot/tag.py
```diff
@@ -0,0 +1,71 @@
...
+
+class TagView(APIView):
+    path = '/tag'
+    name = 'tag'
+    table = 'tag'
+
+    def __init__(self):
+
+        # make sure the table exists
+        conn = g.db.get_connection()
```
gdude2002: Context manager?

pysite/views/api/bot/tag.py
```diff
@@ -0,0 +1,71 @@
...
+        try:
+            rethinkdb.db(g.db.database).table_create(self.table, {'primary_key': 'tag_name'}).run(conn)
+        except rethinkdb.RqlRuntimeError:
+            print(f'Table {self.table} exists')
+        conn.close()
+
+    def get(self):
+
+        rdb = rethinkdb.table(self.table)
+        tag_name = session.get('tag_name')
```
gdude2002: This uses server-side cookie-associated values, I don't think `session` is what you wanted at all.

pysite/views/api/bot/tag.py
```diff
@@ -0,0 +1,71 @@
...
+                data = list(data) if data else []
+        else:
+            data = {'errors': 'invalid api_key'}
+
+        return jsonify(data)
+
+    def post(self):
+
+        rdb = rethinkdb.table(self.table)
+        tag_name = session.get('tag_name')
```
gdude2002: This uses server-side cookie-associated values, I don't think `session` is what you wanted at all.

pysite/views/api/bot/tag.py
```diff
@@ -0,0 +1,71 @@
...
+
+        if self.validate_key(api_key):
+            if tag_name:
+                data = rdb.get(tag_name).run(g.db.conn)
+                data = dict(data) if data else {}
+            else:
+                data = rdb.pluck('tag_name').run(g.db.conn)
+                data = list(data) if data else []
+        else:
+            data = {'errors': 'invalid api_key'}
```
gdude2002: Check out `self.error()` and the `ErrorCodes` enum in `constants`

pysite/views/api/bot/tag.py
```diff
@@ -0,0 +1,71 @@
...
+                    'tag_category': tag_category
+                }).run(g.db.conn)
+                data = {'errors': None}
+
+            else:
+                required = {'tag_name': tag_name, 'tag_content': tag_content}
+                missing = [key for key, value in required.items() if not value]
+                s = 's' if len(missing) > 1 else ''
+                error = f"Missing {len(missing)} required parameter{s}: {', '.join(missing)}"
+                data = {'errors': error}
```
gdude2002: We may or may not want to expose this - needs discussion

gdude2002 (changes requested):

pysite/base_route.py
lemonsaurus: Sure, we can default it to 32 if that sounds reasonable.

lemonsaurus: I'm just gonna hardcode it to 32 for now. not convinced we'll ever need it to be anything else. and also it's missing a `self`, oops.

templates/navigation.html
```diff
@@ -10,9 +10,9 @@
                 {% if current_page == "index" %}
                     <li class="uk-active"><a href="/"><i class="uk-icon fas fa-home"></i> &nbsp;Home</a></li>
                 {% else %}
-                    <li><a href="/"><i class="uk-icon fas fa-home"></i> &nbsp;Home</a></li>
+                    <li><a href="https://pythondiscord.com/"><i class="uk-icon fas fa-home"></i> &nbsp;Home</a></li>
```
gdude2002: Why?

gdude2002 (changes requested):

pysite/views/api/bot/tag.py
lemonsaurus: brainfart. I'm gonna redo this to expect JSON

templates/navigation.html
lemonsaurus: I accidentally left this in, but it was added because if we use this navbar on any of the subdomains, they will have unexpected behavior.

gdude2002: I'm not sure why we'd use this navbar on other domains

gdude2002 (approved):

PR #12, opened 2018-02-13
WIP: request for comments on a logging implementation

Inveracity:
Essentially I want to have some more uniform logging, instead of using prints() everywhere.
I have no idea how to actually go about implementing this, so I've written up what I've got so far, so every one can take a look and rewrite it to be better.


pysite/route_manager.py
```diff
@@ -19,6 +20,12 @@ def __init__(self):
         self.app = Flask(
             __name__, template_folder=TEMPLATES_PATH, static_folder=STATIC_PATH, static_url_path="/static",
         )
+
+        # Store logger in the Flask global context
+        self.log = Logs(self.app)
+        with self.app.app_context():
+            g.log = self.log  # type: RethinkDB
```
jb3: Shouldn't the `# type: RethinkDB` here be something else?

jb3 (dismissed):

pysite/database.py
```diff
@@ -11,7 +11,7 @@ class RethinkDB:
 
     def __init__(self, loop_type: str = "gevent"):
         self.host = os.environ.get("RETHINKDB_HOST", "127.0.0.1")
-        self.port = os.environ.get("RETHINKDB_PORT", "28016")
+        self.port = os.environ.get("RETHINKDB_PORT", "28015")
```
lemonsaurus, line 17: 28016 actually was there for a reason. it's because when we test the flask app locally, if we run rethinkdb with standard settings, it will spin up its admin panel on 8080 which is also where flask wants to run. so, as a solution, I just did `rethinkdb --port-offset 1`, which makes it run on 8081 and also 28016.

If you know a better way to do this that's fine too.

pysite/logs.py
```diff
@@ -0,0 +1,29 @@
+import logging
+import os
+
+
+class Logs:
```
lemonsaurus, line 5: not a fan of plural class names. maybe Logger?

pysite/logs.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, app):
```
lemonsaurus, line 6: type annotations?

pysite/logs.py
```diff
@@ -0,0 +1,29 @@
...
+        ch = logging.StreamHandler()
+        ch.setLevel(logging.DEBUG)
+
+        formatter = logging.Formatter('[%(asctime)s] [%(levelname)s] %(message)s')
+
+        ch.setFormatter(formatter)
+        self.log.addHandler(ch)
+        self.log.debug('info message')
+
+    def debug(self, message):
```
lemonsaurus, line 25: type annotation

pysite/logs.py
```diff
@@ -0,0 +1,29 @@
...
+        self.log.debug(message)
+
+    def info(self, message):
```
lemonsaurus, line 28: aaaand type annotation

pysite/route_manager.py
```diff
@@ -19,6 +20,12 @@ def __init__(self):
         self.app = Flask(
             __name__, template_folder=TEMPLATES_PATH, static_folder=STATIC_PATH, static_url_path="/static",
         )
+
+        # Store logger in the Flask global context
+        self.log = Logs(self.app)
+        with self.app.app_context():
+            g.log = self.log  # type: logger
```
lemonsaurus, line 27: this comment annotation is invalid. should be `# type: Logs` currently, or `# type: Logger` / `# type: SiteLogger` if you change the class name to that.

lemonsaurus (changes requested):
This seems like a reasonable approach. Most of my feedback is very minor, but we do need to figure out the thing with the rethinkDB port before we merge.

pysite/logs.py
lemonsaurus: or even SiteLogger?

pysite/database.py
gdude2002, line 17: This trivial change doesn't belong in this PR

pysite/logs.py
gdude2002, line 5: Do we really need this class? I'd have just used the `logging` module directly, I don't see anything we get from this class that the module doesn't already provide

pysite/route_manager.py
```diff
@@ -31,25 +38,27 @@ def __init__(self):
...
             self.app.register_blueprint(self.sub_blueprint)
-            print("")
+            self.log.debug("")
 
     def run(self):
         self.app.run(
-            port=int(os.environ.get("WEBPAGE_PORT", 8080)), debug="FLASK_DEBUG" in os.environ
+            host=os.environ.get("FLASK_HOST", "127.0.0.1"),
+            port=int(os.environ.get("WEBPAGE_PORT", 8080)),
+            debug="FLASK_DEBUG" in os.environ
```
gdude2002, line 61: I don't understand this change, does it even belong in this PR?

pysite/views/api/bot/tag.py
```diff
@@ -1,5 +1,4 @@
 # coding=utf-8
-
```
gdude2002, line 2: Formatting commits don't really belong in PRs, especially not trivial ones like this

gdude2002 (changes requested):

gdude2002:
I don't think this is the right way to go about this kind of logging at all, I don't think we need a special class for it, and the PR itself is full of trivial formatting or just generally unrelated changes.

I don't think this approach brings any advantage over traditional use of the `logging` module. For those reasons, my vote is to reject this PR.

lemonsaurus:
I think rejection is too harsh, the core idea is still fine, but you make a compelling argument that this may be a bit overengineered. Our logger is too simple to warrant an implementation this complex. It would probably have all the same functionality with a tenth of the lines if we'd just gone for a more traditional logger approach. 

Inver, do you think you could just do more or less the same thing without the use of a class, dial back the complexity?

pysite/route_manager.py
lemonsaurus: I actually don't fully understand this either. why wasn't host already here? and how did it work without host? and if it did work without it, why are we adding it?

pysite/views/api/bot/tag.py
lemonsaurus: I, on the other hand, think trivial formatting commits are okay.

pysite/logs.py
lemonsaurus: this is a very astute point. see the comments on the PR itself as well as #dev-code for more discussion on this topic.

pysite/database.py
Inveracity: the only reason I made this change is because the default port of rethinkdb is 28015, thought it easier to dev locally when using default settings in rethinkdb

pysite/views/api/bot/tag.py
Inveracity: this was an accident lol

pysite/route_manager.py
Inveracity: I needed to be able to override the default localhost value with 0.0.0.0 when running it in vagrant to be able to get to the site from my browser

pysite/logs.py
Inveracity: Unfortunately I have to override the flask logger due to the nature of gunicorn and flask, which is why I instantiate a new logger object from werkzeug. This lets the logging pass through to gunicorn and out to stdout.

Inveracity:
The point was not to review the code, but to discuss how to approach logging, I apologise if I didn't make that clear. I appreciate the feedback though!
I'll close it for now as I have no idea how to approach logging properly, but it would be nice with some logging, especially now when the site is broken and it's not giving any indication as to why.

pysite/logs.py
gdude2002: Does werkzeug not use the standard python logging stuff? If it does, all you should need is the name of the logger, then you can get a handle to it

pysite/route_manager.py
gdude2002: Ah, right, I getcha

PR #14, opened 2018-02-15
API view for syncing users from the bot #yumr

gdude2002:


pysite/views/api/bot/user.py
```diff
@@ -0,0 +1,45 @@
...
+from pysite.constants import ErrorCodes
+from pysite.decorators import valid_api_key
+
+REQUIRED_KEYS = [
+    "user_id",
+    "role"
+]
+
+
+class TagView(APIView, DBViewMixin):
```
lemonsaurus: why is this called TagView?

pysite/views/api/bot/user.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @valid_api_key
+    def post(self):
+        data = request.get_json()
+
+        if not isinstance(data, list):
+            data = [data]
+
+        for user in data:
+            if not all(k in user for k in REQUIRED_KEYS):
```
lemonsaurus, line 29: how about we repurpose the `valid_api_key` decorator to be `valid_api_request` instead, then we make `required_keys` (maybe rename to `required_params`?) be a class attribute, and then the decorator can check if they're present.

That way we only have to define the required params and all the validation is handled automatically.

lemonsaurus (changes requested):

pysite/views/api/bot/user.py
gdude2002: My plan is to have API key management via the database, so it doesn't make sense to supply sets of API keys here.

EDIT: I'm stupid, one second..

gdude2002: I think that might be a little complex - what do we do if a list is expected?

lemonsaurus (approved):

PR #16, opened 2018-02-17
Simple logging. 

lemonsaurus:
I don't know if this will show up in the docker logs like Inver wanted, so it might need testing in prod.

### DIFF
--- a/pysite/__init__.py
+++ b/pysite/__init__.py
@@ -5,28 +5,14 @@ import os
 
 # region Logging
 # Get the log level from environment
-log_level = os.environ.get("LOG_LEVEL", "info").lower()
+log_level = os.environ.get("LOG_LEVEL", "info").upper()
 formatter = logging.Formatter('[%(asctime)s] [%(levelname)s] %(message)s')
 
-if log_level == 'debug':
-    log_level = logging.DEBUG
-elif log_level == 'info':
-    log_level = logging.INFO
-elif log_level == 'error':
-    log_level = logging.ERROR
-elif log_level == 'critical':
-    log_level = logging.CRITICAL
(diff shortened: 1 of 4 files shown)


pysite/__init__.py
```diff
@@ -1 +1,32 @@
...
+import logging
+import os
+
+
+# region Logging
+# Get the log level from environment
+log_level = os.environ.get("LOG_LEVEL", "info").lower()
+formatter = logging.Formatter('[%(asctime)s] [%(levelname)s] %(message)s')
+
+if log_level == 'debug':
```
gdude2002: Could do this via hasattr/getattr?

gdude2002 (changes requested):
It might be worth using useful names for specific loggers in specific places, but for now I think this is mostly fine.

gdude2002 (dismissed):

pysite/route_manager.py
```diff
@@ -25,33 +26,34 @@ def __init__(self):
...
 
         for sub in self.subdomains:
             sub_blueprint = Blueprint(sub, __name__, subdomain=sub)
 
-            print(f"Loading Blueprint: {sub_blueprint.name}")
+            self.log.debug(f"Loading Blueprint: {sub_blueprint.name}")
             self.load_views(sub_blueprint, f"pysite/views/{sub}")
             self.app.register_blueprint(sub_blueprint)
-            print("")
+            self.log.debug("")
```
swedgwood: Why is there an empty print here? (Now self.log.debug)

swedgwood (commented):
Looking good to me!

swedgwood (approved):
👍

PR #20, opened 2018-02-18
adds beginnings of docs

Inveracity:


Zwork101:
This seems overly complicated, I see no reason to run it in a VM. In addition, the docs don't including setting up the database, and creating the environment variables.

Inveracity:
@Zwork101 it is entirely optional, there is no hard requirement to run this in a VM. Good points though, I've included notes on these things now.

gdude2002:
Would we not use [the wiki](https://github.com/discord-python/site/wiki) for this?

Inveracity:
@gdude2002 good point, didn't think about that

Inveracity:
Moved to wiki

PR #25, opened 2018-02-25
Help page #z62n

gdude2002:


pysite/views/main/help.py
```diff
@@ -0,0 +1,10 @@
+# coding=utf-8
+from pysite.base_route import RouteView
+
+
+class IndexView(RouteView):
```
lemonsaurus: probably shouldn't be IndexView

static/style.css
```diff
@@ -3,6 +3,13 @@
 }
 
 .uk-background-secondary .uk-navbar-dropdown-nav .uk-nav-header {
+    color: #666 !important;
```
lemonsaurus, line 6: I don't think you need to solve this by just slapping `!important` on all your CSS (that's considered bad form cause it can create _mystical problems_ further down the line). It should be enough to just craft a selector that has sufficient specificity, or make use of the fact that whichever rule is defined last overrules the other. In other words, if you import style.css after you've imported the UIKit stuff, a selector of equal specificity in style.css will take precendence.

that said, it's possible you had a really good reason for doing this, I haven't dug into the css, but if you did not please consider refactoring to not use `!important`

templates/main/base.html
```diff
@@ -15,12 +15,14 @@
 <body>
 {% include "main/navigation.html" %}
 {% if current_page != "index" %}
-<div class="uk-container uk-section">
-    <div class="uk-alert-danger" uk-alert>
-        <p>
-            Please note: this site is under construction. What you see now may be vastly different
-            from the final project state. Feel free to chat to us on Discord if you're curious!
-        </p>
+<div class="uk-container uk-section" style="padding-top: 10px; padding-bottom: 10px;">
```
lemonsaurus, line 18: why style in the tag? Can't we do this in style.css? maybe we can add some padding classes?

templates/main/help.html
```diff
@@ -0,0 +1,459 @@
+{% extends "main/base.html" %}
+{% block title %}Getting Help{% endblock %}
+{% block content %}
+    <div class="uk-section" style="padding-top: 20px; padding-bottom: 30px;">
```
lemonsaurus: another problem that could probably be solved with padding classes and without tag styling

lemonsaurus (commented):
I had some problems with this PR on Chromium, the paragraph icon transition css did not seem to work. not sure why. I also think the paragraph icon is probably too big. Other than that, this looks excellent. The article itself is perfect.

lemonsaurus (changes requested):
couple of minor changes required.

static/style.css
gdude2002: That doesn't work for everything because UIKit has a JS portion. That portion often applies styling directly to elements - which is a problem, as you can imagine.

I always try without `!important` first.

lemonsaurus (dismissed):

lemonsaurus:
was the view really supposed to be called IndexView?

app_test.py
```diff
@@ -27,6 +27,11 @@ def test_index(self):
         response = self.client.get('/', 'http://pytest.local')
         self.assertEqual(response.status_code, 200)
 
+    def test_help(self):
+        ''' Check the help path responds with 200 OK '''
+        response = self.client.get('/help', 'http://pytest.local/help')
```
Inveracity: it's not strictly necessary to include this: `'http://pytest.local/help'`, it defaults to the main url, I'm not sure why I included it myself in the top, but it's totally optional :D

lemonsaurus (approved):
Looks great! Let's merge it!

PR #26, opened 2018-02-25
Add Oauth2 login to site.

Zwork101:
The feature we've all been waiting for...

### DIFF
--- a/pysite/constants.py
+++ b/pysite/constants.py
@@ -1,6 +1,7 @@
 # coding=utf-8
 
 from enum import Enum, IntEnum
+from os import environ
 
 
 class ErrorCodes(IntEnum):
@@ -20,3 +21,7 @@ OWNER_ROLE = 267627879762755584
 ADMIN_ROLE = 267628507062992896
 MODERATOR_ROLE = 267629731250176001
 HELPER_ROLE = 267630620367257601
+
+CLIENT_ID = environ.get('CLIENT_ID', '')
+CLIENT_SECRET = environ.get('CLIENT_SECRET', '')
+SCOPE = 'identify email'
--- a/pysite/database.py
+++ b/pysite/database.py
@@ -207,7 +207,7 @@ class RethinkDB:
         """
 
         query = self.query(table_name).insert(
-            objects, durability=durability, return_changes=return_changes, conflict=conflict
+            *objects, durability=durability, return_changes=return_changes, conflict=conflict
         )
 
         if return_changes:
--- a/pysite/database.py
+++ b/pysite/database.py
@@ -207,7 +207,7 @@ class RethinkDB:
         """
 
         query = self.query(table_name).insert(
-            *objects, durability=durability, return_changes=return_changes, conflict=conflict
+            objects, durability=durability, return_changes=return_changes, conflict=conflict
         )
 
         if return_changes:
--- a/pysite/route_manager.py
+++ b/pysite/route_manager.py
@@ -8,7 +8,10 @@ from flask import Blueprint, Flask
 
 from flask_sockets import Sockets
 
+from flask_dance.contrib.discord import make_discord_blueprint
+
 from pysite.base_route import APIView, BaseView, ErrorView, RouteView
+from pysite.constants import CLIENT_ID, CLIENT_SECRET, SCOPE
 from pysite.database import RethinkDB
 from pysite.websockets import WS
 
@@ -38,6 +41,17 @@ class RouteManager:
         self.load_views(self.main_blueprint, "pysite/views/main")
         self.load_views(self.main_blueprint, "pysite/views/error_handlers")
         self.app.register_blueprint(self.main_blueprint)
+        self.log.debug("")
+
+        # Load the oauth blueprint
+        self.oauth_blueprint = make_discord_blueprint(
+            CLIENT_ID,
+            CLIENT_SECRET,
+            SCOPE,
+            '/'
+        )
+        self.log.debug(f"Loading Blueprint: {self.oauth_blueprint.name}")
+        self.app.register_blueprint(self.oauth_blueprint)
         self.log.debug("")
 
         # Load the subdomains
--- a/pysite/views/error_handlers/http_404.py
+++ b/pysite/views/error_handlers/http_404.py
@@ -7,8 +7,6 @@ from pysite.base_route import ErrorView
 class Error404View(ErrorView):
     name = "error_404"
     error_code = 404
-    description = "This is not the page you're looking for... (status 404)"
 
     def get(self, error: NotFound):
-        print("Called!")
-        return self.render('error.html', description=self.description), 404
+        return "Unable to find page", 404
--- a/pysite/views/error_handlers/http_404.py
+++ b/pysite/views/error_handlers/http_404.py
@@ -7,6 +7,8 @@ from pysite.base_route import ErrorView
 class Error404View(ErrorView):
     name = "error_404"
     error_code = 404
+    description = "This is not the page you're looking for... (status 404)"
 
     def get(self, error: NotFound):
-        return "replace me with a template, 404 not found", 404
+        print("Called!")
+        return self.render('error.html', description=self.description), 404
--- a/pysite/views/main/oauth.py
+++ b/pysite/views/main/oauth.py
@@ -1,0 +1,16 @@
+# coding=utf8
+from flask import redirect
+
+from pysite.base_route import RouteView
+
+
+class OathLogin(RouteView):
+    path = "/oauth"
+    name = "oauth"
+
+    def get(self):
+        client_id = ''
+
+        oauth_redirect_url = f"https://discordapp.com/api/oauth2/authorize?client_id={client_id}&" \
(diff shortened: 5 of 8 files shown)


pysite/constants.py
```diff
@@ -20,3 +21,7 @@ class ValidationTypes(Enum):
 ADMIN_ROLE = 267628507062992896
 MODERATOR_ROLE = 267629731250176001
 HELPER_ROLE = 267630620367257601
+
+CLIENT_ID = environ.get('CLIENT_ID', '')
+CLIENT_SECRET = environ.get('CLIENT_SECRET', '')
+SCOPE = 'identify email'
```
gdude2002, line 27: Too generic - can we prefix all of these with, eg, `DISCORD_OAUTH`? Env vars too.

pysite/views/error_handlers/http_404.py
```diff
@@ -9,4 +9,4 @@ class Error404View(ErrorView):
     error_code = 404
 
     def get(self, error: NotFound):
-        return "replace me with a template, 404 not found", 404
+        return "Unable to find page", 404
```
gdude2002, line 12: This change is out of scope for this PR. Also, should be templated. :P

gdude2002 (changes requested):
Huh, this is really all there is? We'll definitely have to expand on this.

pysite/views/error_handlers/http_404.py
Zwork101: Yeah, sorry. I had it changed, but then I needed to revert it back, and I forgot what it originally said.

pysite/constants.py
Zwork101: Alrighty

gdude2002 (dismissed):
Tentatively approving - I'm not 100% sure this is entirely ready for merge to master just yet as it's not functional without several views and database integration, but it won't hurt to add it either.

gdude2002:
Wow, that coverage change. That is.. wow.

gdude2002 (dismissed):
OK, so here's what needs doing before we consider this to be in a proper mergeable state.

* Views: Login, logout, authorized, error
    * My advice is to place these all under the same path, so eg `/auth/login`, `/auth/logout`, etc
* Session persistence: Sessions **must** be persisted in the database. They should have their own table, mapping session keys to the users table via user IDs
* Deny login to users that aren't on our server - there's an OAuth scope for this - and instruct them on how to rectify that problem
* Add login to the navbar, and display the logged-in user in place of the login link (it could be used to logout in that case)

pysite/route_manager.py
```diff
@@ -5,10 +5,14 @@
 import os
 
 from flask import Blueprint, Flask
+from flask_dance.contrib.discord import make_discord_blueprint
 from flask_sockets import Sockets
 
 from pysite.base_route import APIView, BaseView, ErrorView, RouteView
+from pysite.constants import DISCORD_OAUTH_ID, DISCORD_OAUTH_SCOPE, \
```
gdude2002: Please use parens for line splitting, not backslahes.

```py
from x import (
    a, b, c, d
)
```

pysite/oauth.py
```diff
@@ -0,0 +1,71 @@
...
+from flask_dance.contrib.discord import discord
+import requests
+
+from pysite.constants import DISCORD_API_ENDPOINT, SERVER_ID
+from pysite.mixins import DBMixin
+
+_MAIN_BACK = None  # type: pysite.database.RethinkDB
+
+
+class OauthBackend(BaseBackend, DBMixin):
```
gdude2002: I don't understand what this class does - can you comment it?

gdude2002:
Missing a feature:

> * Add login to the navbar, and display the logged-in user in place of the login link (it could be used to logout in that case)

I don't fully understand how session persistence is working, here - can you walk me through it?

gdude2002 (changes requested):
See comments

app_test.py
```diff
@@ -79,6 +80,21 @@ def test_ws_test(self):
         response = self.client.get('/ws_test')
         self.assertEqual(response.status_code, 200)
 
+    def test_oauth_login(self):
+        """check oauth redirects """
```
jerbob: These Docstrings aren't very consistent. I think we'll need something like:
```py
""" Check OAuth redirects """
```

app_test.py
```diff
@@ -79,6 +80,21 @@ def test_ws_test(self):
...
+        response = self.client.get(DISCORD_OAUTH_REDIRECT)
+        self.assertEqual(response.status_code, 302)
+
+    def test_oauth_logout(self):
+        """check oauth redirects """
```
jerbob: Can we have docstrings like this:
```py
""" Check OAuth redirects """
```

app_test.py
```diff
@@ -79,6 +80,21 @@ def test_ws_test(self):
...
+        response = self.client.get('/auth/logout')
+        self.assertEqual(response.status_code, 302)
+
+    def test_oauth_authorized(self):
+        """check oauth authorization"""
```
jerbob: Docstring consistency again :D 
```py
""" Check OAuth authorization"""
```

app_test.py
```diff
@@ -29,7 +30,7 @@ def create_app(self):
 class BaseEndpoints(SiteTest):
     """ test cases for the base endpoints """
```
jerbob: Docstring consistency pls
```py
""" Test cases for the base endpoints """
```

pysite/constants.py
```diff
@@ -21,6 +22,17 @@ class ValidationTypes(Enum):
...
+
+DISCORD_API_ENDPOINT = "https://discordapp.com/api"
+
+DISCORD_OAUTH_REDIRECT = "/auth/discord"
+DISCORD_OAUTH_AUTHORIZED = "/auth/discord/authorized"
+DISCORD_OAUTH_ID = environ.get('DISCORD_OAUTH_ID', '')
+DISCORD_OAUTH_SECRET = environ.get('DISCORD_OAUTH_SECRET', '')
+DISCORD_OAUTH_SCOPE = 'identify email guilds.join'
+
+
```
jerbob: Any reason why we have this extra space?

pysite/oauth.py
```diff
@@ -0,0 +1,85 @@
...
+    def get(self, *args, **kwargs):  # Not used
+        pass
+
+    def set(self, blueprint, token):
+        user = get_user()  # get the user's information, based on a local proxy
+        join_discord(token["access_token"], user["id"])  # Attempt to join a discord
+        sess_id = str(uuid5(uuid4(), self.key))  # Create a session ID, encrypted to protect against timing attacks
+        session["session_id"] = sess_id  # Set it
+        self.db.insert("oauth_data", {"id": sess_id, "access_token": token["access_token"], "snowflake": user['id'],
+                                      "refresh_token": token["refresh_token"], "expires_at": token["expires_at"]})
```
jerbob: Can we have something more readable here and for the insert below? Something like:
```py
self.db.insert("oauth_data", {
        "id": sess_id,
        "access_token": token["access_token"],
        "snowflake": user['id'],
        "refresh_token": token["refresh_token"],
        "expires_at": token["expires_at"]
    }
)
```

pysite/oauth.py
```diff
@@ -0,0 +1,85 @@
...
+        creds = _MAIN_BACK.db.get("oauth_data", id)
+        if creds:
+            return _MAIN_BACK.db.get("users", creds["snowflake"])
+    elif not _MAIN_BACK:
+        logging.warning("Failed to get user data: _MAIN_BACK not loaded")
+
+
+def logout():
+    sess_id = session.get("session_id")
+    if sess_id and _MAIN_BACK.db.get("oauth_data", sess_id):  # If user exists in db,
```
jerbob: Can this inline comment be above the line? It might be worth making the comments more descriptive too (aka `remove the user's session` rather than `remove them (at least, their session)`)

jerbob (changes requested):

pysite/database.py
```diff
@@ -103,6 +103,34 @@ def create_table(self, table_name: str, primary_key: str="id", durability: str="
             self.log.debug(f"Table created: '{table_name}'")
             return True
 
+    def delete(self, table_name: str, primary_key: Union[str, None] = None,
+               durability: str="hard", return_changes: Union[bool, str] = False
+               ) -> Union[Dict[str, Any], None]:
```
jerbob: Can we change this to something like:
```py
def delete(
    self,
    table_name: str,
    primary_key: Union[str, None] = None,
    durability: str="hard",
    return_changes: Union[bool, str] = False    
) -> Union[Dict[str, Any], None]:
```

jerbob (changes requested):

pysite/constants.py
gdude2002: Did you get around to looking into this?

pysite/views/error_handlers/http_404.py
gdude2002: Won't this conflict with joph's stuff now?

gdude2002 (changes requested):
Where's the database stuff gone?

PR #27, opened 2018-02-26
remove snekchek and only do flake8

Inveracity:


lemonsaurus (approved):
yep, let's wreck the snek

PR #30, opened 2018-02-27
Add error messages

jb3:
Signed-off-by: JoeBanks13 <joseph@josephbanks.me>


jb3:
Note: CI is currently enforcing incorrect import order at the moment. Potentially related to #27.

templates/errors/404.html
```diff
@@ -0,0 +1,15 @@
+{% extends 'main/base.html' %}
+{% block title %} 404 - Page not found! {% endblock %}
+{% block beta_error %}{% endblock %}
+{% block content %}
+    <div class="uk-background-muted">
+        <div class="uk-section">
+            <div class="uk-text-center"><code style="font-size: 170%;">AttributeError: Page not located!</code></div>
```
gdude2002: Are you sure there's no UIKit class that can make the text the size that you want it?

lemonsaurus:
I'd like to see a combination of the approach used in 404 and 500.

- The div with the red text from 404 should always be at the top in all errors. all the text needed in this one is a big, fat, bold error code.
- Below that div, we put the repl simulation.
- The repl simulation should have a dark background
- Only the output the user would actually type should be typed, not the stuff the system would print.
- Dark green for system prints, brighter green for user typed text, and maybe a blue color for returns? Color coding of some kind would make it much nicer to read. Make it look like one of your fancy riced terminals!
- Type speed should probably be increased slightly. Would be amazing if the type speed could be a little dynamic, fast on the simple lines, a little slower on more complex lines, and with slowdowns when there are typos to correct.

templates/errors/404.html
jb3: Upon looking at https://getuikit.com/docs/text I think `.uk-text-large` should do the job.

jb3:
I can add the error div to 5XX, I'll change the styling on the REPL a bit later. Regarding the typing by the system, as far as I'm aware that's not possible using our current library, however, I can probably use the DOM in some way to do this. Coloring of the terminal would probably require more third party libraries, I can look into it though, as well as this there aren't many keywords in the code I've added to the simulation that should be highlighted. I may also switch this out for an IPython terminal. I'll also try fix the type speed, I've tried to make it so system is fast and user is slow, but the library does allow me to change settings dynamically which I can probably use.

As for the coverage issues I need to rewrite some tests for websockets.py as that's currently what's bringing down coverage (my additions brought it down 0.3%). I'll try do that later.

jb3:
Another note, this is still WIP, I probably won't be finished for another day or so, I'll request review once the issues highlighted are fixed as well as coverage is not failing.

jb3:
Syntax highlighting was practically impossible without causing major browser lag, however, I added a Mac like window to the error so it now looks like this:
Example

jb3:
I'm marking this PR as no longer WIP, feel free to leave more reviews.

pysite/views/error_handlers/http_4xx.py
```diff
@@ -0,0 +1,22 @@
...
+class Error400View(ErrorView):
+    name = "error_4xx"
+    error_code = range(400, 430)
+
+    def get(self, error: NotFound):
+        error_desc = ERROR_DESCRIPTIONS.get(error.code, "We're not really sure what happened there, please try again.")
+
+        return render_template("errors/error.html", code=error.code, req=request, error_title=error_desc,
+                               error_message=error_desc +
+                               " If you believe we have made a mistake "
```
jerbob: I think there should be a comma after `mistake`:
```
If you believe we have made a mistake, please open an issue...
```

pysite/views/error_handlers/http_5xx.py
```diff
@@ -1,12 +1,25 @@
...
     def get(self, error: HTTPException):
-        return "Internal server error. Please try again later!", error.code
+        error_desc = ERROR_DESCRIPTIONS.get(error.code, "We're not really sure what happened there, please try again.")
+
+        return render_template("errors/error.html", code=error.code, req=request, error_title=error_desc,
+                               error_message="An error occured while "
+                                             "processing this "
+                                             "request, please try "
+                                             "again later. "
+                                             "If you believe we have made a mistake "
```
jerbob: Ditto with the comma after `mistake`

templates/errors/404.html
```diff
@@ -0,0 +1,15 @@
...
+{% block beta_error %}{% endblock %}
+{% block content %}
+    <div class="uk-background-muted">
+        <div class="uk-section">
+            <div class="uk-text-center"><code class="uk-text-large">AttributeError: Page not located!</code></div>
+        </div>
+    </div>
+
+    <div class="uk-section uk-text-center">
+        <p>The page you requested does not exist, try go back <a href="/">home</a> and try again, if you think this is
```
jerbob: "The page you requested does not exist. Try going back __home__."
"If you think this shouldn't happen, open a ticket on our __GitHub__."

templates/errors/error.html
```diff
@@ -0,0 +1,60 @@
...
+    <style>
+        #terminal {
+            background-color: black;
+            color: white;
+        }
+    </style>
+
+    <noscript>
+        <h1>There was an issue processing your request</h1>
+        <p>Please try again in a little while, if the problem persists, feel free to open a bug on our <a
```
jerbob: "Try again after a while. If the problem persists, open a bug on our __GitHub issue tracker.__"

jerbob (changes requested):

templates/errors/404.html
jb3: This template is not in use anymore. This template will be removed.

jerbob: understandable

jerbob (approved):

PR #34, opened 2018-03-03
API for tags

lemonsaurus:


pysite/base_route.py
```diff
@@ -116,6 +117,9 @@ def error(self, error_code: ErrorCodes) -> Response:
         elif error_code is ErrorCodes.incorrect_parameters:
             data["error_message"] = "Incorrect parameters provided"
             http_code = 400
+        elif error_code is ErrorCodes.database_error:
```
gdude2002: I'm not gonna fail your PR for this block because we need to rethink this entire system - but this is inconsistent.

pysite/route_manager.py
```diff
@@ -27,7 +27,7 @@ def __init__(self):
         self.db = RethinkDB()
         self.log = logging.getLogger()
         self.app.secret_key = os.environ.get("WEBPAGE_SECRET_KEY", "super_secret")
-        self.app.config["SERVER_NAME"] = os.environ.get("SERVER_NAME", "pythondiscord.com:8080")
+        self.app.config["SERVER_NAME"] = os.environ.get("SERVER_NAME", "pythondiscord.local:8080")
```
gdude2002, line 30: You guys really hate actually setting your env vars, don't you? :P

pysite/views/api/bot/tags.py
```diff
@@ -0,0 +1,95 @@
...
+
+from flask import jsonify, request
+
+from pysite.base_route import APIView
+from pysite.constants import ErrorCodes
+from pysite.decorators import api_key
+from pysite.mixins import DBMixin
+
+
+class TagsView(APIView, DBMixin):
```
gdude2002: Should the methods in this class be validated? We have a decorator for that.

gdude2002 (changes requested):

pysite/database.py
```diff
@@ -103,6 +103,34 @@ def create_table(self, table_name: str, primary_key: str="id", durability: str="
             self.log.debug(f"Table created: '{table_name}'")
             return True
 
+    def delete(self, table_name: str, primary_key: Optional[str] = None,
+               durability: str="hard", return_changes: Union[bool, str] = False
```
AnonGuy: Change `str="hard"` to `str = "hard"` to remain consistent with PEP8

pysite/base_route.py
lemonsaurus: I think we should remove all that stuff, after the discussion we already had. But it's not really something we should do on this PR maybe. I can remove the additions I made into the system, though.

gdude2002 (approved):

jerbob (approved):

jerbob (approved):

PR #37, opened 2018-03-05
Add a brief footer with a link to the repo

schwartzadev:
Of course, we can add to this and include whatever y'all like -- let me know if you would like to include any specifics and I'd love to add them!


templates/main/base.html
```diff
@@ -37,5 +37,8 @@
     {% endblock %}
 {% endif %}
 {% block content %}{% endblock %}
+<footer>
+    <p>Developed collaboratively on <a href="https://github.com/discord-python/site">Github</a> • 2018</p>
```
gdude2002: It's generally good practise to use HTML entities with unusual characters. Helps avoid possible encoding issues.

lemonsaurus:
Does the UIKit add any special visual elements to the page based on this code? If so, can we see a picture in the PR of how it looks?

lemonsaurus:
@schwartzadev did you see this comment?

schwartzadev:
@1mn yes - I'm having trouble with my environment as of now -- maybe someone else could pull and take a screenshot?

gdude2002:
Here's how it looks. Needs styling for sure.

image

schwartzadev:
Here's how it looks now
sample

lemonsaurus:
That's not too shabby, actually.

templates/main/base.html
```diff
@@ -37,5 +37,12 @@
     {% endblock %}
 {% endif %}
 {% block content %}{% endblock %}
+<footer>
+<div class="uk-section uk-section-secondary uk-container-medium uk-text-meta">
+    <div class="uk-text-center">
+        <p>Developed collaboratively on <a href="https://github.com/discord-python/site">Github</a> • 2018</p>
```
lemonsaurus: The • symbol should not be a literal, but instead should use the equivalent html entity.

lemonsaurus (changes requested):
local_vars.env is not relevant to this PR.

lemonsaurus (approved):

lemonsaurus:
This seems okay to me. We can merge if someone else approves of it as well. @gdude2002, can you look over it?

gdude2002 (approved):
Looking good! I would have liked it to stick to the bottom of the screen when the page is too short to fill it, but honestly that isn't a massive big deal, and it's *super complicated* (I just gave it a try myself).

Good work!

PR #38, opened 2018-03-05
Add an .env file with default environment vars

schwartzadev:
this should be useful to anyone using an IDE like pycharm since this way they can just point to this file, without having to enter each variable one by one


README.md
```diff
@@ -1,3 +1,7 @@
+# To Run Locally:
+1. CD into the directory with the `rethinkdb.exe`
+2. In a command prompt: `rethinkdb --driver-port 28016 --http-port 28010`
```
gdude2002: This edit is out of scope of this PR.

Also, we prefer `--port-offset 1`

gdude2002:
I don't fully understand the advantages of this approach. We already default env vars in the code - would developers have to modify this file if they needed to change env vars for their development environment?

lemonsaurus:
Not sure this is better, just _different_. The `os.environ` default values seems like it makes this approach redundant.

schwartzadev:
Where are the default vars for config? @1mn 
It seems like some of the variables do have defaults but the `gunicorn_config.py` file doesn't. Maybe I'm missing something?

Inveracity:
@schwartzadev the gunicorn_config.py should actually be removed, it's a remnant for previous infrastructure automation. I'll make a PR to remove that file. Apologies for the confusion!

lemonsaurus:
@schwartzadev if there are any env vars that are missing defaults, that's basically a bug. feel free to raise an issue if that's the case. I'm gonna reject this PR. It's not a bad idea, but I don't think it's necessary for this project.

schwartzadev:
Gotcha. Thanks!

PR #42, opened 2018-03-18
Udacity Python course

ahmedalbayati:
I've added the "Programming Foundations with Python" course from Udacity.

### DIFF
--- a/static/resources.json
+++ b/static/resources.json
@@ -26,17 +26,17 @@
         "payment": "free",
         "payment_description": null
       },
+      "Programming Foundations with Python": {
+        "description": "Learn the Python programming foundations with one of the internet's best courses.",
+        "url": "https://eu.udacity.com/course/programming-foundations-with-python--ud036",
+        "payment": "free",
+        "payment_description": null
+      },
       "Python Tutorials by Corey Schafer on YouTube": {
         "description": "An in-depth look at the Python programming language, from one of YouTube's most popular Python tutors.",
         "url": "https://www.youtube.com/playlist?list=PL-osiE80TeTt2d9bfVyTiXJA-UTHn6WwU",
         "payment": "free",
         "payment_description": null
-      },
-      "Programming Foundations with Python": {
-       "description": "Learn the Python programming foundations with one of the internet's best courses.",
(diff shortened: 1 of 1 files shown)


gdude2002:
These items are supposed to be in alphabetical order - please do that.

gdude2002:
It may also be worth mentioning that this course does expect a foundational Python experience - for example, conditionals, loops and functions - and assumes that you already know them.

gdude2002:
Additionally, you should read the reviews over [here](https://www.class-central.com/course/udacity-programming-foundations-with-python-2013#reviews) and consider those in your submission.

Do you feel that this course is appropriate for absolute beginners?

ahmedalbayati:
Okay, so I alphabetically ordered the course.

ahmedalbayati:
I think the course is appropriate cause it teaches OOP in a very good way. Many beginners have a hard time  differentiating classes from functions. However I think you're still right, cause it doesn't teach much about loops and other necessary stuff. Could making a "beginner - intermediate - advanced" section be a good idea?

ahmedalbayati:
What do you think about [this course](https://eu.udacity.com/course/introduction-to-python--ud1110)?

ahmedalbayati:
That one could be good for the "beginner section" and then we'll put the other one in the "intermediate section".

schwartzadev:
@ahmedalbayati I submitted a PR to your branch to remove the European prefix on the Udacity URL. I would be weary of adding courses that you just find online - you might want to either take the course or spend a while reading the reviews to make sure its quality. Cheers!

ahmedalbayati:
This is not a random course. I took the course while learning Python and really enjoyed it, though it could be considered slightly "intermediate" that's why we should make sections for each level.

gdude2002:
I don't think making sections at this early stage where we have like.. maybe 10 resources is worth doing. We'd just clutter up the page; it'd look awful.

I haven't had time to really look into the course as we're preparing for the code jam at the moment.

gdude2002:
I'm going to close this for now. I don't feel that this recommendation had enough research into it, and I'm specifically cautious because Udacity courses can be created by literally anyone.

PR #43, opened 2018-03-21
Added a Code Jam Countdown (#1eevq)

swedgwood:
countdown1

 - Added countdown to the Code Jam beginning

countdown2

- and to the Code Jame ending

countdown3

- and after it ends, it will say so

(The page automagically reloads upon the code jam starting/ending)

countdown4

 - Also added it to the navigation menu

countdown4

 - For future jams simply change these lines in countdown.js (Months start at 0 for some reason)
 - It is also currently set to the next jam


static/js/countdown.js
```diff
@@ -0,0 +1,50 @@
...
+            minutes = "0"+minutes;
+        }
+        if (seconds < 10) {
+            seconds = "0"+seconds;
+        }
+        daysDisplay.innerHTML = days;
+        hoursDisplay.innerHTML = hours;
+        minutesDisplay.innerHTML = minutes;
+        secondsDisplay.innerHTML = seconds;
+    }, 100);
```
jb3, line 49: Is an interval of 100 needed, probably better to use 1000 and have it update every second

jb3 (commented):
I'll grab your opinion on that comment I added and then approve, and it appears flake8 wants a newline on one of your files.

static/js/countdown.js
swedgwood: I changed it to 100 because at 1000 it can get slightly out of time as executing the code takes time also, this simply reduces that error. It isn't needed though, so I can change it if you wish.

jb3:
Ahh also, your coverage seems to have decreased there, you'll probably want to add a unit test for the countdown route (just assert that it returns 200)

static/js/countdown.js
jb3: If it reduces the error, then that's fine. Address the coverage & flake8 and it's ready

jb3 (approved):
Looks good to me.

PR #47, opened 2018-03-30
Changes -am to -m

robertt:
Pretty sure this is an error, and it should be `git commit -m`, not `git commit -am`

### DIFF
--- a/templates/main/info/jams.html
+++ b/templates/main/info/jams.html
@@ -135,7 +135,7 @@
                 </p>
                 <p>
                     Next up, we'll need to bundle up our changes into a commit, and push it to GitHub. To do that, we'll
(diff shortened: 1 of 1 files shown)


gdude2002:
Why do you think this is an error?

jb3:
No, that's intentional. The `-a` flag adds all unstaged files, note how we don't say anything about `git add` in that. If you see this as an error then feel free to bring it up but this is not a typo or error.

robertt:
Yes but its more readable, also maybe do `git commit -a -m "Message"`, then you can explain what -a does.

gdude2002:
This wasn't supposed to be a definitive guide. @JoeBanks13 is incorrect in that we do talk about adding files directly, but I included `a` because the guide is for absolute beginners to contribute to their code jam projects - not for actually learning Git.

robertt:
Understandable

PR #49, opened 2018-03-30
Adding atom

hargoniX:


static/resources.json
```diff
@@ -54,6 +54,12 @@
         "url": "https://www.sublimetext.com/",
         "payment": "optional",
         "payment_description": "Nagware; will ask you to buy the full version after every X saves"
+      },
+      "Atom (Free)": {
+          "description": "A free Electron based editor, a \"hackable text editor for the 21st century\".",
```
gdude2002: Should be `Electron-based`, not `Electron based`

gdude2002: Perhaps also mention that Atom is maintained by the GitHub team - at least, I think that's the case, right?

gdude2002 (changes requested):
Resources should be in alphabetical order

gdude2002:
(I just noticed that VS code is at the top, so rearrange it all anyway)

static/resources.json
hargoniX: It is

(oh now it is alphabetic :stuck_out_tongue:)

PR #50, opened 2018-04-11
Added an About category with Partners view (#1pv7h)

jerbob:
• Added the "About" category, accessible via the Navbar
• The About category now contains the Server Rules, as well as a new Partners page
• The Partners page contains cards for each partner, with responsive columns

image

image


templates/main/index.html
```diff
@@ -32,4 +32,4 @@ <h1 class="uk-header">
 
     </div>
 </div>
-{% endblock %}
+{% endblock %}
```
gdude2002: Missing newline at end of file

templates/main/about/partners.html
```diff
@@ -0,0 +1,119 @@
...
+                                Join the server
+                            </a>
+                        </div>
+                    </div>
+
+                </div>
+            </article>
+        </div>
+    </div>
+{% endblock %}
```
gdude2002: Missing newline at end of file

templates/main/about/index.html
```diff
@@ -0,0 +1,34 @@
...
+                <p>
+                    <strong>Server Rules</strong> &nbsp;
+                    <a href="{{ url_for('main.about.rules') }}"><i class="fas fa-link"></i></a>
+                    <br />
+                    <span class="uk-text-meta">Rules and infraction information for our server</span>
+                </p>
+            </article>
+        </div>
+    </div>
+{% endblock %}
```
gdude2002: Missing newline at end of file

templates/main/about/partners.html
```diff
@@ -0,0 +1,119 @@
+{% extends "main/base.html" %}
+{% block title %}Partners{% endblock %}
+{% block og_title %}Partners{% endblock %}
+{% block og_description %}List and Links to Partner servers{% endblock %}
+{% block content %}
+    <div class="uk-section">
+        <div class="uk-container uk-container-small">
+            <article class="uk-article">
+                <h1 class="uk-article-title hover-title" id="top">
+                    Partners
```
gdude2002, line 10: Can we generate this from a JSON file or something? It seems pretty standard and it'd be nice if we could modify it easily later on

gdude2002 (dismissed):

gdude2002:
Oh yeah, and since you're moving the routes to `/about`, maybe you'll want to leave behind redirects for the `/info` routes

templates/main/about/partners.html
jerbob: Fixed

templates/main/about/index.html
jerbob: Fixed

templates/main/about/partners.html
jerbob: The Partners cards are now generated from `static/partners.json`.

lemonsaurus (dismissed):
Overall the code looks good, but the partners page does not have a clean system for containing the elements, and it creates severe bugs such as this one:

image

The flexboxing needs to be fixed here before we can commit this. The main container should expand to contain the entire card, and it should be split into two divs. Then we should spit out the partners into the right div by using jinja to alternate the display position.

in simpler terms: 

_build a big flexbox, set it to columns. put two boxes in it. loop through the partner list and put all even partners in the left column. loop through it again and put all odd partners in the right column._

static/style.css
```diff
@@ -125,4 +125,13 @@ Rendered document fixes
...
     margin-bottom: 0;
-}
\ No newline at end of file
+}
+
+/*
+Styling for Partners page cards
+ */
+
+.uk-card {
```
lemonsaurus: Are we sure we want all objects with this class to have this styling in the future, or should this perhaps have greater specificity?

templates/main/about/index.html
```diff
@@ -0,0 +1,34 @@
...
+
+                    <a href="#top" class="uk-text-primary" title="Permanent link to this header">
+                        <i class="fas fa-paragraph" data-fa-transform="shrink-8"></i>
+                    </a>
+                </h1>
+                <p class="uk-article-meta">
+                    Learn about the server, find links to our Partners, and more!
+                </p>
+                <p>
+                    <strong>Server Partners</strong> &nbsp;
```
lemonsaurus: wait what's that nonbreaking space for?

templates/main/about/partners.html
```diff
@@ -0,0 +1,48 @@
...
+                    <a href="#top" class="uk-text-primary" title="Permanent link to this header">
+                        <i class="fas fa-paragraph" data-fa-transform="shrink-8"></i>
+                    </a>
+                </h1>
+                <div class="uk-column-1-2@m uk-column-divider">
+                    {% if partners is none %}
+                    <div class="uk-alert-danger" uk-alert>
```
lemonsaurus: I don't agree with this indentation - conditional content should be indented inside of jinja conditionals imo.

templates/main/about/partners.html
```diff
@@ -0,0 +1,48 @@
...
+                        <p>
+                            We were unable to load the <code>partners.json</code> file. If you see this, please
+                            notify us!
+                        </p>
+                    </div>
+                    {% else %}
+                        {% for partner_name, partner_data in partners.items() %}
+                        <div class="uk-inline">
```
lemonsaurus: also applies to this.

templates/main/about/partners.html
```diff
@@ -0,0 +1,48 @@
...
+                            <div class="uk-card uk-card-hover uk-card-default">
+                                <div class="uk-card-media-top">
+                                    <a href="{{ partner_data['invite'] }}">
+                                        <img src="{{ partner_data['banner'] }}" alt="{{ partner_name }}">
```
lemonsaurus: unclosed tag. should be `{{ partner_name }}`

lemonsaurus (dismissed):
fuckin' change some shit.

templates/main/about/partners.html
```diff
@@ -0,0 +1,48 @@
...
+                    {% if partners is none %}
+                        <div class="uk-alert-danger" uk-alert>
+                            <p>
+                                We were unable to load the <code>partners.json</code> file. If you see this, please
+                                notify us!
+                            </p>
+                        </div>
+                    {% else %}
+                        {% for partner_name, partner_data in partners.items() %}
+                            <div class="uk-inline">
```
lemonsaurus: I don't think you need this div. You could just apply this class to the card div itself, no? Be wary of div bloat.

templates/main/about/partners.html
```diff
@@ -0,0 +1,48 @@
...
+                                <div class="uk-card uk-card-hover uk-card-default" id="partner-card">
```
lemonsaurus: id's must be unique, so I'm afraid this isn't good enough. You've got the right idea, though. how about you put an ID on the `<div class="uk-column-1-2@m uk-column-divider">`, and use a selector that says "select the divs inside of a div with this id, but only if they have the class `uk-card`. That's very specific so it won't accidentally style someone elses code, and it also won't apply to the uk-alert-danger div.

lemonsaurus (dismissed):
aaalmost there.

lemonsaurus (approved):
Looks :ok_hand:

PR #53, opened 2018-04-12
Initial data for tables and refactoring the dev-mode database handling to be similar to prod-mode.

lemonsaurus:
Changed the dev-mode logic to be the same as prod for creating new tables if they don't exist. Also added a new feature where a table can be initialized with data if you create a JSON file in the pysite/database/table_init/ folder and fill it with a list of dicts where each dict represents a row in your table. Included a hiphoppers json so that I can actually test if it works in production. It will only init the table if the table is empty.


gdude2002:
Your coverage fails are probably related in some way to the fact that I haven't unit-tested the wiki yet.

pysite/database.py
```diff
@@ -8,13 +9,16 @@
...
+from pysite.constants import DEBUG_MODE
+
 ALL_TABLES = {
     # table: primary_key
-
     "oauth_data": "id",
     "tags": "tag_name",
     "users": "user_id",
     "wiki": "slug",
+    "hiphoppers": "hophop_name",
```
gdude2002: Shouldn't this be `hiphop_name`?

gdude2002 (changes requested):

gdude2002:
I'm not sure I'm entirely happy with the names of the tables or their primary keys, but I guess that's okay.

pysite/database.py
lemonsaurus: oh.. yeah it should. :D 

It could probably just be `name`

lemonsaurus: btw I am open to suggestions for table and primary key names. I'm not happy with them either.

gdude2002: Well really I do think that tables should be sectioned off from other parts of the database - so like hh_users, hh_names, that kind of thing.

Also, this dict should be alphabetical, surely.

gdude2002 (approved):

PR #55, opened 2018-04-19
Added image URLs for each of these famous rappers.

biskette:
I should probably leave a comment here.


lemonsaurus (approved):
Excellent. Thanks for helping out.

PR #66, opened 2018-05-01
Fixed those pesky rapper names.

biskette:
I think I've done this properly now.


biskette:
Nope, i went and messed it up again.

lemonsaurus (approved):
Yep, looks good now!

PR #68, opened 2018-05-10
[WIP] JSON table init upgraded to JSON table sync.

lemonsaurus:
Tables that are initialized with JSON files in the table_init folder will now sync properly if the JSON file is changed after initialization. Any lines that are either added or removed from the JSON file will be updated to the corresponding table whenever site is rebooted.

### DIFF
--- a/pysite/database.py
+++ b/pysite/database.py
@@ -96,31 +96,43 @@ class RethinkDB:
 
         for table, primary_key in ALL_TABLES.items():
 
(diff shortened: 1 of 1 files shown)


lemonsaurus:
Hmm, this accounts for additions and deletions, but not for edits. We should probably catch that too. I guess I'll set this to WIP while I figure out how.

Zwork101 (approved):
Looks good!

lemonsaurus:
gdude has a better idea.

PR #69, opened 2018-05-10
Migration runner and migrations

gdude2002:


pysite/database.py
```diff
@@ -500,7 +452,7 @@ def insert(self, table_name: str, *objects: Dict[str, Any],
         )
 
         if return_changes:
-            return self.run(query, coerce=list)
+            return self.run(query, coerce=dict)
         else:
             return self.run(query, coerce=dict)
```
jchristgit: if we always return it as a `dict`, changing the docstring to reflect that would possibly be useful?
Also, these two branches seem to do the same.

pysite/views/staff/tables/edit.py
```diff
@@ -0,0 +1,105 @@
...
+            )
+
+            return self.render(
+                "staff/tables/edit.html", table=table, primary_key=primary_key, document=document,
+                message="Please provide some data to save", old_primary=old_primary
+            )
+
+        try:
+            data = json.loads(data)
+        except Exception as e:
```
jchristgit: a bit broad, isn't it? What exceptions could this throw? `JSONDecodeError`?

pysite/migrations/runner.py
```diff
@@ -0,0 +1,95 @@
...
+                    # We do this in the loop to save our progress, in case we fail during a migration
+
+                    db.insert(
+                        VERSIONS_TABLE,
+                        {"table": table, "version": current_version},
+                        conflict="replace",
+                        durability="soft"
+                    )
+            except Exception as e:
+                # TODO: Should this fail hard and just raise?
```
lemonsaurus: probably, yes.

pysite/tables.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Table(NamedTuple):
+    primary_key: str
+    keys: List[str]
+    locked: bool = True
+
+
+TABLES = {
+    "hiphopify": Table("user_id", sorted(["user_id", "end_timestamp", "forced_nick"])),
```
lemonsaurus: I don't like these. I think they're too dense, and I don't understand why you're using sorted instead of just listing them alphabetically. I think I'd prefer these multilined with a blank line seperating them or something like that.

How about something like

```py
"hiphopify": Table(
    primary_key="user_id", 
    keys=[
        "end_timestamp", 
        "forced_nick",
        "user_id",
    ]
),

"hiphopify_namelist": Table(
    primary_key="name", 
    keys=[
        "image_url",
        "name",
    ], 
    locked=False
),
```

pysite/views/staff/tables/edit.py
```diff
@@ -0,0 +1,112 @@
...
+from flask import redirect, request, url_for
+from werkzeug.exceptions import BadRequest, NotFound
+
+from pysite.base_route import RouteView
+from pysite.constants import TABLE_MANAGER_ROLES
+from pysite.decorators import csrf, require_roles
+from pysite.mixins import DBMixin
+from pysite.tables import TABLES
+
+DEFAULT_OLD_PRIMARY = "!None!"
```
lemonsaurus: can you explain what this is all about? never seen anything like `!None!` before, almost looks spanish 😛. Why can't it just be `None`?

lemonsaurus (changes requested):
Looks reasonable for the most part, but I still need to test the UX and stuff when I get back into linux. A couple of minor requests for changes. will review again after testing.

pysite/tables.py
gdude2002: The keys must be sorted the same way they're sorted by the editor interface. There's no real other option there.

pysite/views/staff/tables/edit.py
gdude2002: It goes into a hidden input on the page. You need to know what the old primary key was, so you can check whether it was edited.

lemonsaurus (approved):

pysite/database.py
```diff
@@ -480,7 +410,7 @@ def insert(self, table_name: str, *objects: Dict[str, Any],
                        [Dict[str, Any], Dict[str, Any]],  # ...takes two dicts with string keys and any values...
                        Dict[str, Any]  # ...and returns a dict with string keys and any values
                    ]
-               ] = "error") -> Union[List, Dict]:  # flake8: noqa
+               ] = "error") -> Dict[str, Any]:  # flake8: noqa
```
jchristgit, line 413: Judging from the annotation, this always seems to return a dictionary, the docstring specifies:
```py
         :return: A list of changes if `return_changes` is True; a dict detailing the operations run otherwise
```
should probably be updated

pysite/migrations/runner.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    migrations = {}
+    final_version = 0
+
+    for filename in os.listdir(table_path):
+        if filename.startswith("v") and filename.endswith(".py"):
+            final_version = int(filename[1:-3])
+            migrations[final_version] = f"pysite.migrations.tables.{table}.v{final_version}"
+
+    return dict(sorted(migrations.items())), final_version
```
jchristgit: why the `sorted`?

jchristgit (dismissed):
looks good. i feel like the new views should have some docstrings, and i haven't spotted tests either, buuut okay

lemonsaurus (changes requested):
Tested the UX. For the most part this is excellent, but the pagination has infuriating UX. Right now there's 127 pages on hiphopify_namelist, and if I want to change some random rapper whose name starts with a t, I would have to go through a ton of pages to locate him.

I'd like to see either a filter search bar where I can just type the data I'm looking for, or at the very least a "display everything on one long-ass-page" button, so I can use CTRL F. Ideally both.

pysite/migrations/runner.py
gdude2002: Because migrations have to be run in order

lemonsaurus (approved):

lemonsaurus:
yep, now it's excellent.

jchristgit (approved):
apart from my ??? on the table key sorting it looks pretty clean

PR #73, opened 2018-05-20
Tests directory

Inveracity:
```
pipenv run test
```


gdude2002 (dismissed):

tests/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+from flask import Blueprint
+from flask_testing import TestCase
+
+os.environ["BOT_API_KEY"] = "abcdefg"  # This is a constant, must be done first
+os.environ["PAPERTRAIL_ADDRESS"] = 'localhost'  # satisfies coverage
+os.environ["DATADOG_ADDRESS"] = 'localhost'  # satisfies coverage
+try:
+    del os.environ["FLASK_DEBUG"]  # Some unit tests fail if this is set
+except KeyError:
+    pass
```
jchristgit: Probably just me, but I think just checking whether it exists would be a fair bit cleaner.
```py
if "FLASK_DEBUG" in os.environ:
    del os.environ["FLASK_DEBUG"]  # Some unit tests fail if this is set
```

tests/test_api.py
```diff
@@ -0,0 +1,25 @@
+from tests import SiteTest, app
+
+class ApiEndpointsRootEndpoints(SiteTest):
```
jchristgit, line 3: doesn't flake8 demand two newlines before this?

tests/test_api_bot_snake.py
```diff
@@ -0,0 +1,44 @@
...
+class ApiBotSnakeEndpoints(SiteTest):
+    """
+    Tests the following endpoints:
+        - snake_movies
+        - snake_quiz
+        - snake_names
+        - snake_idioms
+        - snake_facts
+    """
+    os.environ['BOT_API_KEY'] = 'abcdefg'
```
jchristgit: why is this in the class?

tests/test_decorators.py
```diff
@@ -0,0 +1,27 @@
...
+        SCHEMA = Schema([{"user_id": int, "role": int}])
+
+        @api_params(schema=SCHEMA, validation_type=ValidationTypes.json)
+        def try_json_type(data):
+            return data
+
+        try:
+            try_json_type("not json")
+        except Exception as error_message:
+            self.assertEqual(type(error_message), AttributeError)
```
jchristgit: how about [`assertRaises`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) instead?
```py
with self.assertRaises(AttributeError):
    try_json_type("not json")
```

tests/test_mixins.py
```diff
@@ -0,0 +1,68 @@
...
+    def test_route_view_runtime_error(self):
+        """ Check that wrong values for route view setup raises runtime error """
+        from pysite.base_route import RouteView
+
+        rv = RouteView()
+        try:
+            rv.setup(manager, 'sdfsdf')
+        except RuntimeError:
+            return True
+        raise Exception('Expected runtime error on setup() when giving wrongful arguments')
```
jchristgit: this would probably also be nicer with `self.assertRaises`

tests/test_mixins.py
```diff
@@ -0,0 +1,68 @@
...
+
+        from pysite.route_manager import RouteView
+
+        class TestRoute(RouteView):
+            name = "test"
+            path = "/test"
+
+        tr = TestRoute()
+        tr.setup(manager, Blueprint("test", "test_name"))
+        self.assertIs(tr.user_data, None)
```
jchristgit: [`TestCase.assertIsNone`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsNone) might be cleaner

tests/test_mixins.py
```diff
@@ -0,0 +1,68 @@
...
+
+        from pysite.route_manager import RouteView
+
+        class TestRoute(RouteView):
+            name = "test"
+            path = "/test"
+
+        tr = TestRoute()
+        tr.setup(manager, Blueprint("test", "test_name"))
+        self.assertIs(tr.logged_in, False)
```
jchristgit: [`TestCase.assertFalse`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertFalse) might be cleaner

tests/test_oauth_backend.py
```diff
@@ -0,0 +1,39 @@
...
+        """ Make sure the get function returns nothing """
+        self.assertIs(manager.oauth_backend.get(), None)
+
+    def test_delete(self):
+        """ Make sure the delete function returns nothing """
+        self.assertIs(manager.oauth_backend.delete(None), None)
+
+    def test_logout(self):
+        """ Make sure at least apart of logout is working :/ """
+        self.assertIs(manager.oauth_backend.logout(), None)
```
jchristgit: these would probably all be cleaner with the `assertIsNone` as shown above

tests/test_roots.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def test_not_found(self):
+        """ Check paths without handlers returns 404 Not Found """
+        response = self.client.get('/nonexistentpath')
+        self.assertEqual(response.status_code, 404)
+
+    def test_error(self):
+        """ Check the /error/XYZ page """
+        response = self.client.get('/error/418')
+        self.assertEqual(response.status_code, 418)
```
jchristgit: this only checks a single error page, i'm not sure if it's desired, but how about checking all of them with something like
```py
for code in HANDLED_ERROR_CODES:
    response = self.client.get(f'/error/{code}')
    self.assertEqual(response.status_code, code)
```

tests/test_utilities.py
```diff
@@ -0,0 +1,27 @@
...
+    def test_error_view_runtime_error(self):
+        """ Check that wrong values for error view setup raises runtime error """
+        import pysite.base_route
+
+        ev = pysite.base_route.ErrorView()
+        try:
+            ev.setup(manager, 'sdfsdf')
+        except RuntimeError:
+            return True
+        raise Exception('Expected runtime error on setup() when giving wrongful arguments')
```
jchristgit: this might also benefit from `self.assertRaises`

tests/test_utilities.py
```diff
@@ -0,0 +1,27 @@
...
+        import pysite.websockets
+
+        class TestWS(pysite.websockets.WS):
+            pass
+
+        try:
+            TestWS(None).on_message("test")
+            return False
+        except NotImplementedError:
+            return True
```
jchristgit, line 25: what would `return True` / `return False` do in a test? `self.assertRaises` is probably also wanted here?

tox.ini
```diff
@@ -2,5 +2,5 @@
 max-line-length=120
 application_import_names=pysite
 ignore=P102,B311,W503,E226,S311
-exclude=__pycache__, venv, app_test.py, .venv
+exclude=__pycache__, venv, .venv, tests
```
jchristgit, line 5: this is flake8, right? why do we exclude tests from linting?

jchristgit (changes requested):
some thoughts and ideas

tox.ini
Inveracity: we exclude tests from linting because some tests require some terrible python code that flake8 really dislikes

tests/__init__.py
Inveracity: good point!

tests/test_api_bot_snake.py
Inveracity: fixed 👍

tests/test_api.py
Inveracity: skipping

tests/test_decorators.py
Inveracity: I like it, sorted :+1:

tests/test_mixins.py
Inveracity: done! 👍

tests/test_mixins.py
Inveracity: done 👍

tests/test_mixins.py
Inveracity: captain done 👍

tests/test_oauth_backend.py
Inveracity: triple done 👍

tests/test_roots.py
Inveracity: The other status codes are satisfied in other tests, but I do like this, so I'm gonna try it

Inveracity: looped done 👍

tests/test_utilities.py
Inveracity: I don't know who wrote this, so I'm gonna skip it

jchristgit (approved):
👍

PR #75, opened 2018-05-20
Privacy/Usability updates

gdude2002:
* Rewrite and update the privacy policy
* Allow people to remove their code jam profile/code jam responses
    * Infractions will be added automatically if they were participating in a code jam
* Remove profile/responses and hand out infractions as appropriate on server leave as well
* Prevent users younger than 13 from submitting their code jam profile, therefore preventing them from joining
* Also redirecting to login will return you to the right page after login.


pysite/migrations/tables/users/v1.py
```diff
@@ -0,0 +1,7 @@
+def run(db, table, table_obj):
+    for document in db.get_all(table):
+        if "email" in document:
+            del document["email"]
+
+            db.insert(table, document, conflict="update", durability="soft")
+    db.sync(table)
```
jchristgit, line 11: I feel like these tests should have some docs to them. I can *somewhat* get behind that this is removing all email fields from users, but I feel like having a docstring describe it would be nice regardless. or alternatively, the django way:tm:, have the migration filenames reflect it, e.g. `v1_remove_email_field`. not sure if the migration engine supports that though

pysite/views/api/bot/user.py
```diff
@@ -56,10 +63,42 @@ def post(self, data):
...
+                    coerce=list
+                )
+
+                for response in responses:
+                    jam = response["jam"]
+                    jam_obj = self.db.get(self.jams_table, jam)
+
+                    if jam_obj:
+                        if jam_obj["state"] in BANNABLE_STATES:
+                            banned = True
```
jchristgit, line 87: maybe
```py
banned = jam_obj["state"] in BANNABLE_STATES
```
?

pysite/views/main/jams/retract.py
```diff
@@ -0,0 +1,84 @@
+from werkzeug.exceptions import BadRequest
+
+from pysite.base_route import RouteView
+from pysite.decorators import csrf
+from pysite.mixins import DBMixin, OAuthMixin
+
+BANNABLE_STATES = ["preparing", "running"]
```
jchristgit: do we need to modify this? if not, why not make it a tuple?

pysite/views/api/bot/user.py
```diff
@@ -26,13 +26,18 @@
     }
 ])
 
+BANNABLE_STATES = ["preparing", "running"]
```
jchristgit: do we need to modify this? if not, why not make it a tuple?

pysite/views/main/jams/retract.py
```diff
@@ -0,0 +1,84 @@
...
+        participant = self.db.get(self.table_name, user_id)
+
+        banned = False
+
+        if participant:
+            responses = self.db.run(self.db.query(self.responses_table).filter({"snowflake": user_id}), coerce=list)
+
+            for response in responses:
+                jam = response["jam"]
+                print(jam)
```
jchristgit: this looks like a leftover from debugging?

pysite/views/main/jams/retract.py
```diff
@@ -0,0 +1,84 @@
...
+
+        responses = self.db.run(self.db.query(self.responses_table).filter({"snowflake": user_id}), coerce=list)
+
+        for response in responses:
+            jam = response["jam"]
+            jam_obj = self.db.get(self.jams_table, jam)
+
+            if jam_obj:
+                if jam_obj["state"] in BANNABLE_STATES:
+                    banned = True
```
jchristgit, line 65: huh, I've seen this before.. but again, what about 
```py
banned = jam_obj["state"] in BANNABLE_STATES
```
instead?

templates/main/about/privacy.html
```diff
@@ -20,119 +20,212 @@ <h1 class="uk-article-title hover-title" id="top">
...
+                            <th class="uk-table-shrink">When</th>
+                            <th style="max-width: 30rem;">Usage</th>
+                            <th>Access</th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                        <tr>
+                            <td>Discord user ID</td>
+                            <td class="uk-table-shrink"><code>self.accept()</code> run on Discord</td>
+                            <td style="max-width: 30rem;">Statistics, data association (infractions, code jam applications, etc)</td>
```
jchristgit: aren't inline styles frowned upon?

templates/main/jams/profile.html
```diff
@@ -75,19 +86,47 @@ <h1 class="uk-header uk-article-title">
...
+            if (!dob.isBefore(earliest_dob)) {
+                UIkit.notification({
+                    "message": "You must be aged 13 or older to participate in our code jams.",
+                    "status": "danger",
+                    "pos": "top-center",
+                    "timeout": 5000,
+                });
+
+                dob_output.classList.add("uk-form-danger");
+                return submit_button.disabled = true;
```
jchristgit: what

jchristgit (changes requested):
i think i saw some stuff twice. i might be going insane though. some minor thoughts

templates/main/jams/profile.html
gdude2002: I figured it was more concise than:

```js
submit_buton.disabled = true;
return;
```

templates/main/about/privacy.html
gdude2002: It depends, we do have them sprinkled liberally around. I do write actual CSS in the `style.css` if I think it'll be useful elsewhere, but often it's just specific styling for the one page.

pysite/views/api/bot/user.py
gdude2002: Nope, users might be in more than one jam, so we have to at least do this check to break out of the loop.

pysite/views/main/jams/retract.py
gdude2002: Nope, users might be in more than one jam, so we have to at least do this check to break out of the loop.

pysite/views/main/jams/retract.py
gdude2002: Done.

pysite/views/api/bot/user.py
gdude2002: Done.

pysite/views/main/jams/retract.py
gdude2002: Done.

pysite/migrations/tables/users/v1.py
gdude2002: Done.

pysite/views/api/bot/user.py
jchristgit: ah, okay

templates/main/about/privacy.html
jchristgit: alright

jchristgit (dismissed):
with my stuff fixed i like it

pysite/views/main/jams/retract.py
lemonsaurus: wait, uh.. am I stuck in a time loop? will it just scroll forever?

templates/main/about/privacy.html
lemonsaurus: personally I do frown upon them.

tests/test_oauth_backend.py
```diff
@@ -23,7 +23,7 @@ def test_add_user(self):
 
         sess_id = "hey bro wazup"
```
lemonsaurus, line 24: :sunglasses:

templates/main/jams/profile.html
lemonsaurus: no no no no no no no no no. why do you even need to return at all?

lemonsaurus (changes requested):
This is quite excellent work.

templates/main/jams/profile.html
gdude2002: Oh, good point. Blame copypasta.

pysite/views/main/jams/retract.py
gdude2002: It might get set to False again when we iterate over the next response.

lemonsaurus (approved):

PR #77, opened 2018-05-21
Revisions improvements

jb3:
This brings two new additions to the revision system:
* Moves revisions upon page slug change.
* Deletes revisions upon page deletion.


lemonsaurus (approved):
seems entirely reasonable. let's do it.

PR #79, opened 2018-05-25
Add Google's Python course to resources

AileenLumina:
Seems worth a mention to me. :)

### DIFF
--- a/static/resources.json
+++ b/static/resources.json
@@ -221,6 +221,17 @@
                     }
                 ]
             },
+            "Google's Python Class": {
+                "description": "A free class for people with a little bit of programming experience who want to learn Python.",
+                "payment": "free",
+                "urls": [
+                    {
+                        "icon": "regular/link",
+                        "title": "Website",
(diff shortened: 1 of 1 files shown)


gdude2002:
This course enforces Google's Python style guide, rather than the much preferred PEP8. That could be problematic.

lemonsaurus:
I am not sure I see this as a problem. The google style guide agrees with PEP8 on nearly everything. it's really meant more as a complimentary guide than a substitute

justalemon:
As a side note, this course is for Python 2.6+ (from https://developers.google.com/edu/python/set-up#python-on-linux-mac-os-x-and-other-os):

```
Although Python 3.x is becoming more popular, this course is designed for Python 2.6 or later.
```

lemonsaurus:
that, I am far more troubled by. I think that disqualifies it.

AileenLumina:
As the one who originally brought this up, I agree, and I think this should be closed, but I'll let you decide. ^^

PR #85, opened 2018-06-02
[Wiki] Prefer horizontal whitespace over vertical

nnsee:
This should fix the issue with the long vertical bar of whitespace under the menu
toggle button on mobile, whilst also not breaking centering of content.

Before this commit, pages looked like this on mobile:

wiki pythondiscord com_wiki_home

This was quite inefficient use of space, especially on pages with more content. This commit aims to replace that long vertical whitespace on the left with a horizontal one above the page content, so the page looks like this:

wiki pythondiscord com_wiki_home 1

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -249,7 +249,7 @@ textarea.fira-code {
 }
 
 #doc-view div {
-    width: calc(100% - 6rem);
+    width: calc(100% - 2rem);
 }
 
 @media (max-width: 639px) {
--- a/templates/wiki/base.html
+++ b/templates/wiki/base.html
@@ -188,11 +188,13 @@
                         </li>
                     </ul>
                 </div>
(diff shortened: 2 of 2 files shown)


lemonsaurus:
This is an improvement. If you think you can improve it even more, keep iterating. there's still a lot of whitespace at the top, the button is in a slightly awkward spot, there may be more optimal configurations. feel free to think outside the box a bit.

also I hate the inline style. both the stuff you added, and the stuff that was already here. a cleanup of all our templates is needed, imo, because we really shouldn't be doing them at all for reasons outlined in #meta.

lemonsaurus:
Also please add screenshots showing how these changes affect desktop as well.

nnsee:
This is what the page looks with the changes in desktop mode (pardon the low screen width, I've got an older screen):
wiki pythondiscord local_8080_wiki_home

gdude2002 (approved):

PR #86, opened 2018-06-03
Move from CSS to SCSS

gdude2002:
- [x] Convert existing `style.css` into a set of `.scss` files
- [x] Bring in the UIKit scss
- [x] Replace in-page `style` attributes with appropriately-specific `.scss` files
- [x] Factor out all the significant constants (eg colours) into variables
- [x] Rewrite `window.css` and any other relevant `.css` files into `.scss` files
    - [x] Have @JoeBanks13 fix `_errors.scss`
- [x] Linting (and CI)
- [ ] [Documentation on the wiki](https://wiki.pythondiscord.com/wiki/contributing/project/site)


scss.py
```diff
@@ -0,0 +1,35 @@
...
+Single-file libsass wrapper that can compile SASS/SCSS files into CSS.
+
+Usage: `python scss.py input.scss:output.css input2.scss:output2.css`
+"""
+
+import sys
+
+import sass
+
+if not len(sys.argv) > 1:
```
jchristgit: this is weird. maybe it's just me but this feels a bit like doing `if not x == 3` instead of `if x != 3`. Why not just use `if len(sys.argv) <= 1`?

gdude2002:
We're looking good, and the checklist is complete. Just need to finish the site contributor documentation!

gdude2002:
Reviewers: Please note, this pull request also contains **all of UIKit's SCSS**. If you're reviewing then you might have a better time doing so from within your IDE, as UIKit is something like 150 files.

gdude2002:
Because the site is currently down, I can't write the documentation.

.scss-lint.yml
```diff
@@ -0,0 +1,17 @@
+linters:
+  Comment:
+    enabled: false
+  HexLength:
+    style: "long"
+  HexNotation:
+    style: "uppercase"
+  IdSelector:
+    enabled: false
+  ImportantRule:  # While I agree with this one, we do need to use !important in some cases
```
jchristgit, line 10: Is there maybe a way to exclude certain lines from the linting, or would that be overkill?

scss/pysite/pages/errors/_common.scss
```diff
@@ -0,0 +1,138 @@
...
+      height: 500px;
+      margin: 20px auto auto;
+      width: 100%;
+    }
+
+    .inside {
+      background: $black;
+      height: 100%;
+      padding-right: 20px;
+
```
jchristgit: do we need the extra newline here? it's just a bit weird to the eyes and the others dont have a trailing newline either

static/css/uikit_blurple.css
```diff
@@ -0,0 +1,3 @@
...form:translateY(10px)}100%{opacity:1;transform:translateY(0)}}@keyframes uk-fade-left-small{0%{opacity:0;transform:translateX(-10px)}100%{opacity:1;transform:translateX(0)}}@keyframes uk-fade-right-small{0%{opacity:0;transform:translateX(10px)}100%{opacity:1;transform:translateX(0)}}@keyframes uk-fade-top-medium{0%{opacity:0;transform:translateY(-50px)}100%{opacity:1;transform:translateY(0)}}@keyframes uk-fade-bottom-medium{0%{opacity:0;transform:translateY(50px)}100%{opacity:1;transform:translateY(0)}}@keyframes uk-fade-left-medium{0%{opacity:0;transform:translateX(-50px)}100%{opacity:1;transform:translateX(0)}}@keyframes uk-fade-right-medium{0%{opacity:0;transform:translateX(50px)}100%{opacity:1;transform:translateX(0)}}@keyframes uk-fade-scale-02{0%{opacity:0;transform:scale(0.2)}100%{opacity:1;transform:scale(1)}}@keyframes uk-fade-scale-18{0%{opacity:0;transform:scale(1.8)}100%{opacity:1;transform:scale(1)}}@keyframes uk-scale-kenburns{0%{transform:scale(1)}100%{transform:scale(1.2)}}@keyframes uk-shake{0%,100%{transform:translateX(0)}10%{transform:translateX(-9px)}20%{transform:translateX(8px)}30%{transform:translateX(-7px)}40%{transform:translateX(6px)}50%{transform:translat
```
jchristgit: this is a bit weird, why does the line start super late?

Wim-De-Clercq (commented):
All looks good. I can't even find missing semicolons. All !importants that were 'removed' from the html are still present in the css, so that's good.

The only nitpick would be that scss/pysite/pages/errors/_common.scss seems to have more unusual [Enter]s than other files.

I didn't actually run it though. This is just from reading.

jchristgit (dismissed):
Nice to see the inline styles gone.

static/css/uikit_blurple.css
jchristgit: uh, ok. For some reason the majority of the line didn't show up in the review thing. My bad

.scss-lint.yml
gdude2002: You can exclude blocks, but not specific lines

static/css/uikit_blurple.css
gdude2002: That's generated CSS anyway, we don't write this

Wim-De-Clercq (dismissed):

PR #148, opened 2018-11-18
Fix security vulnerabilities

jb3:
As suggested by GitHub Alerts, this PR updates:
- Gulp to 4.0.0
- All node dependencies to ones compatible with Gulp 4.0.0
- Upgrades Flask to a version without a vulnerability.
- Unused node dependencies removed


jb3:
actually, hol up boys, i'm going to use this as an excuse to clean out the old node deps which are unused.

MarkKoz (approved):

PR #149, opened 2018-11-19
Django - Add Infractions API

### LINKED ISSUE #124: Django - API: Infractions
**Originally posted by Johannes Christ:**

Add the infractions API. The current implementation along with routes can be found [here](https://github.com/python-discord/site/blob/master/pysite/views/api/bot/infractions.py).

MarkKoz:
Closes #124.

To-do:

- [ ] Validation
    - [ ] `expires_at`
        - [ ] Is in the future
        - [x] Must be `null` depending on `type`
    - [x] Choices for `type`
    - [x] `hidden`
        - [x] Must be `False` for superstar
- [x] Implement all get endpoints
    - [x] GET
        - [x] Filter using query parameters
            * This should also replace functionality of `dangling`.
        - [x] Expand FK fields
    - [x] PATCH
        - [x] Only allow certain fields to be updated.
- [ ] Write tests
    - [x] Unauthenticated
    - [x] List
        - [x] Query parameters
            - [x] Filter fields
            - [x] Search field
        - [x] Bad values
    - [x] Retrieve (detail)
        - [x] Bad values
    - [x] Create
        - [ ] Bad values
            - [x] `expired_at` format
            - [ ] `expired_at` not in future
            - [x] `expired_at` set for kick/warn
            - [x] `type` choice
            - [x] unknown `user`
            - [x] `hidden` `True` for superstar
    - [x] Partial update
    - [x] Expansion
- [x] Document the endpoints

Anything else?


api/models.py
```diff
@@ -228,3 +229,54 @@ class Tag(ModelReprMixin, models.Model):
...
+    active = models.BooleanField(
+        default=True,
+        help_text="Whether the infraction is still active."
+    )
+    user = Member(
+        help_text="The user to which the infraction was applied."
+    )
+    actor = Member(
+        help_text="The user which applied the infraction."
+    )
```
jchristgit, line 276: These two don't really make sense to me. I'm certain you need a `models.ForeignKey` instead.

api/models.py
```diff
@@ -228,3 +229,54 @@ class Tag(ModelReprMixin, models.Model):
...
+        ("mute", "Mute"),
+        ("ban", "Ban"),
+        ("kick", "Kick"),
+        ("superstar", "Superstar")
+    )
+    id = models.UUIDField(
+        primary_key=True,
+        default=uuid.uuid4,
+        help_text="The UUID of the infraction."
+    )
```
jchristgit: You can omit this and Django will generate an ID field. Why UUIDs, anyways?

api/models.py
```diff
@@ -228,3 +229,54 @@ class Tag(ModelReprMixin, models.Model):
...
+    inserted_at = models.DateTimeField(
+        auto_now_add=True,
+        help_text="The date and time of the creation of this infraction."
+    )
+    expires_at = models.DateTimeField(
+        null=True,
+        help_text=(
+            "The date and time of the expiration of this infraction."
+            "Null if the infraction is permanent or it can't expire."
+        )
```
jchristgit, line 259: A validator for this being in the future might make sense.

api/models.py
MarkKoz: The comment for the old schema said it's a UUID.

api/models.py
```diff
@@ -228,3 +229,54 @@ class Tag(ModelReprMixin, models.Model):
...
+        help_text="The user to which the infraction was applied."
+    )
+    actor = Member(
+        help_text="The user which applied the infraction."
+    )
+    type = models.CharField(
+        max_length=9,
+        choice=TYPE_CHOICES,
+        help_text="The type of the infraction."
+    )
```
jchristgit, line 281: maybe we should use an enum field for this. might be overkill

api/serializers.py
```diff
@@ -13,6 +13,14 @@
...
         model = DocumentationLink
+        fields = (
+            'id', 'inserted_at', 'expires_at', 'active', 'user', 'actor', 'type', 'reason', 'hidden'
+        )
+
+
+class InfractionSerializer(ModelSerializer):
+    class Meta:
+        model = Infraction
         fields = ('package', 'base_url', 'inventory_url')
```
jchristgit: I think something went wrong here.

api/models.py
MarkKoz: Did you read the description of the PR? I have concerns about using a FK for the `user` particularily.

api/models.py
jchristgit: Well, but that's NoSQL... I don't think using a regular ID would hurt here.

api/serializers.py
MarkKoz: Yeah, I mixed up the serialisers!

jchristgit:
Looks solid so far. Thanks for looking into it!

> I'm unsure of how to handle the user and actor fields. I do not think a FK can simply be made for user given that infractions can be added for users that have never been in the server. For actor, I think a FK is fine.

Just use a nullable foreign key for the user, and document it in the `help_text`.

api/models.py
jchristgit: See my comment on the PR.

MarkKoz:
What would the serialisation be like if `user` is nullable? Doesn't that mean that users not found in DB will simply end up `null` in the infractions?

api/models.py
MarkKoz: In the sense that it shouldn't allow an expiration for certain types that can't expire (e.g. kick or warn)?

MarkKoz: Oh I misread, but I think my idea is good.

~~Also, why would we validate if an expiration is in the future. Isn't that the point of an expiration - to be at a later date?~~ Ok, I misread twice. Sorry.

jchristgit:
Yes

MarkKoz:
Wouldn't we want to still track which user the infraction is for?

api/models.py
jchristgit: > In the sense that it shouldn't allow an expiration for certain types that can't expire (e.g. kick or warn)?
 
Yeah, the expiry should either be `None` or a datetime in the future, IMO.

> Also, why would we validate if an expiration is in the future. Isn't that the point of an expiration - to be at a later date?
 
Yes, that's kind of what I want validations to cover: impossible states. After all, the Django API is the gateway to our database, and we aren't placing the constraints directly on the database (we probably should, but that's out of scope for now). I also added validators that check whether snowflakes are positive and permissions are between 0 and the max permission bitset, for instance.

However, using a validator like that will pretend us from importing infractions from the old site to the new site, so we probably shouldn't do that, or only put it in later.

jchristgit: So for now, I think this is fine.

api/models.py
```diff
@@ -1,3 +1,4 @@
+import uuid
```
jchristgit: You can probably drop this now

jchristgit:
Well, the user situation is still something we need to decide on. Do we delete users who leave the server? Do we add a field, `on_guild`, that determines whether they are part of the server?

jchristgit:
I think in the current setup we always have the users in the database. Not 100% certain though.

MarkKoz:
My point is that Discord and our moderation commands support things like banning by ID. This means the user could have never visited the server and still be banned from it.

Would it be possible to serialise an "incomplete" `Member` if one doesn't exist? I'd imagine this would involve custom serialisation and changing more of the `Member`'s fields to `null=True`. Maybe it's not worth the trouble.

jchristgit:
I don't think the serialization is our issue here - maybe you're mistaking something. Serialization and deserialization is what happens between the HTTP API and the viewsets, not the database, I think the database model is the issue here. The incomplete / "shadow" member idea is interesting, but that would also kinda hurt the data integrity. Hmmmmm

MarkKoz:
For what it's worth, we could get the user's info through [`discord.Client.get_user_info()`](https://discordpy.readthedocs.io/en/rewrite/api.html#discord.Client.get_user_info)

Sorry for the confusion with the terminology. I though it meant serialising into a model.

jchristgit:
Well, yes, sort of, okay, I think I'm confusing you here 😄 

For us, what matters about serializers is that they more or less just represent the data we have somehow. I think we're pretty flexible as far as this goes, but we need to come up with a smart solution on putting this on the database. A "user-id-that-is-maybe-foreign-key-and-maybe-not" sounds a bit ugly for my tastes, and it also doesn't play nice with Django's ORM.

MarkKoz:
If we can rely on the members in the database being up-to-date, then we can optionally pass the extra info from the function I linked when we POST the infraction and add the Member before going forward with the infraction. This may work since, on the bot side, I believe we can figure out when a user is unknown without having to even query the DB.

jchristgit:
That's a good idea. Let's do that. However, we will need to adapt our member model to work with it.

MarkKoz:
In that case, should we rename it to `User`? Maybe do that in another PR?

jchristgit:
Yeah, that sounds good. We will also need to think about how to organize role membership and actual guild membership then, so that should probably be done in another PR as you said.

MarkKoz:
I think it'd be much simpler to delegate the responsibility of adding a new user to the consumers of the API, but it may make the API feel more cumbersome to use.

MarkKoz:
I don't think the current endpoints for infractions are possible without a custom router. Is that true? If so, should we implement a custom one or just change the endpoints to use query params?

jchristgit:
> I think it'd be much simpler to delegate the responsibility of adding a new user to the consumers of the API, but it may make the API feel more cumbersome to use.

Yes, that's more or less what I'm currently trying to do, but `aiohttp` is causing some weird issues that I haven't quite figured out yet.

> I don't think the current endpoints for infractions are possible without a custom router. Is that true? If so, should we implement a custom one or just change the endpoints to use query params?

I don't think we should try to exactly replicate the current API in terms of routes. IMO, ideally all endpoints would work more or less the same without any weird quirks.

MarkKoz:
I was thinking of doing it with [DjangoFilterBackend ](https://www.django-rest-framework.org/api-guide/filtering/#djangofilterbackend), but it comes from another external dependency. Paired with a `SearchFilter`, I think it would cover all the query functionality of the endpoints and it'd be very easy to implement. Only thing left would be the `expand` for `User`s and whatever `dangling` is.

MarkKoz:
Oh wow, there's a [thing](https://github.com/silverlogic/djangorestframework-expander) for expansion too.

jchristgit:
I think the expansion thing is more or less built-in now, but I'm not 100% sure. There are various settings that control how relationships are serialized.

The django filter backend thing looks great, it looks very maintained and stable from a quick glance. I personally wouldn't mind adding this as a dependency as it provides really useful functionality and doesn't bring any extra dependencies.

MarkKoz:
Great, I'll go with `DjangoFilterBackend`.

I didn't realise the expansion thing was quite dated. I'll do some more research into that after I implement the filters; it's not a priority. Meanwhile, any pointers regarding resources on expansion are appreciated :)

api/tests/test_infractions.py
```diff
@@ -75,3 +76,96 @@ def test_retrieve_single_from_id(self):
...
+            'actor': self.user.id,
+            'type': 'ban',
+            'reason': 'He terk my jerb!',
+            'hidden': True,
+            'expires_at': '5018-11-20T15:52:00+00:00'
+        }
+
+        response = self.client.post(url, data=data)
+        self.assertEqual(response.status_code, 201)
+
```
MarkKoz, line 183: JSON response isn't being tested here because I'm unsure of how to deal with optional/auto fields. Furthermore, the date format in the response is not consistent with the request's.

api/tests/test_infractions.py
```diff
@@ -0,0 +1,171 @@
...
+    def test_list_all(self):
+        url = reverse('bot:infraction-list', host='api')
+        response = self.client.get(url)
+
+        self.assertEqual(response.status_code, 200)
+        infractions = response.json()
+
+        self.assertEqual(len(infractions), 2)
+        self.assertEqual(infractions[0]['id'], self.ban_hidden.id)
+        self.assertEqual(infractions[1]['id'], self.ban_inactive.id)
```
MarkKoz, line 74: Is it fine to check only IDs for the tests in this class or should I still explicitly check each field? I do check all fields in `CreationTests`.

api/tests/test_infractions.py
```diff
@@ -75,3 +76,96 @@ def test_retrieve_single_from_id(self):
...
+            'reason': 'He terk my jerb!',
+            'hidden': True,
+            'expires_at': '5018-11-20T15:52:00+00:00'
+        }
+
+        response = self.client.post(url, data=data)
+        self.assertEqual(response.status_code, 201)
+
+        infraction = Infraction.objects.get(id=1)
+        self.assertEqual(infraction.expires_at, dt.fromisoformat(data['expires_at']))
```
MarkKoz: `fromisoformat()` is 3.7 only. What's a good alternative to this? The two strings differ in format if I use `isoformat()`.

api/viewsets.py
```diff
@@ -97,6 +97,24 @@ class InfractionViewSet(CreateModelMixin, RetrieveModelMixin, ListModelMixin, Ge
...
+        data = {k: v for k, v in request.data.items() if k not in self.frozen_fields}
+
+        serializer = self.get_serializer(instance, data=data, partial=True)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+
+        if getattr(instance, '_prefetched_objects_cache', None):
+            # If 'prefetch_related' has been applied to a queryset, we need to
+            # forcibly invalidate the prefetch cache on the instance.
+            instance._prefetched_objects_cache = {}
```
MarkKoz: It's complaining this line isn't covered. The majority of the code for `partial_update` is copied from `UpdateModelMixin`. I could `# pragma: no cover`, but is there a "proper" way to fix this?

MarkKoz: I'm not applying `prefetch_related` anywhere myself, so could this be removed? Or could it be getting applied behind the scenes? Not too clear on what this is anyway.

api/models.py
```diff
@@ -284,3 +284,6 @@ class Infraction(ModelReprMixin, models.Model):
         default=False,
         help_text="Whether the infraction is a shadow infraction."
     )
+
+    def __str__(self):
+        return self.type
```
jchristgit: I think this should be a bit more descriptive, I'm not 100% sure about whether this is shown in the admin or whether it's `__repr__` but the type on its own doesn't really tell much. Maybe something like
```py
return f"#{self.id}: {self.type} on {self.user or 'unknown user'}"
```
and maybe some more magic with active, or expiry?

api/tests/test_infractions.py
jchristgit: No, this is enough.

api/viewsets.py
```diff
@@ -92,6 +94,9 @@ class DocumentationLinkViewSet(
 class InfractionViewSet(ModelViewSet):
     serializer_class = InfractionSerializer
     queryset = Infraction.objects.all()
+    filter_backends = (DjangoFilterBackend, SearchFilter)
+    filter_fields = ('user__id', 'actor__id', 'active', 'hidden', 'type')
+    search_fields = ('$reason',)
```
jchristgit, line 197: Do you think we should add a GIN index on this field? That could help us tremendously with infractions search. It might be better off to hold this off until later, though. See https://www.postgresql.org/docs/11/textsearch-tables.html#TEXTSEARCH-TABLES-INDEX.

jchristgit (commented):
Looks great. A few more things.

api/viewsets.py
```diff
@@ -97,6 +97,24 @@ class InfractionViewSet(CreateModelMixin, RetrieveModelMixin, ListModelMixin, Ge
     filter_backends = (DjangoFilterBackend, SearchFilter)
     filter_fields = ('user__id', 'actor__id', 'active', 'hidden', 'type')
     search_fields = ('$reason',)
+    frozen_fields = ('id', 'inserted_at', 'type', 'user', 'actor', 'hidden')
+
+    def partial_update(self, request, *args, **kwargs):
+        instance = self.get_object()
+
+        # Ignore updates for frozen fields.
+        data = {k: v for k, v in request.data.items() if k not in self.frozen_fields}
```
jchristgit: I think it would make sense to allow editing `hidden`, and throw an exception if a request tries to update one of the fields here, or does that mess with the existing update scheme?

api/viewsets.py
jchristgit: I think we can remove this.

api/tests/test_infractions.py
jchristgit: That's okay, I don't think we need to test DRF internals like this anyways

api/models.py
MarkKoz: I agree. The type was just a temporary value to use as I was unsure of what to put there.

api/viewsets.py
MarkKoz: Editing `hidden` doesn't make sense to me and I think the bot doesn't have an edit command for it either. This is because `hidden` is only meant to prevent DMing the user initially. Editing it afterwards will not DM the user (not currently, at least).

MarkKoz: An exception could be raised instead of ignoring them. Presumably a 400? Look like I'd need to subclass `APIException` for that, right?

api/viewsets.py
MarkKoz: Sure, sounds like a good idea.

MarkKoz:
For simplicity's sake, the `note` type should just be replaced with `warning` and distinguished by `hidden`. Otherwise, `note` would be the only field I would have to set a default of `True` (which I was planning on doing by overriding `save()` in the model). It would also involve further validation when inserting data (`warning` can't be hidden, `note` must be hidden). Thoughts?

I'm probably going to keep `note` for now and do the validation for it; I can remove it after if we decide against it.

api/viewsets.py
jchristgit: Doesn't this influence the infractions count the user can see about himself?

MarkKoz: I did not know it was even possible for a user to do that.

jchristgit: Oh, I thought it was. But you're right in that there's currently no way to do this via the bot.

jchristgit:
> validation > expiry date > Is in the future

I don't think we should validate this for now, considering it's going to mess with the import of the current infractions.

MarkKoz:
Yeah, I remember you saying that before. I just left it in there as a reminder.

MarkKoz:
So here's what's going on with expansion:

Specifying `depth` causes the field to change from `PrimaryKeyRelatedField` to a read-only `NestedSerializer`. Upon validation, read-only fields are removed from the data that will be serialised. This means the user and actor are removed and the not null constraints on their FKs are tripped.

It is possible for a `NestedSerializer` to be writeable, but it would mean custom a `create()` method. I didn't understand how to fix things with that approach.

I'm working on a solution that looks promising, though perhaps a bit ugly. It involves overriding `to_representation`. This isn't that bad on its own, but it may get ugly trying to make the expansion optional.

MarkKoz:
So it isn't that ugly after all; the try-finally blocks were not needed. I was a dummy and mixed up my breakpoints which lead to false data and the impression I needed to change back `serializer_class`.

MarkKoz:
I still need to document the extended endpoints, but was not sure how to do that nicely while avoiding repetition. Let me know if you have any advice on that and if you disagree with the current style of the docstring.

api/serializers.py
```diff
@@ -16,6 +16,42 @@ class Meta:
...
+        return attrs
+
+
+class ExpandedInfractionSerializer(InfractionSerializer):
+    def to_representation(self, instance):
+        ret = super().to_representation(instance)
+
+        user = User.objects.get(id=ret['user'])
+        user_data = UserSerializer(user).data
+        ret['user'] = user_data
```
jchristgit, line 46: I think this needs to account for the user being None, doesn't it?

MarkKoz: No. User is not nullable.

jchristgit (commented):
I like it. What's missing? :shipit:

MarkKoz:
Just the docs for extended endpoints. I'll try to get that done tomorrow.

MarkKoz:
I think this is ready to be merged.

jchristgit:
This looks very good. One more thing: Since all the other endpoints use it, do you think we should name the endpoint `infractions` instead of `infraction`?

jchristgit (approved):
Very solid. Thank you!

jchristgit (approved):
Very solid. Thank you!

PR #150, opened 2018-11-19
Django - Add Support for Storing Users Not in Guild

MarkKoz:
As discussed in  #149, we need the `Members` (now renamed to `Users`) to facilitate users not in our guild.

I think a simple boolean field is all that is needed. The only field that is unique to members is `roles`, but that may be valuable to keep track of in the event a member leaves the guild and rejoins. I made `roles` not required, but I am not sure if that is the best thing to do.


api/admin.py
```diff
@@ -1,7 +1,7 @@
 from django.contrib import admin
 
 from .models import (
-    DocumentationLink, Member,
+    DocumentationLink, User,
```
jchristgit: Doesn't cornflakes complain about this?

api/admin.py
```diff
@@ -10,7 +10,7 @@
 
 
 admin.site.register(DocumentationLink)
-admin.site.register(Member)
+admin.site.register(User)
```
jchristgit: Would be nice to keep this in order.

api/serializers.py
```diff
@@ -3,7 +3,7 @@
 
 from .models import (
     DocumentationLink,
-    Member, OffTopicChannelName,
+    User, OffTopicChannelName,
```
jchristgit: the cornflakes they will shutter

api/tests/test_models.py
```diff
@@ -1,7 +1,7 @@
 from django.test import SimpleTestCase
 
 from ..models import (
-    DocumentationLink, Member, ModelReprMixin,
+    DocumentationLink, User, ModelReprMixin,
```
jchristgit: the f̤͍͍́l͖̣͙̥ạke͇̳̘̤s͏͍͈̝̳ must n̷o̕t be a̧͞ǹ̴g̨̀͝ere͘d̵͟

api/tests/test_users.py
```diff
@@ -1,7 +1,7 @@
 from django_hosts.resolvers import reverse
 
 from .base import APISubdomainTestCase
-from ..models import Member, Role
+from ..models import User, Role
```
jchristgit: a͢͏͟a̛͞ąa̕aa help

api/urls.py
```diff
@@ -3,7 +3,7 @@
 
 from .views import HealthcheckView
 from .viewsets import (
-    DocumentationLinkViewSet, MemberViewSet,
+    DocumentationLinkViewSet, UserViewSet,
```
jchristgit: the f̴͡͞l͘a̶ḱe͡͞͝man h̐ͤ̍̃ͯͩͭê̵̑̿ͣͨ ̡̌̉c̀́oͭ́ͫ̒ͭ̓̃͞m̡̅̑ͦ̄͒͗e̢ͣ͆̒ͭ̅́s̛̅

api/viewsets.py
```diff
@@ -10,15 +10,15 @@
 from rest_framework_bulk import BulkCreateModelMixin
 
 from .models import (
-    DocumentationLink, Member,
+    DocumentationLink, User,
```
jchristgit: aaaaaaaa

api/viewsets.py
```diff
@@ -10,15 +10,15 @@
...
     OffTopicChannelName,
     SnakeFact, SnakeIdiom,
     SnakeName, SpecialSnake,
     Tag
 )
 from .serializers import (
     DocumentationLinkSerializer,
-    MemberSerializer, OffTopicChannelNameSerializer,
+    UserSerializer, OffTopicChannelNameSerializer,
```
jchristgit: I think you know the drill

jchristgit (commented):
Looks good so far. I think the flakes will be mad, though.

Do you think we should have `roles` be `NULL` if `not is_member`?

MarkKoz:
Right now, `roles` will just be an empty list. `null=True` has no effect on many to many relationships.

jchristgit:
Ah, good point, I forgot about that.

jchristgit (approved):
Looks solid. Thanks for the PR!

PR #164, opened 2019-01-02
Add traceback to info help

SebastiaanZ:
Tracebacks are often a vital part of the help process on our server. Unfortunately, users often exclude vital parts of the traceback when formulating their questions and most beginners feel overwhelmed when confronted by tracebacks after running their code. This PR intents to help us in that process by adding information on how to read tracebacks to the help-page, telling the asker to inspect the traceback before asking a question, and telling them to include the full traceback in their question if their code raises an exception.

From top to bottom, these are the proposed changes to the page:

- Adding traceback inspection to the **Before you ask** section:

Before you ask

- Adding the traceback to things you need to have at hand in **A good question**:

A good question

- A new section called "Examining Tracebacks":

Examining Tracebacks Part I

(Continued with small overlap:)

Examining Tracebacks Part II

(Continued with small overlap:)

Examining Tracebacks Part III

### DIFF
--- a/templates/main/info/help.html
+++ b/templates/main/info/help.html
@@ -45,6 +45,7 @@
                 <ul>
                     <li>Read the official documentation for whatever you're working with</li>
                     <li>Use a debugger to inspect your code</li>
+                    <li>Examine the traceback when your code raises an exception</li>
                     <li>Do some research online - for example, on Stack Overflow</li>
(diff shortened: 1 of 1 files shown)


jchristgit (approved):
Very useful. I like it.

scragly (approved):
Looks like a good addition. Nice stuff.

PR #166, opened 2019-01-03
WIP Site requirements for starboard

tagptroll1:
Table and endpoints for bots starboard features, WIP.

python-discord/bot#238


pysite/views/api/bot/starboard.py
```diff
@@ -13,11 +13,11 @@
...
-    "bot_message_id": str,
-    "message_id": str,
-    "guild_id": str,
-    "channel_id": str,
-    "author_id": str,
+    "bot_message_id": int,
+    "message_id": int,
+    "guild_id": int,
+    "channel_id": int,
+    "author_id": int,
```
jchristgit: are you sure about this? I think we used strings here because rethink had some issues storing snowflakes as ints. I could be wrong though. But this should definitely be checked - maybe @heavysaturn or @jos-b know?

pysite/views/api/bot/starboard.py
```diff
@@ -110,5 +115,28 @@ def delete(self, data):
...
+        return "", 400
+
+
+class StarboardDeletionView(APIView, DBMixin):
+    path = "/bot/starboard/delete"
+    name = "bot.starboard.delete"
+    table_name = "starboard_messages"
+
+    @api_key
+    @api_params(schema=DELETE_ALL_STARBOARD, validation_type=ValidationTypes.params)
```
jchristgit: I think you can omit the validation altogether if you don't take any params anyways?

jchristgit (dismissed):
important comment

pysite/views/api/bot/starboard.py
tagptroll1: I believe i had it as strings to begin with, but found it easier to work with ints.  But if they don't work with rethink i can change them back

jb3: Yeah we definitely had issues with RethinkDB not being able to store snowflakes, these should probably be strings.

MrHemlock:
Has been open for a while and it doesn't seem like it'll make the cutoff time for when the Django rewrite finishes.  It can be implemented once the Django stuff is completed in a new PR.

PR #168, opened 2019-01-04
Django: Documentation improvements

delterr:
Improvements for development documentation.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -26,6 +26,9 @@ in your environment variables:
 export DATABASE_URL=postgres://pysite@localhost/pysite
 ```
 
+After this step, inside the `.env` file, set the `SECRET_KEY` variable which can be anything you like. Also, set the `ENGINE` variable
+to `django.db.backends.postgresql`.
+
 A simpler approach to automatically configuring this might come in the
 near future - if you have any suggestions, please let us know!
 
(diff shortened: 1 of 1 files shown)


docs/setup.md
```diff
@@ -26,6 +26,9 @@ in your environment variables:
 export DATABASE_URL=postgres://pysite@localhost/pysite
 ```
 
+After this step, inside the `.env` file, set the `SECRET_KEY` variable which can be anything you like. Also, set the `ENGINE` variable
+to `django.db.backends.postgresql`.
+
```
jchristgit, line 31: `ENGINE` is configured in the settings already. Why would you export it?

docs/setup.md
```diff
@@ -67,5 +70,24 @@ pip install -e .[lint,test]
...
+127.0.0.1   pythondiscord.local
+127.0.0.1   api.pythondiscord.local
+127.0.0.1   staff.pythondiscord.local
+127.0.0.1   admin.pythondiscord.local
+127.0.0.1   wiki.pythondiscord.local
+127.0.0.1   ws.pythondiscord.local
+```
+When trying to access the site, you'll be using the domains above instead of the usual `localhost:8000`.
+
+To run the server, run `DEBUG=1 python manage.py runserver`. If it gives you an error saying
```
jchristgit, line 89: DEBUG should be set in the environment file - you won't get the local allowed hosts otherwise and  and I also recommend setting LOG_LEVEL=INFO

jchristgit (changes requested):

jchristgit:
These have been merged to the Django branch via cherry-picks and adjustments on another branch. Thanks!

PR #182, opened 2019-01-20
Remove RMQ, add Discord webhook support.

### LINKED ISSUE #174: Site status log messages to stop using RMQ and use Webhooks instead.
With RMQ being removed, the current method of site status logs being sent via the bot for then sending to a discord channel won't be functionable. 

As an alternative, and to avoid the reliance on bot being alive at the time of a status update, a basic webhook should be setup in the site for updates to be sent directly to a discord channel. This should be more reliable and less complicated.

scragly:
Closes #174 

This entirely removes RMQ from the current Site, replacing any the message send events with Discord Webhooks that are grabbed from environmental variables.

The role adding and removal events that were used for code jams have been simply removed for now, so if the site is used for jams sign ups in the future before the django rewrite is deployed, roles will need to be manually added and removed when members are approved/unapproved.

The webhook envs that can be set are:
`WEBHOOK_DEVLOG`
`WEBHOOK_MODLOG`
`WEBHOOK_USERLOG`
`WEBHOOK_JAMLOG`

Currently `WEBHOOK_USERLOG` is not used in Site so can be omitted in production if wanted. The other three are used.

Wiki change messages have been moved to `#dev-logs` instead of `#mod-logs` to keep messages relevant and clean.

There may be special considerations for the production setup, as it's slightly different regarding deployment. Bot and snekbox will still require RMQ for the moment, so we don't wish to remove RMQ from production quite yet.


lemonsaurus:
As discussed in todays staff meeting, we should close this PR and reopen it after the django rewrite, where much of this may already have been done.

PR #199, opened 2019-04-07
Add an API to post logs to.

jchristgit:
As promised. I ALWAYS DELIVER

However this is draft because I still want some suggestions on fields. Is this too much? Is it sufficient?


MarkKoz:
Can you provide some context? What's the plan for using this API? So far I have gathered that the model has no field for the log's actual message and that the API only allows for creation, not retrieval.

jchristgit:
Hello @MarkKoz thanks for the review for some reason I forgot that log lines tend to contain messages. The new commit accounts for this. The idea is that this will serve as our internal logging platform, this is the corresponding API to store logs. We will later build a UI in the web app to browse these, but we'll get there in steps.

MarkKoz (commented):
Is the plan for messages to just post them as plain text? What about exceptions? It may be annoying to parse out the exceptions should we choose to do so.

jchristgit:
Yes, messages will be plain text. Parsing out exceptions might be overthinking it a bit, we're not remaking Sentry after all

jerbob (approved):
Looks good to me.

Issue #200, opened 2019-04-08
Django: Wiki


gdude2002:
Issue #142 states that the wiki should be moved entirely to GitHub, but the decision has since been made to integrate a wiki with the new Django site.

We will be using [django-wiki](https://github.com/django-wiki/django-wiki) for this task, and we'll customize its templates so that it integrates with the rest of the site.

Before doing this, setting up the basic site templates is probably a good idea.

lemonsaurus:
This has been merged in.

Issue #201, opened 2019-04-08
Django: Oauth integration


scragly:
There's a need to integrate Discord OAuth for site account creation, linking and authorization, using the default Django Auth system so it's ubiquitous throughout all the site and any apps we use.

An existing solution was found with Django-Allauth that might suit the job and includes both Discord and Github as supported providers:
https://django-allauth.readthedocs.io/en/latest/index.html

The Github integration would be a great bonus as they could be utilised with the Code Jams.

lemonsaurus:
I'm a fan of using allauth for this. It looks easy enough to implement, and would play well with both the overall page and the wiki.

What we'd want to do is make is so that, on first OAuth login with Discord, a django auth user is created. We'd then need to somehow figure out the Discord username, use this to check what the users roles are, and use the roles to determine which usergroup to put the user in. A user who was a moderator would get a moderator usergroup, and so on.

We don't need one usergroup for every single role on the server, just major ones. Right off the bat I'm thinking Developer, Contributor, Helper, Moderator, Admin, Owner. This will allow us to restrict editing and reading of certain pages to these roles, as well as eventually using the same usergroups to hide more interactive parts of the new website behind, such as editing tags, reviewing staff applications or ban appeals, that sort of thing.

Admins and Owners should both have full access to the site in every way.

jchristgit:
Sounds good. I will do it.

lemonsaurus:
Stale, unassigning @jchristgit

gdude2002:
I'll work on this.

gdude2002:
A few questions:

* Do we want to allow users to sign in with a username/email and password?
* Similarly, what about signing up? Social only?
* Do we want to fully replace the Django Admin login with this system?
* Where should the link to the accounts system be? (Login, my settings, etc)

scragly:
I'd prefer to avoid standard logins and rely on both account creation and logins being reliant on oauth.

The django admin login should use this system, but is there a way to make sure that there's a way to still use the generic admin account as a backup still, just in case something goes wrong with oauth.

I don't know enough about the django login system to be able to comment on the last point.

gdude2002:
> The django admin login should use this system, but is there a way to make sure that there's a way to still use the generic admin account as a backup still, just in case something goes wrong with oauth.
> 
> I don't know enough about the django login system to be able to comment on the 
last point.

[...]

PR #202, opened 2019-04-08
[#200] Django Wiki

gdude2002:
This is an early PR just to keep track of the progress on the wiki. I'll update the title when it's ready to go!

Some notes:

* The wiki will be at `/wiki` on the base domain, not a `wiki` subdomain. This overall seems to be preferential anyway, but I was also having trouble actually getting `django-wiki` to work at all on another host.
* I've included the Monokai Pygments theme for code block highlighting. The main reason for this is that it might be useful to have it accessible to other parts of the site - not just the wiki.
* This could be a long PR - there's a lot of template work to do. It may or may not also be blocked by the lack of a working login system, but we'll see how we go.


gdude2002:
And with that, the PR is ready for review!

gdude2002:
There's quite a lot of stuff in this PR, so here's a summary:

* Changes to `settings.py`:
    * Added a bunch of stuff that the wiki requires
    * `LOGIN_URL` - this will need to be edited when an auth system is in place
    * Updated `PARENT_HOST` env var so it can be used in DEBUG mode and will be added to `ALLOWED_HOSTS` in that scenario
    * A few accessibility changes to the Bulma variables
    * Django's media storage has been set up - that'll be found under `media/` in the root
* Wiki
    * Fully implemented
    * There are a few things that don't look fully polished - these are purely on the backend, and can be addressed later if needs be
    * Let me tell ya: This thing was a damn pain. Lots of weird decisions. Lots of odd code. But it works!
        * [me_irl](https://youtu.be/rvrZJ5C_Nwg?t=127)
* Dependencies
    * Re-locked the Pipfile, since it had to be merged
    * Added `wiki` and `pygments`

pydis_site/settings.py
```diff
@@ -255,5 +279,49 @@
     "variables": {
         "primary": "#7289DA",
         "link": "$primary",
+
+        "green": "#21c65c",  # Better contrast with the light text
```
lemonsaurus: Not sure it's necessary to justify this change with an inline comment.

pydis_site/settings.py
```diff
@@ -255,5 +279,49 @@
     "variables": {
         "primary": "#7289DA",
         "link": "$primary",
+
```
lemonsaurus, line 283: not a big fan of these mid-dict newlines. I mean, green is a color, the previous two are colors, they shouldn't even be in separate groups as far as I can reason?

pydis_site/settings.py
```diff
@@ -255,5 +279,49 @@
...
+            "baselevel": 2
+        }
+    }, "extensions": [
+        "markdown.extensions.abbr",
+        "markdown.extensions.attr_list",
+        "markdown.extensions.extra",
+        "markdown.extensions.footnotes",
+        "markdown.extensions.nl2br",
+        "markdown.extensions.sane_lists",
+
```
lemonsaurus, line 313: this newline, however, seems fine. logical groupings of 5+ items like that. it just looks weird in the other one, though.

pydis_site/settings.py
```diff
@@ -255,5 +279,49 @@
...
+        "wiki.core.markdown.mdx.codehilite",
+        "wiki.core.markdown.mdx.previewlinks",
+        "wiki.core.markdown.mdx.responsivetable",
+        "wiki.plugins.macros.mdx.toc",
+        "wiki.plugins.macros.mdx.wikilinks",
+    ]
+}
+
+WIKI_MESSAGE_TAG_CSS_CLASS = {
+    messages.DEBUG: "",  # is-info isn't distinctive enough from blurple
```
lemonsaurus, line 323: wait, so what does this do exactly? How will debug messages look when there's nothing here?

pydis_site/static/js/wiki/dropdown.js
```diff
@@ -0,0 +1,37 @@
+// TODO: Move to django-simple-bulma
```
lemonsaurus: Maybe instead of leaving this comment in, you could make an issue over on that repo and just stick this code right in the issue?

LICENSE-SimpleMDE
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
jchristgit, line 21: this is from the markdown editor, right? shoudln't it reside in the static/js folder instead of top-level? otherwise we will probably end up with dozens of licenses t some point

pydis_site/templates/base/base.html
```diff
@@ -1,6 +1,5 @@
 {# Base template, with a few basic style definitions. #}
-{% load django_simple_bulma %}
-{% load static %}
+{% load django_simple_bulma sekizai_tags static %}
```
lemonsaurus: I think I prefer the readability of splitting this into three loads. I feel like the python equivalent to this would be 

```py
import django_simple_bulma, sekizai_tags, static
```

and we discourage that, for good reasons.

pydis_site/templates/wiki/base.html
```diff
@@ -0,0 +1,49 @@
+{% extends "base/base.html" %}
+{% load sekizai_tags i18n wiki_tags static %}
```
lemonsaurus: same applies here - multiple loads will be more readable.

Pipfile
```diff
@@ -16,16 +16,18 @@ coverage = "~=4.5.3"
 unittest-xml-reporting = "~=2.5.1"
 
 [packages]
-django = "~=2.2"
+django = "~=2.1"
```
jchristgit: Not a fan of this at all. I'm also not sure if this can break things for us if we port back up to 2.2 later. Also, 2.2 is LTS, so it's a good version to stay on.

Looks like django-wiki depends on < 2.2, but I'm not sure if this is an actual "it breaks on 2.2" or "we didn't update the version specifier yet" - judging from [this commit](https://github.com/django-wiki/django-wiki/commit/018074b62800e2ec91abc6e3cb36f73bc1d4f6c1) they didn't add it to the test matrix yet and maybe could theoretically support it. Maybe we can send in a PR to add 2.2 to the test matrix? I can look into that later today.

pydis_site/apps/home/templatetags/wiki_extra.py
```diff
@@ -0,0 +1,75 @@
...
+def render(template_path: str, context: Dict[str, Any]):
+    return mark_safe(get_template(template_path).render(context))  # noqa: S703 S308
+
+
+@register.simple_tag
+def render_field(field: Field, render_labels: bool = True):
+    if isinstance(field, BoundField):
+        unbound_field = get_unbound_field(field)
+    else:
+        unbound_field = field
```
jchristgit: do you need the extra `isinstance` check here? I feel like this entire block could just be
```py
unbound_field = get_unbound_field(field)
```
instead.

pydis_site/templates/wiki/base.html
```diff
@@ -0,0 +1,49 @@
...
+        {% wiki_messages %}
+
+        <div class="content">
+          {% block wiki_contents %}{% endblock %}
+        </div>
+      </div>
+    </section>
+  {% endblock %}
+
+  {% block site_footer %}
```
lemonsaurus: FYI: you may need to remove these on a lot of templates because in order to do a push-footer thing where it would always be at the bottom of the screen even if the content doesn't fill the page, I moved the footer to the base template in my branch.

pydis_site/settings.py
```diff
@@ -255,5 +279,49 @@
...
+WIKI_ANONYMOUS = True
+WIKI_ANONYMOUS_WRITE = False
+
+WIKI_MARKDOWN_KWARGS = {
+    "extension_configs": {
+        "wiki.plugins.macros.mdx.toc": {
+            "anchorlink": True,
+            "baselevel": 2
+        }
+    }, "extensions": [
```
jchristgit, line 306: this should probably be on a newline

pydis_site/templates/wiki/base.html
```diff
@@ -0,0 +1,49 @@
...
+    {% include "base/footer.html" %}
+  {% endblock %}
+
+  <script src="{% static "js/wiki/dropdown.js" %}" type="text/javascript"></script>
```
lemonsaurus, line 44: any particular reason these are at the end of the file and not at the start? I know I've seen people do that as a sort of hack to overcome load-order related problems but if at all possible I'd love to see them in the head where they belong.

pydis_site/static/js/wiki/dropdown.js
jchristgit: this should probably be an issue on the django-simple-bulma repo?

pydis_site/templates/wiki/base_site.html
```diff
@@ -0,0 +1,170 @@
+{% extends "base/base.html" %}
+{% load sekizai_tags i18n wiki_tags static %}
+
+{# This template is no longer used - we can remove it later when the rest are done #}
```
lemonsaurus: okay.. the rest seem to be done. maybe it's time to remove it?

pydis_site/templates/wiki/base_site.html
```diff
@@ -0,0 +1,170 @@
...
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Wiki | {% block wiki_pagetitle %}{% endblock %}{% block wiki_site_title %}{% endblock %}</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="description" content="">
+    <meta name="author" content="www.django-wiki.org">
+
+    <!-- Le styles -->
```
lemonsaurus: nice. I can tell this one wasn't meant to be in the PR. :stuck_out_tongue:

pydis_site/templates/wiki/delete.html
```diff
@@ -0,0 +1,90 @@
...
+        <a href="{% url 'wiki:get' path=urlpath.path article_id=article.id %}" class="button is-primary">
+          <span class="icon">
+            <i class="fas fa-arrow-left"></i>
+          </span>
+          <span>Go back</span>
+        </a>
+      </div>
+    </article>
+  {% else %}
+    {% if cannot_delete_children %}
```
jchristgit, line 26: shouldn't these checks go over django's permission system or how does this work?

pydis_site/templates/wiki/includes/breadcrumbs.html
```diff
@@ -0,0 +1,87 @@
...
+            {% endif %}
+
+            <a class="dropdown-item" href="{% url 'wiki:dir' path=urlpath.path %}">
+              Browse other articles
+            </a>
+          </div>
+        </div>
+      </div>
+
+      {% if request.user.is_authenticated %}
```
jchristgit, line 52: i imagine not everybody should be able to create articles later? so we should make this an issue

pydis_site/templates/wiki/includes/render.html
```diff
@@ -0,0 +1,26 @@
+{% load wiki_tags i18n cache sekizai_tags static %}
```
jchristgit: same thing with multiline loads here

jchristgit (dismissed):
- i think the load template tags should be on separate lines for git blame retention and other reasons but would be curious what others think

pydis_site/templates/wiki/create.html
```diff
@@ -0,0 +1,43 @@
+{% extends "wiki/base.html" %}
+{% load wiki_tags i18n sekizai_tags static %}
```
lemonsaurus: multi-line loads please.

pydis_site/templates/wiki/create_root.html
```diff
@@ -0,0 +1,53 @@
+{% extends "wiki/base.html" %}
+{% load wiki_tags crispy_forms_tags i18n sekizai_tags static %}
```
lemonsaurus: multi-line loads please.

pydis_site/templates/wiki/delete.html
```diff
@@ -0,0 +1,90 @@
+{% extends "wiki/base.html" %}
+{% load wiki_tags sekizai_tags static %}
```
lemonsaurus: I'll stop mentioning these now, you'll obviously do all of them.

pydis_site/templates/wiki/history.html
```diff
@@ -0,0 +1,155 @@
...
+
+{% block wiki_contents_tab %}
+  {% include "wiki/includes/modals.html" %}
+
+  {% addtoblock "js" %}
+    <script type="text/javascript" src="{% static "wiki/js/diffview.js" %}"></script>
+    <script type="text/javascript" src="{% static "wiki/js/diff.js" %}"></script>
+  {% endaddtoblock %}
+  {% addtoblock "css" %}
+    <style type="text/css">
```
lemonsaurus: I'd prefer if this PR didn't have `<style>` tags either, much like how the js was moved into files, I suggest you do the same with the css.

pydis_site/templates/wiki/plugins/links/sidebar.html
```diff
@@ -0,0 +1,56 @@
...
+
+    The package.json states MIT - but there is no prose or license
+    text available for the project itself.
+  {% endcomment %}
+
+  <script src="https://cdn.rawgit.com/mattmezza/bulmahead/master/dist/bulmahead.bundle.js"></script>
+  <script src="{% static "js/wiki/links_sidebar.js" %}"></script>
+
+  <script type="text/javascript">
+    {# If you figure out a way to do this without putting some JS directly in the page... let me know #}
```
lemonsaurus: this comment is weird. Let's just agree that this is fine.

lemonsaurus (changes requested):
Whew. Big PR, but for the most part looks like high quality stuff.

If you've got that demo still running, feel free to put that link in a comment so people can play with it a little as they review. A couple of screenshots may have been nice for a PR this big, too.

LICENSE-SimpleMDE
gdude2002: I think we probably need a place for them, by convention they go in the root but if you can propose somewhere we could collect them, that'd work

Pipfile
gdude2002: I think there isn't much chance of getting that merged, but if you're hopeful, be my guest.

Could delay the PR quite a bit if we're waiting on it, though.

pydis_site/templates/wiki/delete.html
gdude2002: They do. That's a wiki thing.

pydis_site/settings.py
gdude2002: It's not an extremely different shade - it would feel weird to _not_ justify it, imo

pydis_site/settings.py
gdude2002: They use the default grey colour.

image

image

pydis_site/templates/wiki/base.html
gdude2002: That's the first reason people do it - the other is so that scripts that don't need to be loaded for the page to function don't block the rendering of the page.

We can use a JS event, though. It just means the page will take longer to render.

pydis_site/templates/wiki/base_site.html
gdude2002: Haha, wasn't even my comment. I'll get it.

pydis_site/templates/wiki/base.html
gdude2002: As long as your branch still has the block - I need to be able to remove the footer in the preview iframes

pydis_site/templates/wiki/includes/breadcrumbs.html
gdude2002: #201

pydis_site/templates/wiki/history.html
gdude2002: Actually, this isn't used, so I'll just remove it.

LICENSE-SimpleMDE
gdude2002: As discussed on Discord, we're going to solve this one when it becomes a problem.

pydis_site/templates/wiki/base.html
lemonsaurus: I'm not sure it does, but it's been merged now so you can find out.

pydis_site/templates/wiki/base.html
lemonsaurus: naw, fuck it, leave it as is.

Pipfile
gdude2002: From Discord: We're not gonna wait to solve this before merging the PR, but it's something that should be prioritised after.

lemonsaurus: huh? which discord is this from? If you were paraphrasing me, you may have misunderstood. I absolutely do want Django 2.2 before we merge this, but I won't wait for months for the django-wiki people to merge a PR, though. If we have to, we will temporarily fork the wiki and change it ourselves.

gdude2002: Ah okay, fair

gdude2002:
Reviews addressed and latest changes from `django` have been merged.

image

gdude2002:
image

lemonsaurus (approved):
I think we're good to go.

jb3 (approved):

Pipfile
lemonsaurus: This has now been resolved.

PR #203, opened 2019-04-10
Initial work on some base templates

gdude2002:
This is just some work to establish a basic style/layout. In order to figure out positioning/layout elements, I ended up partially recreating a mockup @lemonsaurus created, but this is of course modifiable later.

Since this is just a base work PR, I didn't try to exactly match the mockup.

image


pydis_site/templates/base/navbar.html
```diff
@@ -0,0 +1,110 @@
...
+
+      <a role="button" class="navbar-burger" aria-label="menu" aria-expanded="false" data-target="navbar_menu">
+        <span aria-hidden="true"></span>
+        <span aria-hidden="true"></span>
+        <span aria-hidden="true"></span>
+      </a>
+    </div>
+    <div class="navbar-menu is-paddingless" id="navbar_menu">
+      <div class="navbar-start">
+        {#                    {% if dropdowns %}#}
```
lemonsaurus: what's all this commented out template code for?

lemonsaurus (changes requested):
I think there should only be a single base template which doesn't do anything inside the body tags. the current approach with one hero base and one regular base isn't DRY (duplicate head and footer code).

Let's just define stuff like heroes inside the individual page templates themselves, at least for the time being. If we end up with a lot of duplicate code we can find a solution later on. Keeping

pydis_site/templates/base/navbar.html
gdude2002: Since I based these templates off of my usual template system/layout, I left the content from this navbar in the template so you'd have an example of how it's specifically meant to be used.

It can be removed if you don't need it.

lemonsaurus: yeah let's remove it.

lemonsaurus (approved):
Okay, good stuff. Let's get this merged.

PR #204, opened 2019-04-10
Only install uWSGI in the Docker image.

jchristgit:
As discussed in the dev channels.

Not sure why the psycopg2 hashes were updated.


docker/app/Dockerfile
```diff
@@ -28,7 +28,8 @@ RUN rm -r /opt/bitnami/python/lib/python3.*/site-packages/setuptools* && \
     pip install --no-cache-dir -U setuptools
 
 RUN python3 -m pip install pipenv \
-    && python3 -m pipenv install --dev --system --deploy
+    && python3 -m pipenv install --dev --system --deploy \
```
jerbob: Do we need to install dev dependencies in this image?

jchristgit: No, will fix in separate PR.

jchristgit: fixed

jerbob (approved):
cool and ok

PR #206, opened 2019-04-14
Update resources page

sco1:
Per https://github.com/python-discord/organisation/issues/18

* Add partnered Discord communities
* Replace IssueHaven with direct link to a GitHub "good first issue" (Python) search
* Move Python Cheat Sheet to misc
* Replace TheNewBoston with Sentdex
* Remove Modern Python 3 Bootcamp


lemonsaurus:
While I appreciate this, I feel like perhaps a partners page is something we've sort of outgrown. The sponsors (adafruit being one of these) are going to be front and center on the new page, but building a partners page to showcase stuff like RLBot seems like a pointless thing to spend time on, as does making more partner banners. This was a carrot we used to get new partners back when we were small, and needed to convince larger discord servers we were worth partnering up with.

So my suggestion is: 

- Keep the changes to the resources
- Remove the partners who are no longer partners
- Don't worry about adding the new ones.

jerbob (approved):

GhostofGoes (approved):

Issue #205, opened 2019-04-14
Django: Refactor models to use multiple files instead of a single monolith.


lemonsaurus:
In the `api` app, we currently have a `models.py` with like 25 classes in it and nearly 500 lines of code. This makes the relevant model far more annoying to find than it needs to be.

I think it's time to refactor this so that models go in a `models` folder with files.

jchristgit:
Sounds good. Also viewsets and serializers. I will do it.

PR #207, opened 2019-04-14
#205: split models into multiple modules

jchristgit:
As discussed in #205, this merge request refactors models to use one model per file instead of a single huge monolith.


pydis_site/apps/api/models/bot/infraction 3.py
```diff
@@ -0,0 +1,67 @@
+from django.db import models
```
lemonsaurus: this file is a duplicate of `infraction.py`, right?

lemonsaurus (changes requested):
looks good, but I think you left a duplicate file in there.

lemonsaurus (approved):

jchristgit:
Thanks!

PR #213, opened 2019-04-18
Django front page

lemonsaurus:
Initial draft of the new front page, including a completely refurbished navbar, GitHub API integration to list our projects, and screen-reader friendly footer.

This PR also includes minor fixes to make local dev using the docker-compose more pleasant, including a default admin user and a way to deal with `makemigrations`.

Site is fully responsive for tablets and mobile.

image


lemonsaurus:
I think I need to write a couple of tests and do some linting, I seem to have forgotten that. After that this should be good to merge.

docker/app/migrate_and_serve.sh
```diff
@@ -9,7 +9,12 @@
 
 echo [i] Applying migrations.
 python manage.py migrate --verbosity 1
+
 echo [i] Collecting static files.
 python manage.py collectstatic --no-input --clear --verbosity 0
+
+echo [i] Creating a superuser
+echo "from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser('admin', 'admin', 'admin')" | python manage.py shell
```
SebastiaanZ: Unfortunately, this will fail the second time you start the docker-compose, since the username already exists at that point. See trace back at the bottom.

This solves that and prints "Account already exists" when it already exists:

```suggestion
echo "from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser('admin', 'admin', 'admin') if not User.objects.filter(username='admin').exists() else print('Username already exists')" | python manage.py shell
```

The full traceback you'll get otherwise:

```py
web_1       | [i] Creating a superuser
postgres_1  | 2019-04-18 13:23:51.787 UTC [26] ERROR:  duplicate key value violates unique constraint "auth_user_username_key"
postgres_1  | 2019-04-18 13:23:51.787 UTC [26] DETAIL:  Key (username)=(admin) already exists.
postgres_1  | 2019-04-18 13:23:51.787 UTC [26] STATEMENT:  INSERT INTO "auth_user" ("password", "last_login", "is_superuser", "username", "first_name", "last_name", "email", "is_staff", "is_active", "date_joined") VALUES ('pbkdf2_sha256$150000$FoXDVt2Gbs85$gpB7NgN9RD7Z+R/AVOXawJW0Yoq1SoTXvSjB0s0AWnc=', NULL, true, 'admin', '', '', 'admin', true, true, '2019-04-18T13:23:51.591734+00:00'::timestamptz) RETURNING "auth_user"."id"
web_1       | Traceback (most recent call last):
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
web_1       |     return self.cursor.execute(sql, params)
web_1       | psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "auth_user_username_key"
web_1       | DETAIL:  Key (username)=(admin) already exists.
web_1       | 
web_1       | 
web_1       | The above exception was the direct cause of the following exception:
web_1       | 
web_1       | Traceback (most recent call last):
web_1       |   File "manage.py", line 21, in <module>
web_1       |     main()
web_1       |   File "manage.py", line 17, in main
web_1       |     execute_from_command_line(sys.argv)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
web_1       |     utility.execute()
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
web_1       |     self.fetch_command(subcommand).run_from_argv(self.argv)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/core/management/base.py", line 323, in run_from_argv
web_1       |     self.execute(*args, **cmd_options)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/core/management/base.py", line 364, in execute
web_1       |     output = self.handle(*args, **options)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/core/management/commands/shell.py", line 92, in handle
web_1       |     exec(sys.stdin.read())
web_1       |   File "<string>", line 1, in <module>
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/contrib/auth/models.py", line 162, in create_superuser
web_1       |     return self._create_user(username, email, password, **extra_fields)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/contrib/auth/models.py", line 145, in _create_user
web_1       |     user.save(using=self._db)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/contrib/auth/base_user.py", line 66, in save
web_1       |     super().save(*args, **kwargs)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/models/base.py", line 741, in save
web_1       |     force_update=force_update, update_fields=update_fields)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/models/base.py", line 779, in save_base
web_1       |     force_update, using, update_fields,
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/models/base.py", line 870, in _save_table
web_1       |     result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/models/base.py", line 908, in _do_insert
web_1       |     using=using, raw=raw)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/models/manager.py", line 82, in manager_method
web_1       |     return getattr(self.get_queryset(), name)(*args, **kwargs)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/models/query.py", line 1186, in _insert
web_1       |     return query.get_compiler(using=using).execute_sql(return_id)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1332, in execute_sql
web_1       |     cursor.execute(sql, params)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/backends/utils.py", line 99, in execute
web_1       |     return super().execute(sql, params)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/backends/utils.py", line 67, in execute
web_1       |     return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/backends/utils.py", line 76, in _execute_with_wrappers
web_1       |     return executor(sql, params, many, context)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
web_1       |     return self.cursor.execute(sql, params)
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/utils.py", line 89, in __exit__
web_1       |     raise dj_exc_value.with_traceback(traceback) from exc_value
web_1       |   File "/opt/bitnami/python/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
web_1       |     return self.cursor.execute(sql, params)
web_1       | django.db.utils.IntegrityError: duplicate key value violates unique constraint "auth_user_username_key"
web_1       | DETAIL:  Key (username)=(admin) already exists.
```

SebastiaanZ (changes requested):

Pipfile.lock
```diff
@@ -1,378 +0,0 @@
-{
```
SebastiaanZ: You've also removed the `Pipfile.lock` file.

SebastiaanZ (changes requested):
Leaving a comment as you asked for leaving the `Pipfile.lock` out.

Pipfile.lock
lemonsaurus: Fixed!

Pipfile
```diff
@@ -26,6 +26,11 @@ djangorestframework-bulk = "~=0.2.1"
 psycopg2-binary = "~=2.8"
 django-simple-bulma = ">=1.1.6,<2.0"
 django-crispy-bulma = ">=0.1.2,<2.0"
+requests = "*"
```
jchristgit: This should be fixed so we stay on the same major version

pydis_site/apps/main/models/repo_data.py
```diff
@@ -0,0 +1,33 @@
+from django.db import models
+from django.utils import timezone
+
+
+class RepoData(models.Model):
```
jchristgit: wouldn't `Repository` or `RepositoryMetadata` a more descriptive name for this?

pydis_site/apps/main/views/home.py
```diff
@@ -0,0 +1,106 @@
...
+from django.shortcuts import render
+from django.utils import timezone
+from django.views import View
+
+from pydis_site.apps.main.models import RepoData
+
+GITHUB_API = "https://api.github.com/users/python-discord/repos"
+
+
+class Home(View):
```
jchristgit: The other views and viewsets of the API app follow the convention `xView`, I feel `Home` is a bit undescriptive. How about renaming this to `HomeView`?

pydis_site/apps/main/views/home.py
```diff
@@ -0,0 +1,106 @@
...
+
+            if full_name in settings.HOMEPAGE_REPOS:
+                repo_dict[full_name] = {
+                    "full_name": repo["full_name"],
+                    "description": repo["description"],
+                    "language": repo["language"],
+                    "forks_count": repo["forks_count"],
+                    "stargazers_count": repo["stargazers_count"],
+                }
+        print(f"repo_dict after processing = {repo_dict}")
```
jchristgit: I'm gonna assume these are debug prints and will hop out after the draft but leaving a comment here anyways

pydis_site/apps/main/views/home.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class Home(View):
+
+    def _get_api_data(self):
+        """Call the GitHub API and get information about our repos."""
+        repo_dict = {repo_name: {} for repo_name in settings.HOMEPAGE_REPOS}
+
+        # Fetch the data from the GitHub API
+        api_data = requests.get(GITHUB_API)
```
jchristgit: I don't see the value for `requests` here. We could do this just as simply with `urllib` and parsing its body with `json` and wouldn't need the extra dependency. What do you think?

pydis_site/apps/main/views/home.py
```diff
@@ -0,0 +1,106 @@
...
+
+            # If the data is older than 2 minutes, we should refresh it. THIS PROBABLY ALWAYS FAILS?
+            if (timezone.now() - repo_data.last_updated).seconds > 120:
+
+                diff = (timezone.now() - repo_data.last_updated).seconds
+                print(f"okay baby, it's old! the seconds difference comes to: {diff}")
+
+                # Get new data from API
+                api_data_container = self._get_api_data()
+                repo_data_container = []
```
jchristgit: I feel like this "update if not updated recently" is something we'll reuse a bunch and maybe having this as a separate decorator would be useful

pydis_site/apps/main/views/home.py
```diff
@@ -0,0 +1,106 @@
...
+                            stargazers=api_data["stargazers_count"],
+                            language=api_data["language"],
+                        )
+                    repo_data.save()
+                    repo_data_container.append(repo_data)
+                return repo_data_container
+
+            # Otherwise, if the data is fresher than 2 minutes old, we should just return it.
+            else:
+                return list(RepoData.objects.all())
```
jchristgit: I don't think you need the `list` cast here

pydis_site/apps/main/views/home.py
```diff
@@ -0,0 +1,106 @@
...
+                repo_data_container.append(repo_data)
+
+            return repo_data_container
+
+    def get(self, request):
+
+        # Collect the repo data
+        repo_data = self._get_repo_data()
+
+        # Call the GitHub API and ask it for some data
```
jchristgit: Isn't this what the line above does?

pydis_site/static/css/base/base.css
```diff
@@ -0,0 +1,40 @@
...
+  filter: grayscale(1) invert(0.02);
+  background-size: 52px 25.5px;
+  background-position: -1px -2px;
+  color: #00000000;
+}
+
+#bulma-logo {
+  padding-left: 18px;
+  height: 20px;
+  background: url(https://bulma.io/images/bulma-logo-white.png) no-repeat center;
```
jchristgit: not sure if the django and bulma websites like when we're using them for static hosting like this, can we copy them here or is there copy issues stuff with that?

jchristgit (changes requested):
thinks

pydis_site/apps/main/models/repo_data.py
lemonsaurus: sure, RepositoryMetadata is pretty good I guess.

pydis_site/apps/main/views/home.py
lemonsaurus: You could argue that for _any_ use of requests. I think it's a reasonable dependency to add. This is so far the only webrequest this application makes, but there will undoubtedly be many more in the future and I really don't see any reason not to have them all use `requests`. It's one of the top ten most popular python modules in the world for a reason.

pydis_site/apps/main/views/home.py
lemonsaurus: yep yep, I've already removed this, I'll push new commits soon.

pydis_site/apps/main/views/home.py
lemonsaurus: That's probably true. I just wanted it to be consistent, since I'm returning a list (and not a QuerySet) on the other branches... but I could probably drop it anyway, the distinction is not really valuable.

pydis_site/static/css/base/base.css
lemonsaurus: I seriously doubt they'd have issue with us using their image to give credit to them. Also, css can't use DTL, so it's tricky to fetch something from the static folder, which means we'd have to host them somewhere and refer to that, or do something hacky like build the css with DTL templates..

So I think this is fine for now.

Pipfile
lemonsaurus: reasonable enough. locking to `requests = "~=2.21"`

pydis_site/apps/main/views/home.py
lemonsaurus: done. (although it feels a little like hungarian notation, but you're right that it should be consistent.)

pydis_site/apps/main/views/home.py
lemonsaurus: I don't know, I think that sounds a bit over-engineered until we actually have that problem. Or maybe I'm just too lazy to do it. If we ever need that pattern again, we should probably do that, though.

SebastiaanZ (approved):

SebastiaanZ:
I'm happy with the changes I've requested and I trust Volcyyy's judgement for approval of the changes you're making on his request. Since I'm going to bed, I've approved so I won't hold back the merge.

Pipfile
```diff
@@ -26,6 +26,12 @@ djangorestframework-bulk = "~=0.2.1"
...
 django-crispy-bulma = ">=0.1.2,<2.0"
+requests = "~=2.21"
 
 [requires]
 python_version = "3.7"
+
+[scripts]
+makemigrations = "python manage.py makemigrations"
+django_shell = "python manage.py shell"
+test = "python manage.py test -v 2"
```
jchristgit: shouldn't we remove the `-v 2` here so people can pass along what they want instead?

pydis_site/apps/main/tests/test_repodata_helpers.py
```diff
@@ -0,0 +1,66 @@
...
+            description="testrepo",
+            forks=42,
+            stargazers=42,
+            language="English",
+        )
+        repo_data.save()
+        metadata = home_view._get_repo_data()
+
+        self.assertIsInstance(metadata[0], RepositoryMetadata)
+        print(metadata[0])  # Tests the __str__ in the model
```
jchristgit: how about doing something similar to the tests for the api models, just checking that
```suggest
self.assertIsInstance(str(metadata[0]), str)
```

pydis_site/apps/main/tests/test_repodata_helpers.py
```diff
@@ -0,0 +1,66 @@
...
+        """Tests if the _get_api_data helper returns what it should."""
+
+        home_view = HomeView()
+        api_data = home_view._get_api_data()
+        repo = settings.HOMEPAGE_REPOS[0]
+
+        self.assertIsInstance(api_data, dict)
+        self.assertEquals(len(api_data), len(settings.HOMEPAGE_REPOS))
+        self.assertIn(repo, api_data.keys())
+        self.assertIn("stargazers_count", api_data[repo])
```
jchristgit: don't these tests hit the github API at every run? i feel like having our tests rely on an external system like that is not a good idea

pydis_site/apps/main/views/home.py
```diff
@@ -0,0 +1,101 @@
...
+                    repo_name=api_data["full_name"],
+                    description=api_data["description"],
+                    forks=api_data["forks_count"],
+                    stargazers=api_data["stargazers_count"],
+                    language=api_data["language"],
+                )
+                repo_data.save()
+                repo_data_container.append(repo_data)
+
+            return repo_data_container
```
jchristgit: this reads like lots of duplicated code, maybe `get_or_create` could help?

pydis_site/apps/main/views/home.py
```diff
@@ -0,0 +1,101 @@
...
+            # Otherwise, if the data is fresher than 2 minutes old, we should just return it.
+            else:
+                return RepositoryMetadata.objects.all()
+
+        # If this is raised, the database has no repodata at all, we will create them all.
+        except RepositoryMetadata.DoesNotExist:
+
+            # Get new data from API
+            api_data_container = self._get_api_data()
+            repo_data_container = []
```
jchristgit: not a fan of these varnames, how about `api_repositories` and `database_repositories`

pydis_site/hosts.py
```diff
@@ -3,8 +3,7 @@
 
 host_patterns = patterns(
     '',
-    # host(r"subdomain pattern", "URLs module", "host entry name")
     host(r'admin', 'pydis_site.apps.admin.urls', name="admin"),
     host(r'api', 'pydis_site.apps.api.urls', name='api'),
-    host(r'.*', 'pydis_site.apps.home.urls', name=settings.DEFAULT_HOST)
+    host(r'.*', 'pydis_site.apps.main.urls', name=settings.DEFAULT_HOST)
```
jchristgit: not a fan of the `main` app name, why not leave this as `home`?

pydis_site/settings.py
```diff
@@ -255,5 +254,17 @@
...
+
+# Which of our GitHub repos should be displayed on the front page, and in which order?
+HOMEPAGE_REPOS = [
+    "python-discord/site",
+    "python-discord/bot",
+    "python-discord/snekbox",
+    "python-discord/seasonalbot",
+    "python-discord/django-simple-bulma",
+    "python-discord/django-crispy-bulma",
+]
```
jchristgit: i feel like this is abusing the settings.py a bit because we only need this on the front page view. maybe move it to a class variable for the index view?

pydis_site/static/css/base/base.css
```diff
@@ -0,0 +1,40 @@
+html {
+    overflow:auto;
```
jchristgit: this uses 4 spaces and no space between colon and text then the ones below use two spaces and space between colon and text, we should stay consistent

pydis_site/templates/home/index.html
```diff
@@ -7,65 +7,88 @@
...
-    </div>
-    <div class="hero-foot">
-      {% include "base/navbar.html" with active_item=False dropdowns=True icon_weight="fas" use_logo=False %}
-    </div>
-  </section>
+  {% include "base/navbar.html" %}
+
   <section class="section">
+
+    <!-- Who are we? -->
```
jchristgit: this should be a jinja2 comment like in theo ther template

pydis_site/templates/home/index.html
```diff
@@ -7,65 +7,88 @@
...
+                  allowfullscreen></iframe>
         </div>
       </div>
+
+      <!-- Projects -->
+      <h1 class="is-size-1">Projects</h1>
+      <br>
+      <div class="columns is-multiline">
+
+        <!-- Display projects from settings.HOMEPAGE_REPOS -->
```
jchristgit: shouldn'T these all be jinja comments

pydis_site/templates/home/index.html
```diff
@@ -7,65 +7,88 @@
...
-          <h1 class="title">Join the community</h1>
 
-          {# TODO: Fix URL #}
-          <a href="https://pythondiscord.com/invite">
-            <img class="discord-banner" src="https://discordapp.com/api/guilds/267624335836053506/embed.png?style=banner3" alt="Discord banner" />
-          </a>
+        <!-- Intro video -->
+        <div class="column is-half video-container">
+          <iframe src="https://www.youtube.com/embed/DIBXg8Qh7bA" frameborder="0"
+                  allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
```
jchristgit: this says autoplay but it didnt start playing for me, is that intentional? autoplay is a horrible thing

pydis_site/templates/home/index.html
```diff
@@ -7,65 +7,88 @@
...
       <div class="container">
         <h1 class="title is-6 has-text-grey">
           Sponsors
         </h1>
-        <img src="{% static "images/sponsors/linode.png" %}" alt="" />
-        <img src="{% static "images/sponsors/jetbrains.png" %}" alt="" />
-        <img src="{% static "images/sponsors/adafruit.png" %}" alt="" />
+        <a href="https://linode.com"><img src="{% static "images/sponsors/linode.png" %}" alt="Linode"/></a>
+        <a href="https://jetbrains.com"><img src="{% static "images/sponsors/jetbrains.png" %}" alt=""/></a>
+        <a href="https://adafruit.com"><img src="{% static "images/sponsors/adafruit.png" %}" alt=""/></a>
```
jchristgit: one of these provides an `alt` and the others don't. we should add that to each

pydis_site/urls.py
```diff
@@ -2,5 +2,5 @@
 
 
 urlpatterns = (
-    path('', include('pydis_site.apps.home.urls', namespace='home')),
+    path('', include('pydis_site.apps.main.urls', namespace='home')),
```
jchristgit: this is a bit contradictory because the app name is `main` but you namespace it under `home`

jchristgit (changes requested):
more thouhgts

Pipfile
lemonsaurus: sure.

pydis_site/apps/main/views/home.py
lemonsaurus: get_or_create can't solve this problem, because the entire point is that I only want to call the API if the get fails. with get_or_create I would need to call the API anyway just so I have something to put into the create condition in case get fails. You'll have to PR in a DRYer solution yourself if you want to improve it because I don't have one, and I already spent enough time looking for one.

pydis_site/templates/home/index.html
lemonsaurus: I'm not sure, this embed is just generated by YouTube, it never autoplays. I removed it and it doesn't appear to change anything. Maybe it just allows an autoplay setting to be toggled somewhere else?

pydis_site/settings.py
lemonsaurus: Okay, changed.

pydis_site/apps/main/tests/test_repodata_helpers.py
lemonsaurus: This is probably true. From what I can tell, the correct way to solve this is to somehow patch the `request.get` call to just return a mock object instead. That way the code can run exactly as it does now but without making real calls to the API.

I'll give this a try.

lemonsaurus: Solved.

docker/app/migrate_and_serve.sh
```diff
@@ -9,7 +9,12 @@
 
 echo [i] Applying migrations.
 python manage.py migrate --verbosity 1
+
 echo [i] Collecting static files.
 python manage.py collectstatic --no-input --clear --verbosity 0
+
+echo [i] Creating a superuser
```
jchristgit: the other log lines here end with a dot and i feel like to stay consistent this one should, too

docker/app/migrate_and_serve.sh
```diff
@@ -9,7 +9,12 @@
...
+echo "from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser('admin', 'admin', 'admin') if not User.objects.filter(username='admin').exists() else print('Username already exists')" | python manage.py shell
```
jchristgit: how about "Admin user already exists.", username sounds a bit off

pydis_site/apps/home/models/repo_data.py
```diff
@@ -0,0 +1,33 @@
+from django.db import models
+from django.utils import timezone
+
+
+class RepositoryMetadata(models.Model):
```
jchristgit: you updated the model name but the module name should probably be updated to `repository_metdata.py` too then

pydis_site/apps/home/views/home.py
```diff
@@ -0,0 +1,110 @@
+import requests
+from django.shortcuts import render
+from django.utils import timezone
+from django.views import View
+
+from pydis_site.apps.home.models import RepositoryMetadata
+
+
+class HomeView(View):
+    """The view"""
```
jchristgit: this docstring doesn't spark anything to me, how about something like "The index page of the website."

pydis_site/apps/home/views/home.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def _get_repo_data(self):
+        """Build a list of RepositoryMetadata objects that we can use to populate the front page."""
+
+        # Try to get site data from the cache
+        try:
+            repo_data = RepositoryMetadata.objects.get(repo_name="python-discord/site")
+
+            # If the data is older than 2 minutes, we should refresh it.
+            if (timezone.now() - repo_data.last_updated).seconds > 120:
```
jchristgit: isn't 2 minutes a bit often though? that said, maybe this is also better as a class constant? `repository_cache_ttl` or something like that

pydis_site/templates/home/index.html
```diff
@@ -7,65 +7,88 @@
...
+                  allowfullscreen></iframe>
         </div>
       </div>
+
+      {# Projects #}
+      <h1 class="is-size-1">Projects</h1>
+      <br>
+      <div class="columns is-multiline">
+
+        {# Display projects from settings.HOMEPAGE_REPOS #}
```
jchristgit: this was moved to the view so this comment should be updated

jchristgit (changes requested):
call me aristoteles the ancient greek thinker for i have more comments on this :shipit::shipit::shipit::shipit::shipit::shipit:

lemonsaurus:
I just realized, in mobile mode there's no discord invite. I need to fix that too.

pydis_site/apps/home/views/home.py
lemonsaurus: oh lol I guess I must've started writing this and just forgotten about it.

pydis_site/apps/home/views/home.py
lemonsaurus: Okay so, it looks like we get 60 of these per hour with unauthenticated access. This increases to 5000 if we do bother to auth, so if we needed more, it'd be easy enough to get them, but I'd rather not add authentication if we don't need to.

For 60 an hour, that means that this would exhaust 50% of the rate limit capacity, which, yeah, is maybe a little high when the only advantage is quicker tracking of new stars or whatever.

I'm gonna set it to 10 minutes for now, how does that sound?

jchristgit (approved):
keep spending most our lives living in the review paradise

PR #215, opened 2019-04-21
[#192] Flake8-docstrings

### LINKED ISSUE #192: Django: Completing the move to flake8-docstrings
We've decided to add `flake8-docstrings` to project requirements for this repository.
We've agreed on the following list of errors to ignore:
```sh
ignore=
    P102,B311,W503,E226,S311,
    # Missing Docstrings
    D100,D104,D107,
    # Docstring Whitespace
    D202,D203,D204,D212,D214,D215,
    # Docstring Quotes
    D301,D302,
    # Docstring Content
    D401,D402,D405,D406,D407,D408,D409,D410,D411,D412,D413, D414
```

gdude2002:
Adds flake8-docstrings along with the configuration discussed in #192 and on Discord. Fixes #192.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -14,6 +14,7 @@ mccabe = "~=0.6.1"
 pep8-naming = "~=0.8.2"
 coverage = "~=4.5.3"
 unittest-xml-reporting = "~=2.5.1"
+flake8-docstrings = "*"
 
 [packages]
 django = "~=2.2"
--- a/Pipfile
+++ b/Pipfile
@@ -14,7 +14,7 @@ mccabe = "~=0.6.1"
 pep8-naming = "~=0.8.2"
 coverage = "~=4.5.3"
 unittest-xml-reporting = "~=2.5.1"
-flake8-docstrings = "*"
+flake8-docstrings = "~=1.3.0"
 
 [packages]
 django = "~=2.2"
--- a/pydis_site/apps/home/templatetags/extra_filters.py
+++ b/pydis_site/apps/home/templatetags/extra_filters.py
@@ -5,4 +5,18 @@ register = template.Library()
 
 @register.filter
 def starts_with(value: str, arg: str):
+    """
+    Simple filter for checking if a string value starts with another string.
+
+    Kind of surprising that Django doesn't come with this one, actually...
+
+    Usage:
+
+    ```django
+        {% if request.url | starts_with:"/wiki" %}
+          ...
+        {% endif %}
+    ```
+    """
+
     return value.startswith(arg)
--- a/pydis_site/apps/home/templatetags/extra_filters.py
+++ b/pydis_site/apps/home/templatetags/extra_filters.py
@@ -8,8 +8,6 @@ def starts_with(value: str, arg: str):
     """
     Simple filter for checking if a string value starts with another string.
 
-    Kind of surprising that Django doesn't come with this one, actually...
-
     Usage:
 
     ```django
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,19 @@
 [flake8]
-max-line-length=100
-ignore=S106
+ignore=
+    P102,B311,W503,E226,S311,
+    # Missing Docstrings
+    D100,D104,D107,
+    # Docstring Whitespace
+    D202,D203,D204,D212,D214,D215,
+    # Docstring Quotes
+    D301,D302,
+    # Docstring Content
+    D400,D401,D402,D405,D406,D407,D408,D409,D410,D411,D412,D413,D414
+
+per-file-ignores =
+    **/tests/**:D101,D102,D103,D105,D106,S106
+
 application_import_names=pydis_site
 exclude=__pycache__, venv, .venv, **/migrations
 import-order-style=pycharm
+max-line-length=100
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "1b31f6abe3c7782705e8023d1a53b299e713d19ac31ef2926cba9eb4f4c59efb"
+            "sha256": "de8e753b689c949df77841cf8546c7a1c575d43a460c532faa01642b8773fd23"
         },
         "pipfile-spec": 6,
         "requires": {
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "78085b805ff9d797c2ce4720c4fc84a91347233b93e4b85615c9833e6b208ed8"
+            "sha256": "1b31f6abe3c7782705e8023d1a53b299e713d19ac31ef2926cba9eb4f4c59efb"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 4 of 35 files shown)


gdude2002:
I've done as many docstrings as I could, but I'm a little stuck on them. Here's what's left:

```py
./pydis_site/apps/api/serializers.py:16:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:24:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:46:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:56:1: D102 Missing docstring in public method
./pydis_site/apps/api/serializers.py:68:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:76:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:85:1: D102 Missing docstring in public method
./pydis_site/apps/api/serializers.py:99:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:100:1: D102 Missing docstring in public method
./pydis_site/apps/api/serializers.py:114:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:125:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:132:1: D102 Missing docstring in public method
./pydis_site/apps/api/serializers.py:136:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:144:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:152:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:160:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:168:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:178:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:186:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:194:1: D101 Missing docstring in public class
./pydis_site/apps/api/serializers.py:205:1: D101 Missing docstring in public class
./pydis_site/apps/api/viewsets/bot/nomination.py:9:1: D101 Missing docstring in public class
```

Could someone give me a hand on understand what's going on with these classes, or help me to document them somehow?

jchristgit:
I'll push docstrings to this branch for the missing ones if you don't mind

gdude2002:
Sure thing, go for it

Pipfile
```diff
@@ -14,6 +14,7 @@ mccabe = "~=0.6.1"
 pep8-naming = "~=0.8.2"
 coverage = "~=4.5.3"
 unittest-xml-reporting = "~=2.5.1"
+flake8-docstrings = "*"
```
jchristgit: This should be pinned, or tilded.
```suggestion
flake8-docstrings = "~=1.3.0"
```

jchristgit (changes requested):
Rock solid, but I still see some inconsistencies between spacing after docstrings (in methods) and no spacing. I think we should consistently have a newline after docstrings in methods. what do you think?

gdude2002:
I don't disagree, but it surprises me that we can't actually lint for that - what should we do about it?

pydis_site/apps/home/templatetags/extra_filters.py
```diff
@@ -5,4 +5,18 @@
 
 @register.filter
 def starts_with(value: str, arg: str):
+    """
+    Simple filter for checking if a string value starts with another string.
+
+    Kind of surprising that Django doesn't come with this one, actually...
```
lemonsaurus: I think we can probably live without this line.

lemonsaurus (changes requested):
Excellent quality PR.

lemonsaurus:
As for the "what do we do about that" question from earlier - Well, for now, we just make sure to consistently put a newline. Not much else we can do about it for now, maybe we could make a pull request to `flake8-docstrings`?

gdude2002:
Addressed.

Yeah, maybe that's a good idea. It's actually `pydocstyle` under the hood - I have to wonder if they'd accept it though. Is it part of the PEP?

lemonsaurus:
PEP257 actually is a bit complicated on this subject. It specifies that one-liner docstrings should _not_ be followed by a blank line, but that class docstrings (even one-liners) should. It does not make any specific requirements of multi-line docstrings, nor are there any examples to draw from in the PEPs.

If we look at a modern bit of python code written by a core dev, we see that there are no newlines after multi-line docstrings either for example in https://github.com/python/cpython/blob/master/Lib/asyncio/coroutines.py#L104:L109, which was written by `vstinner`.

That said, some parts of the standard library have absolutely zero regard for PEP8. the `collections` module for instance prefers to use single quotes for docstrings, sometimes not even triple single quotes. Other parts of the standard library uses multi-block comment docstrings (wtf) and many forego docstrings entirely.

So what about Guido himself? Well, after digging through some recent commits that he's made on his `mypy` project, it appears Guido consistently puts a newline after module docstrings and class docstrings, but never after method docstrings (neither oneliners nor multiliners).

image


So I guess that's the de-facto correct way to do it.

lemonsaurus (approved):

gdude2002:
That's pretty interesting, thanks for looking into that.

I just remembered that `pydocstyle` actually does allow you to specify one of a number of different code style presets, so they may actually accept that PR. Can't hurt to ask, right?

lemonsaurus:
Actually this looks perfectly enforcable now that I've had a look at it.

D202 | No blank lines allowed after function docstring
D204 | 1 blank line required after class docstring

These two are all that's preventing us from linting for Guido-style docstring newlines.

lemonsaurus:
I think I'd like to suggest that we don't ignore those two rules, considering my above posts. This would have several advantages:
- Complies with PEP257 and PEP8
- Immediately enforcable with the linter
- It's how Guido and other core devs do it, why should we be any worse?

SebastiaanZ:
I'm all for following the general guidelines, but we've discussed this before and some people were vocal against the no-newline thing, since they didn't like the look of that. I don't think that's a particularly strong argument for going against the general conventions and an easy linter setup, but maybe they have a better reason I don't have for that, so it's not me who should be defending that point.

As always, I'm pretty pliable when it comes to rules like this: As long as we're consistent and the style doesn't hurt readability, I can live with the decisions we make. Personally, I'd stick with the official style guides, unless we have good reasons to deviate from them.

lemonsaurus:
I looked through the previous tickets where we talked about which rules to ignore in `flake8-docstrings` and we seem to have skipped discussing this. I know that in the past some users have been outspoken about keeping the newline, but I feel like in this case there good reasons to consider changing our policy on this. The most important of which is to be able to _enforce_ that consistency.

scragly:
These types of style choices are highly opinionated so will not always get a 100% agreed conclusion among contributors since everyone has their own history, influences and styles. The most important thing is consistency, in my own opinion, and in the end there's no negative outcome from being able to enforce this particular style through linting, it only improves the simplicity of being able to provide a consistent style since linting can be done automatically. It's an improvement for possible first time contributors.

gdude2002:
Personally I prefer the extra newline, but I don't particularly mind not having it and I think consistency is definitely important.

MarkKoz:
Style-wise, either way is fine with me. But from a practical stand point, I think we should avoid making stylistic choices in such vein (i.e. simple formatting rules) that we cannot lint for within reason. Thus I am in favour of not having a newline after every docstring. As for whether it should be abolished completely or only for functions: again, I am impartial to this.

lemonsaurus:
With these two rules unignored, classes would require a newline, functions wouldn't allow one, and module docstrings presumably could go either way. But we don't use module docstrings that often anyway, so I don't really see that as a huge problem.

lemonsaurus:
I'd like to hear from @jchristgit and @sco1 before we make a decision, but I think so far it seems like the consensus is "it doesn't matter, just go with whatever we can lint so we can enforce it". I also agree with this.

jchristgit (approved):

PR #216, opened 2019-04-21
Manually pin urllib3 version to mitigate vulnerability

sco1:
Addresses https://nvd.nist.gov/vuln/detail/CVE-2019-11324, an issue with `urllib3 1.24.1` that is fixed in `1.24.2`, released a few days ago.

`urllib3` is a `flask_dance` and `requests-oauthlib` requirement via `requests`, which pins to `>=1.21.1,<1.25`, so we should be able manually pin `1.24.2` without conflict.

This PR manually pins this version, resynced dependencies, and relocks.


lemonsaurus (dismissed):

lemonsaurus (approved):
it clearly works.

GhostofGoes (approved):

PR #220, opened 2019-04-27
Changing the way nominations work in the backend

SebastiaanZ:
This PR changes the way nominations are stored in the database and how the bot can interact with stored nominations using the API.

A Nomination object now has the following fields, as defined in the model:
- `id`: The primary key in the form of an AutoField
- `active`: If the nomination is still active
- `actor`: The staff member who nominated the `user`
- `reason`:  The reason for nomination the `user`
- `user`: The nominee, the user that was nominated by the `actor`
- `inserted_at`: The datetime of the nomination
- `unnominate_reason`: The reason for ending the nomination
- `unwatched_at`: The datetime the nomination was ended and the user unwatched

The API endpoint for nominations has the following routes:

- `bot/nominations`: the list view to `get` nominations (plural) or to `post` a new nomination;
- `bot/nominations/<id:int>`: the detail view to `get` a single nomination or to `patch` one of the fields;
- `bot/nominations/<id:int>/end_nomination`: the endpoint to end an active nomination.

The API has several restrictions built-in; they can be found in the docstring of the Nomination ViewSet. The most important ones are:

- A `user` can only have one active nomination at a time;
- An `active` nomination cannot have an `unnominate_reason`;
- The `actor` and `user` fields cannot be modified after creation;
- The `active`, `inserted_at`, and `unwatched_at` fields cannot be set or altered manually.

In addition, I've added the `OrderingFilter` to both the `Infraction` ViewSet as well as the `Nomination` ViewSet to be able to get infractions/nominations in a specified order (e.g., in the order of `inserted_at` to display them chronologically). I've also Kaizen'ed two of the filter parameters in the docstring of the `Nomination` ViewSet.

## Things to pay attention to while reviewing:
- I had to manually write a `migration` file to make sure we could migrate from the previous version of the Nomination model to the current one. I've tested in on my system with and without rows in the nomination table, both with `docker-compose` and a separate PostgresDB, but please confirm this. We really want to keep the migration history intact and not break existing builds.
- Like with the `Infraction` model, the `Nomination` model relies on the `User` table via `ForeignKey` fields. This means that users (both `actor` and `user` fields) need have an entry in the `User` table for the `Nomination` endpoint to work. 
- I may have gone a bit overboard with writing tests, as I did not only want to have 100% coverage, but also wanted to test all test cases I could come up with.


pydis_site/apps/api/viewsets/bot/infraction.py
```diff
@@ -28,11 +28,12 @@ class InfractionViewSet(CreateModelMixin, RetrieveModelMixin, ListModelMixin, Ge
 
     #### Query parameters
     - **active** `bool`: whether the infraction is still active
-    - **actor** `int`: snowflake of the user which applied the infraction
+    - **actor__id** `int`: snowflake of the user which applied the infraction
```
MarkKoz, line 31: I don't think the `__id` suffixes belong in the documentation. It's more of an implementation detail, right? DRF seems to map `actor` to `actor__id` just fine based on the tests I've done and written.

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+    #### Query parameters
+    - **active** `bool`: whether the nomination is still active
+    - **actor__id** `int`: snowflake of the user who nominated the user
+    - **user__id** `int`: snowflake of the user who received the nomination
+    - **ordering** `str`: comma-separated sequence of fields to order the returned results
+
+    Invalid query parameters are ignored.
+
+    #### Response format
+    [
```
MarkKoz: These need the "interpreter syntax" prefix things like so:

https://github.com/python-discord/site/blob/ba9fbe428407127a72edb6b7cb5f2555f1be1642/pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py#L25-L27

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+            )
+
+        serializer = self.get_serializer(instance, data=request.data, partial=True)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+
+        return Response(serializer.data)
+
+    @action(detail=True, methods=['patch'])
+    def end_nomination(self, request, pk=None):
```
MarkKoz: Can we be consistent and call this endpoint `unnominate`? Though, that word sounds a bit awkward to me. Is `dismiss` any better? As in "dismiss a nomination".

MarkKoz (changes requested):

pydis_site/apps/api/viewsets/bot/infraction.py
SebastiaanZ: Do you have a test for that? I can't find it, but when I test it, providing `user` as a param is ingored, while `user__id` works:

Full infraction table:
```json
Status code: 200
[
    {
        "id": 1,
        "inserted_at": "2019-04-27T20:54:53.930875Z",
        "expires_at": null,
        "active": false,
        "user": 336843820513755157,
        "actor": 336843820513755157,
        "type": "watch",
        "reason": "Watch that man",
        "hidden": true
    },
    {
        "id": 2,
        "inserted_at": "2019-04-27T21:19:36.568151Z",
        "expires_at": null,
        "active": false,
        "user": 336843820513755157,
        "actor": 336843820513755157,
        "type": "watch",
        "reason": "Unwatched: Away with you!",
        "hidden": true
    }
]
```

Request with non-existing `user` as a filter `param`:
```py
r = requests.get(f"http://api.pythondiscord.local:8000/bot/infractions", params={'user': 1}, headers=auth_headers)
print(f"Status code: {r.status_code}")
print(json.dumps(r.json(), indent=4))
```
results in all fields being returned, regardless of the `user`'s id, since the invalid `user` was ignored:
```json
Status code: 200
[
    {
        "id": 1,
        "inserted_at": "2019-04-27T20:54:53.930875Z",
        "expires_at": null,
        "active": false,
        "user": 336843820513755157,
        "actor": 336843820513755157,
        "type": "watch",
        "reason": "Watch that man",
        "hidden": true
    },
    {
        "id": 2,
        "inserted_at": "2019-04-27T21:19:36.568151Z",
        "expires_at": null,
        "active": false,
        "user": 336843820513755157,
        "actor": 336843820513755157,
        "type": "watch",
        "reason": "Unwatched: Away with you!",
        "hidden": true
    }
]
```
while providing it as `user__id`, it works as expected:
```py
r = requests.get(f"http://api.pythondiscord.local:8000/bot/infractions", params={'user__id': 1}, headers=auth_headers)
print(f"Status code: {r.status_code}")
print(json.dumps(r.json(), indent=4))
```
results in no results being returned, as expected:
```json
Status code: 200
[]
```

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: I'm going to change this to be consistent, but, personally, I don't like that the response format in the docstring is not what the actual response is, but some arbitrary form (REPL output).

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: Yeah, it's ugly, I agree. I was debating with myself what a proper name could be, but I think just `end` would do as well, as it would look like: `bot/infractions/1/end`. As an endpoint, that makes sense as well. Same logic goes for `dismiss`. I don't mind either way, but I don't like `unnominate` as a term.

MarkKoz: As much as I like short words for convenience, I think `dismiss` makes more sense than `end`.

pydis_site/apps/api/viewsets/bot/infraction.py
MarkKoz: Maybe I didn't test it with manual requests after all. But in the tests I wrote, it does use just `user`. However, that would mean tests are not properly reproducing actual usage of the API. 

https://github.com/python-discord/site/blob/django/pydis_site/apps/api/tests/test_infractions.py#L170

pydis_site/apps/api/viewsets/bot/nomination.py
MarkKoz: Yeah, it's just for consistency. I am aware there was some discussion on the style in the past.

pydis_site/apps/api/viewsets/bot/infraction.py
SebastiaanZ: It does accept `user` when you're creating an object, since that's the `Field` name and it will convert that to a  ForeignKey to a user in the `User` table. However, when *filtering*, it takes it as is and I think it only accepts the `user__id` as a filter field, not `user` by itself. That's why I only changed the query parameters, not the fields on `POST`.

I think it's because, in principle, we can designate any of the `Fields` of `User` as a filtering field and Django uses the `__` to select the specific field, in this case `id`. We could also have used `user__name`. But, I do think you need to specify the specific Field of `User` you're filtering against.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: Hmm, for me, `end` makes more sense than `dismiss`, but I don't really care either way. The length of the word doesn't factor, in my opinion.

pydis_site/apps/api/viewsets/bot/infraction.py
MarkKoz: Good catch then. Is there anything we can do to not have the suffix though? It's kinda ugly but also not worth spending much time on.

SebastiaanZ: I don't know how to do that, but my guess is there probably is a way to do that. Maybe someone knows? Otherwise I'd have to spend some time digging through the docs/SO.

---

To be sure, I added these two tests locally to your test suite:
```py
    def test_filter_user(self):
        url = reverse('bot:infraction-list', host='api')
        response = self.client.get(
            url,
            data={
                'user': 99989890898990124
            }
        )

        infractions = response.json()

        self.assertEqual(len(infractions), 0)

    def test_filter_user__id(self):
        url = reverse('bot:infraction-list', host='api')
        response = self.client.get(
            url,
            data={
                'user__id': 99989890898990124
            }
        )

        infractions = response.json()

        self.assertEqual(len(infractions), 0)
```
to your test suite and the first fails:
```py
FAIL: test_filter_user (pydis_site.apps.api.tests.test_infractions.InfractionTests)
----------------------------------------------------------------------
  File "/home/sebastiaan/pydis/site_django/pydis_site/apps/api/tests/test_infractions.py", line 99, in test_filter_user
    self.assertEqual(len(infractions), 0)
AssertionError: 2 != 0
```
I've also tried the `url` notation (which should be the same), `{url}?user=312498397` vs `{url}?user__id=312498397`and that gives the same result.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: I've changed it to `dismiss` in the latest push.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: Changed.

pydis_site/apps/api/viewsets/bot/infraction.py
SebastiaanZ: I've left this as it is now. If someone has a nice way of doing this, don't hesitate to reopen this conversation.

pydis_site/apps/api/viewsets/bot/nomination.py
lemonsaurus: this is really nitpicky but I think you `withdraw` a nomination, you don't dismiss it.

SebastiaanZ: `withdraw` in the context of nominations is always negative right? For us, we end the nomination period both when the nominee is accepted and rejected for helper. We're kinda nitpicky here, but better to have a correct term now than later. It's fairly easy to change, so I don't care what we settle on.

MarkKoz: That is a good point. In which case, `dismiss` is negative too. `End` doesn't make sense because a nomination isn't something that "starts", at least in the sense that one would not say "I started a nomination". To me, `withdraw` only makes sense if it's coming from the person that originally nominated that someone. For everyone else, it's `dismiss`:

> Jane nominated Bob but withdrew it two days later.
> The board would have dismissed the nomination for Bob anyway because he's stinky.

That being said, I remembered another word: `rescind`. To me, this sounds OK coming from both, though it still has a negative connotation.

Edit: I think these are just biases from often having heard these ideas expressed using particular phrasing. Still doesn't sit well with me though.

lemonsaurus: I don't have a problem with end. It implies an end to the nomination process, sounds completely neutral. I agree that both withdraw and dismiss (and rescind) have too much of a negative connotation to be applicable.

MarkKoz: Though it isn't perfect in my eyes (and it's not very important that it should be in the first place), I cannot come up with anything better. `End` seems like the best choice right now.

MarkKoz:
Would it be worthwhile to have a field distinguish accepted and rejected nominations? Is this a feature the staff would even find useful?

As is, it could be filtered if a prefix was included for reasons of one but not the other. Nominations that are accepted could be set to an inactive state automatically (by using `on_member_update` to listen for the role being added), so an "end reason" could be set automatically with a consistent format that could be queried easily.

lemonsaurus:
I don't think it'd be particularly useful to be able to query for that. I don't see how or when we'd ever need it, but maybe that's just me.

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+        - The nomination was already inactiive.
+    - 404: if an infraction with the given `id` could not be found
+    """
     serializer_class = NominationSerializer
-    queryset = Nomination.objects.prefetch_related('author', 'user')
-    frozen_fields = ('author', 'inserted_at', 'user')
+    queryset = Nomination.objects.all()
+    filter_backends = (DjangoFilterBackend, SearchFilter, OrderingFilter)
+    filter_fields = ('user__id', 'actor__id', 'active')
+    frozen_fields = ('id', 'actor', 'inserted_at', 'user', 'unwatched_at', 'active')
```
MarkKoz: Can some of these be implemented using some combination of `read_only=True`, `ReadOnlyField`, and `CreateOnlyDefault` in the serialiser instead? Though this may involve having to write multiple serialisers (or checking which HTTP method was used from the serialiser). I just feel like validation does not belong directly in the view, but I don't know what's best practice here.

SebastiaanZ: I'm not sure, I followed the format/style of some of the other API endpoints that also use a `frozen_fields` check like that. This is my first time doing anything in Django and DRF, so I'll read the docs and experiment for a bit.

pydis_site/apps/api/admin.py
```diff
@@ -3,7 +3,8 @@
 from .models import (
     BotSetting, DeletedMessage,
     DocumentationLink, Infraction,
-    MessageDeletionContext, OffTopicChannelName,
+    MessageDeletionContext, Nomination,
+    OffTopicChannelName,
```
jchristgit: if you already change the number of imports per line, this should probably just be fixed to one symbol per line

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+    ...     'reason': 'They know how to explain difficult concepts',
+    ...     'user': 336843820513755157,
+    ...     'inserted_at': '2019-04-25T14:02:37.775587Z',
+    ...     'unnominate_reason': 'They were helpered after a staff-vote',
+    ...     'unwatched_at': '2019-04-26T15:12:22.123587Z'
+    ... }
+
+    ### Status codes
+    - 200: returned on succes
+    - 404: returned if an nomination with the given `id` could not be found
```
jchristgit:
```suggestion
    - 404: returned if a nomination with the given `id` could not be found
```
unless my english fails me

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+    ## Routes
+    ### GET /bot/nominations
+    Retrieve all nominations.
+    May be filtered and ordered by the query parameters.
+
+    #### Query parameters
+    - **active** `bool`: whether the nomination is still active
+    - **actor__id** `int`: snowflake of the user who nominated the user
+    - **user__id** `int`: snowflake of the user who received the nomination
+    - **ordering** `str`: comma-separated sequence of fields to order the returned results
```
jchristgit, line 33: does this correspond to django's `.order_by` queryset or how does this work? having some docs in here about that would be nice

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+        DRF method for creating a Nomination.
 
         Called by the Django Rest Framework in response to the corresponding HTTP request.
         """
+        for field in request.data:
+            if field in self.frozen_on_create:
+                raise ValidationError({field: ['This field cannot be set at creation.']})
+
+        user_id = request.data.get("user")
+        if Nomination.objects.filter(active=True, user__id=user_id):
```
jchristgit: prefer explicit `exists`, see https://docs.djangoproject.com/en/2.2/ref/models/querysets/#django.db.models.query.QuerySet.exists

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+        instance = self.get_object()
+        if not instance.active:
+            raise ValidationError({'active': ['A nomination must be active to be ended.']})
+
         serializer = self.get_serializer(instance, data=request.data, partial=True)
         serializer.is_valid(raise_exception=True)
+        instance.active = False
+        instance.unwatched_at = timezone.now()
         serializer.save()
+        instance.save()
```
jchristgit: hm, honestly, I'm not entirely sold on the separate endpoint. why can't we do this with the standard put endpoint?

jchristgit (changes requested):
I like it, some nitpicks.

pydis_site/apps/api/viewsets/bot/nomination.py
MarkKoz: It's separate from the queryset from what I can tell. Docs [here](https://www.django-rest-framework.org/api-guide/filtering/#specifying-which-fields-may-be-ordered-against).

jchristgit: Ah, I see

pydis_site/apps/api/viewsets/bot/nomination.py
MarkKoz: Semantically, `PATCH` is more appropriate because `PUT` is meant to replace the entire object. However, `PATCH` is already being utilised for a more generic modification.

That being said, it does seem possible to merge functionality into just `PATCH`. The validation logic would pivot against whether the `active` field is being set. I'm not sure if it'd be good design to merge though, the concern being that it'd be too complicated for a single endpoint. I think it's not that bad though.

pydis_site/apps/api/admin.py
SebastiaanZ: Yes, good point. Didn't really think properly about it.

I think that keeping it at two per line is probably better, due to the sheer number of imports we do here. It's not like it's unreadable with two symbols per line here, it's not a complicated file or import structure. I'll clean it up and reformat it to the old style of two per line.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: Agreed, I'll change this.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: That's why I chose for `PATCH` instead of `PUT`, since that was my understanding as well.

I chose for a different ent point because the function of the two actions was semantically different: The regular patch to update a field of a nomination in its current state (whether it's active or inactive; it's a minor partial update); the `end` endpoint to alter the total state of the nomination. 

I noticed that the `infraction` viewset had a different endpoint for the `extended` view (`/bot/infractions/expanded`) and I liked the explicitness in that. However, infractions also change `active` to `inactive` state using the regular endpoint, so I can see the consistency point.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: I'll change this back to `end` then.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: I think this would be an interesting thing to rewrite in the future, but since it doesn't change the way the API works from the outside or the model, I think we should look into this at a later time. I'd like to focus on the other things that need to be done first, like the getting the django branch of the bot ready.

MarkKoz: Fair enough. Plus, this is apparently consistent with other view sets so any possible changes to how this is done can be batched together in the future.

pydis_site/apps/api/models/bot/nomination.py
```diff
@@ -20,14 +20,22 @@ class Nomination(ModelReprMixin, models.Model):
...
         on_delete=models.CASCADE,
         help_text="The nominated user.",
-        primary_key=True,
         related_name='nomination'
     )
     inserted_at = models.DateTimeField(
         auto_now_add=True,
         help_text="The creation date of this nomination."
     )
+    unnominate_reason = models.TextField(
```
MarkKoz: Should this be changed to `end_reason` for consistency?

pydis_site/apps/api/models/bot/nomination.py
```diff
@@ -20,14 +20,22 @@ class Nomination(ModelReprMixin, models.Model):
...
+        help_text="Why the nomination was ended.",
+        default=""
+    )
+    unwatched_at = models.DateTimeField(
```
MarkKoz: Should this be changed to `ended_at` for consistency?

MarkKoz (changes requested):

pydis_site/apps/api/viewsets/bot/nomination.py
MarkKoz: I am a bit torn on this, so I will be fine whichever way we go with. @jchristgit thoughts?

jchristgit: I agree with merging functionality to keep this consistent with the other endpoints, especially the infraction endpoint.

MarkKoz: Is this really resolved? @jchristgit suggested to merge the endpoints. I think changing it to `PUT` was a step in the wrong direction anyway because as we discussed it's less appropriate than `PATCH`

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+    ### PATCH /bot/nominations/<id:int>
+    Update or end the nomination with the given `id` and return the updated nomination.
+
+    The PATCH route can be used for three distinct operations:
+    1. Updating the `reason` of `active` nomination;
+    2. Ending an `active` nomination;
+    3. Updating the `end_reason` or `reason` field of an `inactive` nomination.
+
+    While the response format and status codes are the same for all three operations (see
+    below), the request bodies vary depending on the operation. For all operations holds
```
MarkKoz:
```suggestion
    below), the request bodies vary depending on the operation. For all operations it holds
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+                raise ValidationError({field: ['This field cannot be set at creation.']})
 
+        user_id = request.data.get("user")
+        if Nomination.objects.filter(active=True, user__id=user_id).exists():
+            raise ValidationError({'active': ['There can only be one active nomination.']})
+
+        serializer = self.get_serializer(
+            data=ChainMap(
+                request.data,
+                {"active": True}
```
MarkKoz, line 170: How is a `ChainMap` better or different than just `dict.update()` in this case?

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+
+        # There are three distinct PATCH scenarios we need to validate.
+        if instance.active and 'active' not in data:
+            # 1. We're updating an active nomination without ending it.
+            if request.data.get('end_reason'):
+                raise ValidationError(
+                    {'end_reason': ["An active nomination can't have an end reason."]}
+                )
+
+        elif instance.active and not data.get('active'):
```
MarkKoz: I think this makes the logic easier to read as it makes it clear that it's checking the value in the key rather than also potentially checking if `get()` returned `None`.

```suggestion
        elif instance.active and not data['active']:
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+            # 2. We're ending an active nomination.
+            if 'reason' in data:
+                raise ValidationError(
+                    {'reason': ['This field cannot be set when ending a nomination.']}
```
MarkKoz, line 206: Why not? To prevent accidentally setting the wrong reason?

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
         instance = self.get_object()
         serializer = self.get_serializer(instance, data=request.data, partial=True)
         serializer.is_valid(raise_exception=True)
+
+        data = serializer.validated_data
+
+        # There are three distinct PATCH scenarios we need to validate.
+        if instance.active and 'active' not in data:
+            # 1. We're updating an active nomination without ending it.
+            if request.data.get('end_reason'):
```
MarkKoz: Does this use `get()` instead of `in` so that it also ignores empty strings?

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+                raise ValidationError(
+                    {'end_reason': ['This field is required when ending a nomination.']}
+                )
+
+            instance.ended_at = timezone.now()
+
+        elif 'active' in data:
+            # 3. The `active` field is only allowed when ending a nomination.
+            raise ValidationError(
+                {'active': ['This field can only be used to end a nomination']}
```
MarkKoz, line 219: This relies on the suggestion to change the if condition
```suggestion
                {'active': ['Cannot make an inactive nomination active again.']}
```

MarkKoz (changes requested):

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -1,32 +1,224 @@
...
+                )
+
+            if 'end_reason' not in request.data:
+                raise ValidationError(
+                    {'end_reason': ['This field is required when ending a nomination.']}
+                )
+
+            instance.ended_at = timezone.now()
+
+        elif 'active' in data:
```
MarkKoz, line 216: It should only matter when `active` is `True`. We shouldn't care if a user redundantly sets it to `False` when the nomination is already inactive.

```suggestion
        elif data.get('active'):
```

MarkKoz (changes requested):

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: I've moved the end nomination functionality to the PATCH endpoint now. There are already some comments on the logic, so I'm going to resolve this discussion to prevent the discussion from happening in two places.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: Not much, I guess, except that update would need called on a previous line since it returns `None`. I like `ChainMap`, but I'll change it.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: I don't agree with this.

While we've merged the `/end` endpoint into this `PATCH` endpoint, I still think there are two semantically different operations: Updating a reason, which has no functional/semantic consequences, and ending a nomination, which does create a functional/semantic difference. Since the `active` field can only be changed by the latter, I think that it's not right to ignore it implicitly when performing an update reason operation.

If we do allow an implicit ignore of `active=False`, we also have no way of determining the intention of the user when they send a request like :

```python
{
    'active': False,
    'end_reason': 'An end_reason',
}
```
While this may seem harmless, it creates a race condition if two different users try to end the same nomination at roughly the same time. The request that's processed first will end the nomination, setting the initial `end_reason`, but one that's processed second will overwrite the `end_reason` of the former as the endpoint has no way of determining whether or not it's an update or a request that tries to end an already ended nomination. 

Therefore, I think it's better that we invalidate requests that would update a field, but also carry a field that should never be part of an update in the first place.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: After a short conversation in dev-core, I've decided to keep the chainmap.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: Good point. `end_reason` should never be provided when updating something of an active nomination, unless it's an "end nomination" action. I'll change it to `in` so it fails whenever `end_reason` is provided.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: Yes, I'll change this. This is left-over from a previous version that was not an `elif`. That would mean that the existance of `'active'` was not guaranteed and a simple look-up could fail. The previous `if` makes that a logical impossibility, though.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: Because ending a nomination is a semantically different action than updating the original reason for the nomination. I feel that we should separate these "update the current information" from "changing the status of a nomination" operations.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: I had this earlier, but changed it because of the arguments I gave in the `elif` discussion above. Depending on the outcome of that, I'll either change this or leave it.

pydis_site/apps/api/viewsets/bot/nomination.py
MarkKoz: Okay, good point on the race condition. I propose this to make the errors clearer:

```py
active = data.get('active')
if instance.active:
    has_reason = 'end_reason' in data

    if active is False and not has_reason:
        raise ValidationError(
            {'end_reason': ['This field is required when ending a nomination.']}
        )
    elif has_reason:
        raise ValidationError(
            {'end_reason': ["An active nomination can't have an end reason."]}
        )
elif active:
    raise ValidationError(
        {'active': ["Cannot make an inactive nomination active again."]}
    )
elif active is False:
    raise ValidationError(
        {'active': ["The nomination is already inactive."]}
    )
```

pydis_site/apps/api/viewsets/bot/nomination.py
MarkKoz: I disagree with that viewpoint. It's been merged into one endpoint yet you still treat it as if its two separate ones. The way I see it, a PATCH is a PATCH - simply an update of the model. We should allow updating anything one wants so long as it doesn't break the logic of the model.

If hypothetically one wishes to edit `reason` and end a nomination simultaneously, you are suggesting they make two separate requests to the same endpoint? Forgive me, but that just seems silly to me.

SebastiaanZ: If having a single endpoint means a single type of action, then merging the two endpoints was a mistake, in my opinion. They are two very different actions with very different consequences in terms of what happens in the database.

When you update information, you're only updating the information that is currently there. When you end a nomination, you're not just updating information, but setting additional fields that for all intents and purposes were considered unset up until that point. Ending a nomination evens sets a field that's not provided by the request, namely the end date of the nomination.

When we just consider it to be all "updating fields", what actually happens in the database can't be predicted by just the request. The same request may or may not modify the `end_at` field, depending on the current and unknown status of the nomination. That doesn't feel right to me.

From a more practical point of view, if users of the endpoint intend to end a nomination, their request may be interpreted in terms of an update of `end_reason` if another request has already ended the nomination before them unknowingly. There's an easy way to prevent this by not allowing an implicitly ignored `active` change on an edit and that's how I would like to do things.

So, I don't agree with your viewpoint at all. It's leads to behavior that user cannot possibly predict.

MarkKoz: Then would you rather revert commits to have the endpoints separate? I was on the fence about that begin with, it was more @jchristgit's suggestion.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: I like the old version better, since it has a separate "top-level" conditional block for the end_nomination action, which also needs to set the `instance.ended_at` field. 

It could be included here as an additional `active is False and has_reason:` in the nested conditional block in `if instance.active` or, since it's not validation, in a separate `if` statement after the validation `if-statement, but I'm not sure if that would make it easier to follow.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: I don't really care what direction we choose, I just don't like two identical requests triggering different behavior in the back-end without the user having explicit control over it or being able to predict what's going to happen.

lemonsaurus:
As far as I can see, the outstanding review requests here are caused by differences in opinion over what approach is more idiomatic. As much as I appreciate these kinds of debates, this pull request has been open for too long. It's time we get it merged. 

This has been thoroughly reviewed and tested, and I've been assured that it's fully functional in its current state, so I'm making the executive decision to pull the plug on this review process so we can continue progressing towards feature unfreeze.

PR #234, opened 2019-08-09
Add support for certain unicode characters in ot names

MarkKoz:


pydis_site/apps/api/tests/test_off_topic_channel_names.py
```diff
@@ -88,10 +88,21 @@ def setUp(self):
...
+        names = (
+            '𝖠𝖡𝖢𝖣𝖤𝖥𝖦𝖧𝖨𝖩𝖪𝖫𝖬𝖭𝖮𝖯𝖰𝖱𝖲𝖳𝖴𝖵𝖶𝖷𝖸𝖹',
+            'ǃ？’',
+        )
+
+        for name in names:
+            response = self.client.post(f'{url}?name={name}')
+            self.assertEqual(response.status_code, 201)
+
     def test_name_in_full_list(self):
```
lemonsaurus: this test name is weird. I don't fully understand what it does without the docstring.

lemonsaurus (approved):

lemonsaurus:
this looks excellent.

pydis_site/apps/api/tests/test_off_topic_channel_names.py
MarkKoz: I agree and it doesn't look like it belong under creation test anyway. In fact, it looks quite similar to `test_returns_name_in_list`

lemonsaurus: Let's just rename it to something that makes sense, then.

MarkKoz: I was trying to say that this test looks redundant.

lemonsaurus: Aha. Okay well I'd be open to you just scrapping it, too, then.

MarkKoz:
03a08db9037904c1dafe42394c33fbc907821b5f was pushed by @jchristgit despite this PR being open (and made me realise this PR lacks a migration). What should I do with this PR? Make it a test-only PR? In that case I should probably clean up the history with a force push 👿.

SebastiaanZ:
> What should I do with this PR? Make it a test-only PR? In that case I should probably clean up the history with a force push

Let's revitalize this PR.

I think it's a good idea to have a couple of test cases in place for that regex validation. I know it's not a strict requirement for the test coverage in terms of lines touched during testing, but it would be beneficial in terms of feature coverage.

MarkKoz:
@SebastiaanZ Not sure what you mean. I already improved on the tests that were there. Did you have something more in mind?

SebastiaanZ:
@MarkKoz No, but you asked what you had to do with this PR and it was getting stale. I tried answering that question by saying that I think that your tests are still a useful addition to the site, but I guess my wording threw you off. 

It's currently conflicting with master, though, because of the now redundant changes in the model file.  I wouldn't be opposed to you cleaning up the history in this case, since I don't think it's going to complicate reviews.

MarkKoz:
This should be good to go now.

SebastiaanZ (approved):
lgtm. Tested locally and everything works.

PR #235, opened 2019-08-14
log everything into the database in addition to stdout

### LINKED ISSUE #195: Django: Put application logs into logging handler

jchristgit:
wow i managed to write 20 lines of code and it only took 5 months

closes #195

image

### DIFF
--- a/pydis_site/apps/api/tests/test_dblogger.py
+++ b/pydis_site/apps/api/tests/test_dblogger.py
@@ -10,7 +10,7 @@ from ..models import LogEntry
 class DatabaseLogHandlerTests(TestCase):
     def test_logs_to_database(self):
         module_basename = __name__.split('.')[-1]
-        logger = logging.getLogger(module_basename)
+        logger = logging.getLogger(__name__)
         logger.handlers = [DatabaseLogHandler()]
         logger.warning("I am a test case!")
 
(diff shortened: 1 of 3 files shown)


pydis_site/apps/api/tests/test_dblogger.py
```diff
@@ -0,0 +1,26 @@
...
+
+from django.test import TestCase
+
+from ..dblogger import DatabaseLogHandler
+from ..models import LogEntry
+
+
+class DatabaseLogHandlerTests(TestCase):
+    def test_logs_to_database(self):
+        logger = logging.getLogger(__file__)
```
MarkKoz:
```suggestion
        logger = logging.getLogger('testing_logger')
```

pydis_site/apps/api/tests/test_dblogger.py
```diff
@@ -0,0 +1,26 @@
...
+        logger.handlers = [DatabaseLogHandler()]
+        logger.warning("I am a test case!")
+
+        # Ensure we only have a single record in the database
+        # after the logging call above.
+        [entry] = LogEntry.objects.all()
+
+        self.assertEqual(entry.application, 'site')
+        self.assertEqual(entry.logger_name, __file__)
```
MarkKoz:
```suggestion
        self.assertEqual(entry.logger_name, 'testing_logger')
```

pydis_site/apps/api/tests/test_dblogger.py
```diff
@@ -0,0 +1,26 @@
...
+        self.assertIsInstance(entry.timestamp, datetime)
+        self.assertEqual(entry.level, 'warning')
+        self.assertEqual(entry.module, 'test_dblogger')
```
MarkKoz:
```suggestion
        self.assertEqual(entry.module, __name__)
```

MarkKoz (changes requested):
Strange to use `__file__` as a logger name since it's a path.

MarkKoz:
Right apparently the log record doesn't store the fully qualified name of the module. I guess `__name__` is out of the picture. Maybe it can parsed out? I am not keen on hard coding cause the test will break if we rename the module for whatever reason.

MarkKoz (approved):

PR #236, opened 2019-08-15
Add position field to Role model

SebastiaanZ:
To be able to easily determine a user's top role, I've added a position field to the Role model and created a `__lt__` method for Role comparisons based on their position in the Guild's role hierarchy. I've also added tests for the complete Role model. Finally, I've added a `top_role` property attribute to the User model and added a test for it to the user tests.

This depends on an update in the `django` branch: python-discord/bot#397

This a step in implementing the Deleted Message Front-end, as it allows us to determine a user's top_role, and therefore display_name color, without much hassle.


pydis_site/apps/api/models/bot/role.py
```diff
@@ -43,7 +45,15 @@ class Role(ModelReprMixin, models.Model):
         ),
         help_text="The integer value of the permission bitset of this role from Discord."
     )
+    position = models.IntegerField(
+        default=-1,
```
MarkKoz: Why does this field need to have a default value?

SebastiaanZ: To account for existing rows in the database that do not yet have a value for `position`. Another option would be to allow the field to be null, but that would mean the value returned by the ORM could be `None` and necessitates more complex logic for the `__lt__` method to account for that. (We can't compare `None` to an `int` with the `<` operator).

I figured a better alternative would be to use a default value that should be lower than the lowest role in the hierarchy (which has a position of `0`; although there's no formal lower bound in the Discord API documentation and the discord.py docs mention that `role.position` is "usually positive") and was compatible with the desired methods (meaning, able to be compared with an `int` as both the left and right operand of the `<` operator).

Hence, an integer value of `-1`.

MarkKoz: Which entries wouldn't have a position? Old ones that are imported? Apart from that I don't see why this can't be a required field.

SebastiaanZ: This migration will be applied to databases that already have entries in them that don't have a position field, so don't have a value for position. Django won't migrate if we don't provide it with a value or allow Null values. The latter would mean more complicated logic, as then it would theoretically be possible for the ORM to give back `NoneType` values instead of `int` (see above).

MarkKoz: Setting a default would make this field always optional, right? If that's the case then, if possible, it'd be better to manually give a default value during migration to only what's currently in the database so this field will still be required for new records.

SebastiaanZ: I've solved it in a different way. Basically, I first add the field with a default value, to ensure that all the existing rows have a non-null value in that column, and then alter the field to remove the default value. 

This is basically the same methodology as the intermediate, temporary field in the link @MarkKoz gave me, but without the actual temporary field to back-up existing data (which we don't have, since the field does not exist before this migration)/creating a PrimaryKey relation using that backed-up data.

MarkKoz (approved):

pydis_site/apps/api/models/bot/role.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
lemonsaurus, line 1: ? Why would we need this?

pydis_site/apps/api/models/bot/role.py
```diff
@@ -43,7 +45,14 @@ class Role(ModelReprMixin, models.Model):
...
     )
+    position = models.IntegerField(
+        help_text="The position of the role in the role hierarchy of the Discord Guild."
+    )
 
     def __str__(self):
         """Returns the name of the current role, for display purposes."""
         return self.name
+
+    def __lt__(self, other: Role):
```
lemonsaurus: this is nice, but I feel like if we're going to implement an `__lt__`, we should implement a `__gt__` as well, just for completeness sake. Probably an `__eq__` as well. I mean, whenever I see someone doing `<` to an object I expect to be able to flip it or change it to `=>` or whatever without being told that the method doesn't exist.

lemonsaurus (changes requested):
an elegant solution, with superb test coverage.

pydis_site/apps/api/models/bot/role.py
SebastiaanZ: It's to annotate the `other` parameter of the `__lt__` method with `Role`. This future import enables delayed evaluation of annotation names, which means you can annotate it with the name of the class itself without resorting to a string representation of the name. 

Without it, this: https://github.com/python-discord/site/blob/4c6c1f9cefe5f6989181716fbf92e7ce8150d013/pydis_site/apps/api/models/bot/role.py#L56

would have to be
```py
    def __lt__(self, other: "Role"):
```
as the name `Role` is not yet defined at that point.

This is behavior is built into P3.8, but needs future import in P3.7.

pydis_site/apps/api/models/bot/role.py
SebastiaanZ: The interpreter is smart enough to figure out a `>` from a `__lt__` dunder, but for checks including equality, we will indeed need `__eq__`. Roles *should* never have an equal position, though, although there's no strict "this will never happen" in the developers documentation.

I can implement the `__eq__` to cover `<`, `>`, `==`, `<=` and `>=`, though. I didn't because I did not think it would be that useful, but I will.

SebastiaanZ: On further inspection, the `Model` class we inherit from has the following method defined:

```py
    def __eq__(self, other):
        if not isinstance(other, Model):
            return False
        if self._meta.concrete_model != other._meta.concrete_model:
            return False
        my_pk = self.pk
        if my_pk is None:
            return self is other
        return my_pk == other.pk
```

Which allows us to test if two instances of the Model are referring to the same database object. Providing a custom `__eq__` would overwrite that functionality.

pydis_site/apps/api/models/bot/role.py
lemonsaurus: Ahaa. Alright, thanks. didn't realize it was 3.8.

pydis_site/apps/api/models/bot/role.py
lemonsaurus: Fair enough.

lemonsaurus (approved):

PR #238, opened 2019-08-23
Deleted Messages Front-End

### LINKED ISSUE #232: Django: Add front-end for deleted messages

SebastiaanZ:
This PR implements the front-end for deleted messages for the `django` branch and will close #232. 

Most important points:
- The URLs follow the same pattern as on the old site, except they are now moved to the `staff` subdomain, e.g., `https://staff.pythondiscord.com/bot/logs/1/`. I will PR this change to the djangobot as well. (It's a small change in the `config-default.yml`.)
- The logs page will now display the actor and the deletion date.
- Embeds are displayed similarly to how Discord displays them (see example below), with the exception of special embeds, like those embedding video players or tweets. The latter will show up as regular embeds, but all the relevant information will still be displayed.
- The username, discriminator of the users will be displayed with the correct role color.

Other changes:
- The embed validator did not accept the `inline` parameter for embed fields, so I changed that. Since `name` and `value` are mandatory for embed fields, I also added validation that checks if they are present.
- I've added a `timestamp` property to the Message model that computes the message timestamp from the snowflake id. 

Example of the front-end:
PR_DeletedMessagesFrontEnd


pydis_site/apps/api/models/bot/message.py
```diff
@@ -49,6 +54,13 @@ class Message(ModelReprMixin, models.Model):
         help_text="Embeds attached to this message."
     )
 
+    @property
+    def timestamp(self) -> datetime:
+        """Attribute that represents the message timestamp as derived from the snowflake id."""
+        tz_naive_datetime = datetime.utcfromtimestamp(((self.id >> 22) + 1420070400000) / 1000)
+        tz_aware_datetime = timezone.make_aware(tz_naive_datetime, timezone=pytz.timezone("UTC"))
```
jchristgit: I would prefer we don't import `pytz` here and use Django's `timezone.utc` directly, as [in its current source, those are equivalent](https://docs.djangoproject.com/en/2.2/_modules/django/utils/timezone/#make_aware). However, we don't directly include `pytz` as a dependency, and in case `pytz` is ever dropped from Django (for instance to use https://docs.python.org/3/library/datetime.html#datetime.timezone instead), we'll still be using what they allow us to. `make_aware` and `utc` are conveniently put in the same module, too.

```suggestion
        tz_aware_datetime = timezone.make_aware(tz_naive_datetime, timezone=timezone.utc)
```

pydis_site/apps/api/models/bot/message.py
```diff
@@ -1,6 +1,11 @@
+from datetime import datetime
+
+import pytz
 from django.contrib.postgres import fields as pgfields
 from django.core.validators import MinValueValidator
 from django.db import models
+from django.utils import timezone
+
 
 from pydis_site.apps.api.models.bot.tag import validate_tag_embed
```
jchristgit, line 8: I was expecting a trillion linters to complain here with all the plugins we have in by now, but do we not usually use a single line in these spots?

pydis_site/apps/api/models/bot/tag.py
```diff
@@ -8,17 +9,29 @@
...
     }
 
+    required_fields = ('name', 'value')
+
     for field in fields:
         if not isinstance(field, Mapping):
             raise ValidationError("Embed fields must be a mapping.")
 
+        if not all(required_field in field for required_field in required_fields):
+            raise ValidationError("Embed fields must contain both a 'name' and a 'value' field.")
```
jchristgit: I would prefer if this error message was kept generic if we add something new to `required_fields`, e.g. something like

```suggestion
            raise ValidationError(f"Embed fields must contain the following fields: {', '.join(required_fields)}.")
```

pydis_site/apps/staff/templatetags/deletedmessage_filters.py
```diff
@@ -0,0 +1,17 @@
...
+
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def hex_colour(color: int) -> str:
+    """Converts an integer representation of a colour to the RGB hex value."""
+    return f"#{color:0>6X}"
```
jchristgit, line 11: very nice

pydis_site/apps/staff/viewsets/logs.py
```diff
@@ -0,0 +1,33 @@
...
+        messages = message_context.deletedmessage_set.all()
+
+        template_fields = {
+            'actor': actor,
+            'actor_colour': message_context.actor.top_role.colour,
+            'creation': creation,
+            'messages': messages
+        }
+
+        return render(request, self.template_name, template_fields)
```
jchristgit: This should use a generic view, specifically `DetailView`. You can add extra template context if needed as shown here: https://docs.djangoproject.com/en/2.2/topics/class-based-views/generic-display/#adding-extra-context , although from what I can tell you don#t need that here in the first place (just the context will be enough)

pydis_site/apps/api/models/bot/message.py
SebastiaanZ: Yes, we do. I'll remove the newline, but I'm not exactly sure why the linters don't complain. It may be worth investigating if we force them to detect this.

pydis_site/apps/api/models/bot/message.py
SebastiaanZ: That's a good point; I'll change it.

pydis_site/apps/api/models/bot/message.py
```diff
@@ -1,6 +1,11 @@
+from datetime import datetime
+
+import pytz
```
SebastiaanZ:
```suggestion
```

pydis_site/apps/api/models/bot/tag.py
SebastiaanZ: Good point.

pydis_site/apps/staff/viewsets/logs.py
SebastiaanZ: Thanks, I'll change this.

One of the reasons I created custom template fields was to avoid the deep nesting the messages had, but since I only refer to it once in the template in the `for`-loop, it's probably better not to do that and just go with the default context. That makes maintaining this a lot easier and reduces the view to just a couple of lines.

pydis_site/apps/staff/tests/test_logs_view.py
```diff
@@ -0,0 +1,160 @@
+from django.test import Client, TestCase
+from django.utils import timezone
+from django_hosts.resolvers import reverse, reverse_host
+
+from pydis_site.apps.api.models.bot import DeletedMessage, MessageDeletionContext, Role, User
+from ..templatetags.deletedmessage_filters import hex_colour
```
lemonsaurus: Feels like this could be an absolute import?

pydis_site/apps/staff/tests/test_deletedmessage_filters.py
```diff
@@ -0,0 +1,21 @@
...
+from ..templatetags import deletedmessage_filters
+
+
+class DeletedMessageFilterTests(TestCase):
+    def test_hex_colour_filter(self):
+        self.assertEqual(deletedmessage_filters.hex_colour(0), "#000000")
+        self.assertEqual(deletedmessage_filters.hex_colour(255), "#0000FF")
+        self.assertEqual(deletedmessage_filters.hex_colour(65280), "#00FF00")
+        self.assertEqual(deletedmessage_filters.hex_colour(16711680), "#FF0000")
+        self.assertEqual(deletedmessage_filters.hex_colour(16777215), "#FFFFFF")
```
lemonsaurus: maybe some constants would help this look a bit less insane.

lemonsaurus (approved):
Excellent work, only a few minor thoughts.

pydis_site/apps/staff/tests/test_logs_view.py
SebastiaanZ: Sure. I'll change it. I've used a relative one because it's a within-app import, but an absolute one makes it more readable.

pydis_site/apps/staff/tests/test_deletedmessage_filters.py
SebastiaanZ: I've changed it to use an IntEnum of the colors.

lemonsaurus:
Looks great. We can probably merge this now.

SebastiaanZ:
> Looks great. We can probably merge this now.

I think so, too. Since you left comments earlier, what do you think, @jchristgit?

PR #242, opened 2019-09-12
Add missing migrations file for the deleted messages model

SebastiaanZ:
I've noticed that I forgot to commit a migration for the deleted messages model, while the change to the model itself has been committed and merged. Hence, Django will complain that the change is not yet reflected in a migration.

This PR is meant to correct that oversight on my part.


jchristgit:
Thanks!

PR #250, opened 2019-09-17
Refine Dockerfiles

scragly:
I've done some refinement for the Dockerfiles. This has resulted in the following improvements:
- Build time is reduced from 1:08 to 0:50 when tested on my machine.
- Image size is reduced from 549.2 MB to 466.2MB
- Image size is further reduced to 440.2MB thanks to @MarkKoz's suggestion of disabling Pip cache.

**Change Details**
- `Dockerfile.local` has been changed to `local.Dockerfile` for better filetype detection.
  
- Changed from interactive `adduser` command that resulted in logged warnings to the non-interactive `useradd` command.
  - `--system` makes it a homeless non-login user.
  - `--shell` has been set to avoid possible security issues.
- `uwsgi` is now installed from the system package repository, as it is already built and actively updated, using the exact same version we did.
  - Since it's already built, previous build packages aren't installed.
  - Since build packages aren't installed, purging is removed (but as a note, this didn't reduce the container size anyway).
- `install_packages` is used instead of `apt-get` as it does cleanup, handles any prompts and doesn't install recommends automatically.
- `COPY` for Pipfiles and project files are now done at the same time.
- `RUN` commands have been tidied to use shorter references as the entrypoints are confirmed to be for the same locations.
- `pipenv` install has been moved to same `RUN` as `setuptools`


docker/app/local.Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY . .
+
+# Update setuptools by removing egg first, add other dependencies
+RUN rm -r /opt/bitnami/python/lib/python3.*/site-packages/setuptools* && \
+    pip install --no-cache-dir -U setuptools pipenv
+RUN pipenv install --system --deploy
+
+RUN SECRET_KEY=placeholder DATABASE_URL=sqlite:// python3 manage.py collectstatic --no-input --clear --verbosity 0
+
+CMD ["uwsgi", "--ini", "docker/app/uwsgi.ini"]
```
MarkKoz, line 26:
```suggestion
ENTRYPOINT ["uwsgi"]
CMD ["--ini", "docker/app/uwsgi.ini"]
```

MarkKoz (commented):
Looks good. None of the Python dependencies need to be built with GCC, right?

scragly:
Correct, it was only uwsgi. I was sure to actually run the resulting containers to ensure things still worked _just_ in case I was mistaken beforehand.

docker/app/local.Dockerfile
scragly: I'm curious, what's the difference?

MarkKoz:
I think doing `ENV PIP_NO_CACHE_DIR=false` would be helpful since that'd also affect the dependencies pipenv installs, right? Could throw in a couple more to perhaps make logs nicer too:

```
ENV PIP_NO_CACHE_DIR=false \
    PIPENV_HIDE_EMOJIS=1 \
    PIPENV_NOSPIN=1
```

docker/app/local.Dockerfile
MarkKoz: It lets you override the arguments when you use `docker run` (or in compose I suppose) without having to also write `uwsgi`. I suppose in this case not as useful, but generally you can set the executable + stable args in `ENTRYPOINT` and then put the args that may change in `CMD`.

MarkKoz: If this was done, the entrypoint still wouldn't be "forced". If you really wanted to change it you could use `docker run --entrypoint`

scragly: It doesn't work with the docker-compose `command` option, exiting with a code 1. Since there's no significant benefit and it's a difference in behaviour, I'll revert to original.

MarkKoz: That's cause it tries to do `uwsgi migate_and_serve.sh`. In the compose you'd need to make that shell script the entrypoint instead of a command.

scragly: Exactly, which is an unnecessary change in behaviour with no significant benefit.

scragly:
@MarkKoz Added the env vars suggested. It hasn't impacted ~~size or~~ build time, and confirmed it doesn't have any issues building and running the container with them.

Edit: I looked at wrong image details for size. It dropped to 440.2MB :tada:

MarkKoz:
It's 440 MB if the cache is disabled, 466 MB if it is enabled. Are you sure you built the right Dockerfile when testing that? I noticed you only set those environment variables for `local.Dockerfile`

scragly:
Yep, built the correct one, but just didn't look at the right image when checking the size. You're right, it's decreased again, which is great news.

I'll add them to the other container also.

MarkKoz (approved):

jchristgit (approved):
rock solid

jchristgit:
Thanks!

PR #251, opened 2019-09-19
Update linting dependencies & relint

sco1:


pydis_site/apps/api/models/bot/tag.py
```diff
@@ -15,7 +15,7 @@ def is_bool_validator(value: Any) -> None:
         raise ValidationError(f"This field must be of type bool, not {type(value)}.")
 
 
-def validate_tag_embed_fields(fields):
+def validate_tag_embed_fields(fields: dict) -> None:
```
SebastiaanZ, line 18: Do we want to go with a generic `dict` annotation for these types or the more specific `Dict[str, Union[str, bool]]`? It's probably a bit much to go down to this level of specificity for such a large linting PR, but it's something we can consider as a general guideline for regular contributions/PRs.

```suggestion
def validate_tag_embed_fields(fields: Dict[str, Union[str, bool]]) -> None:
```

pydis_site/apps/api/models/bot/tag.py
```diff
@@ -42,7 +42,7 @@ def validate_tag_embed_fields(fields):
                 validator(value)
 
 
-def validate_tag_embed_footer(footer):
+def validate_tag_embed_footer(footer: Any) -> None:
```
SebastiaanZ: Same as for the previous comment, according to the [Discord documentation](https://discordapp.com/developers/docs/resources/channel#embed-object-embed-footer-structure), all items will be mappings of `str => str`. This could be:

```suggestion
def validate_tag_embed_footer(footer: Dict[str, str]) -> None:
```

pydis_site/apps/api/models/bot/tag.py
```diff
@@ -67,7 +67,7 @@ def validate_tag_embed_footer(footer):
             validator(value)
 
 
-def validate_tag_embed_author(author):
+def validate_tag_embed_author(author: Any) -> None:
```
SebastiaanZ, line 70: As per [Discord docs](https://discordapp.com/developers/docs/resources/channel#embed-object-embed-author-structure) & like earlier comments:

```suggestion
def validate_tag_embed_author(author: Dict[str, str]) -> None:
```

pydis_site/apps/api/serializers.py
```diff
@@ -106,7 +106,7 @@ class Meta:
             'id', 'inserted_at', 'expires_at', 'active', 'user', 'actor', 'type', 'reason', 'hidden'
         )
 
-    def validate(self, attrs):
+    def validate(self, attrs: dict) -> dict:
```
SebastiaanZ, line 109: I've logged some output and apparently we're passing getting and returning an `OrderedDict` here:

```py
OrderedDict([('active', True), ('user', <User(avatar_hash='cbde0f631413b133fd219bf6c6caa90b' discriminator=4641 id=532614300519497729 in_guild=True name="Hemlock's Cat")>), ('actor', <User(avatar_hash='3a305b2d34f677dd1c050bc98dd8e390' discriminator=3787 id=336843820513755157 in_guild=True name='Ves Zappa')>), ('type', 'warning'), ('reason', 'you kitty.'), ('hidden', False)])
```
We could still just go with `dict`/`OrderedDict` here or `OrderedDict[str, Union[User, str, bool, int]]` although the latter gets ugly quick and may not improve readability (which is, I think, the main reason we add annotations in the first place).

.flake8
```diff
@@ -10,11 +14,7 @@ ignore=
...
+    # Type Annotations
+    TYP002,TYP003,TYP101,TYP102,TYP204,TYP206
 per-file-ignores =
-    **/tests/**:D101,D102,D103,D105,D106,S106
-
-application_import_names=pydis_site
-exclude=__pycache__, venv, .venv, **/migrations
-import-order-style=pycharm
-max-line-length=100
+    **/tests/**:D101,D102,D103,D105,D106,S106,TYP
```
MarkKoz: Is there a reason to specify each one?
```suggestion
    **/tests/**:D1,S106,TYP
```

pydis_site/apps/api/models/bot/tag.py
```diff
@@ -93,7 +93,7 @@ def validate_tag_embed_author(author):
             validator(value)
 
 
-def validate_tag_embed(embed):
+def validate_tag_embed(embed: Dict[str, Union[str, List[dict], dict]]) -> None:
```
MarkKoz: This JSON type is not comprehensive. For the sake of simplicity, I think it will be better to use `Union[List, Dict]`,  or even just `Dict` if we follow the line of thinking that the Discord API does not return a list for Embed objects.

MarkKoz (changes requested):

pydis_site/apps/api/models/bot/tag.py
MarkKoz: In this case, normally I would say a more detailed type annotation is reasonable. But if the context is considered, this function is meant to validate data so it may accept any data type. It even checks that each field is of type `Mapping` so the only implicit type constraint this function has is that `fields` is iterable. Arguably, even that should be validated by the function though I'm not sure how it all ties together.

.flake8
sco1: It’s not each one

MarkKoz: Does it not overlap with with the normal ignores? I thought it did and in that case it would indeed be all the ones that start with `D1`.

pydis_site/apps/api/models/bot/tag.py
SebastiaanZ: My reasoning went along the lines of:

- We're validating embed fields here, which, according the Discord API reference, should always have the type `Dict[str, Union[str, bool]]`.
- The tag validation function itself assumes that you pass an object of this form as well.

However, I think your point is that the purpose of a validation function is to validate if the object we pass in holds those and more stringent constraints. Honestly, I think we/I may be over-complicating this a bit, which is also why I originally left these comments as comments and not required changes.

I'm going to resolve these comments for now; if @MarkKoz or someone else really wants to see a different annotation here, feel free to unresolve the conversation and continue it.

.flake8
sco1: You’re right, sorry, I missed that this was the per-file line

SebastiaanZ (approved):

pydis_site/apps/api/models/bot/tag.py
sco1: I'm not really familiar with how Django's bits go together, isn't this validating what's coming out of our DB rather than what's coming from Discord?

SebastiaanZ: We're validating the data that's being received by the API, which obviously comes from the bot and not directly from Discord. When it comes to embeds, it's a `json` representation of the `to_dict` method of `discord.Embed` from discord.py.

MarkKoz: I think `Dict` is good enough here and I agree with @SebastiaanZ we shouldn't over-think this. But along the the lines of the other comment I made, I may as well make the point that since this is a validation function and it also does `isinstance(..., Mapping)`, I think a more technically correct type would be `Any` or something along those lines.

sco1: So...

```py
def validate_tag_embed(embed: Any) -> None:
```
?

MarkKoz: I'm fine with that. If you or anyone else doesn't object then sure.

sco1: @jchristgit what are your thoughts?

MarkKoz (approved):

PR #257, opened 2019-09-21
Readdressing Docker Refinement

scragly:
The last docker refinement resulted in uwsgi being installed from apt, which ended up causing an issue where the app was not discovered when the container was run. Due to that, it was set back to the original install through pip and building from source. 

This PR applies another method instead, while also looking into the last few things that come to mind that would be able to improve the image speed and size.

### uWSGI
A more python-friendly distribution package that is maintained by the original team, `pyuwsgi`, was found after some looking around and presented an excellent and reliable way to ditch the build dependencies. 

The package is already built and has a standard whl available from pypi, allowing the build tools to be removed again while being an exact in-place install with the same behaviour as the original pip package.

### Base Image
I took this chance to finally directly compare and evaluate the two mini-debian base images that were in discussion; `bitnami/python:3.7-prod` and the official dockerhub `python:3.7-slim` images.

The result was that `python:3.7-slim` had an approx 15% quicker build time, and the size was 4% smaller during the tests, so I've changed the base image over to it instead.

### Wiki & Git
Git was taking a fair amount of time and space, and it appeared to be only needed for the git dependancy `wiki`, our _temporary_ fork of the django-wiki repo for Django 2.2 compatibility.

Once the main lib releases a package to pypi that has Django 2.2 compatibility, we will be removing this fork and using the normal package again. Since we're not about to change the version until then, I've added a .whl for our forked version and have it set in the Pipfile and Pipfile.lock.

_The `build-wiki.Dockerfile` and `build-wiki.sh` files are only helpers for making the wiki whl file that I used in laziness and are not used for testing or production. They are simply here for reference and convenience._

### Improvements
The uncompressed size has dropped from 516.7MB to 310.9MB
The local build speed has dropped from 1:11 to 38s


docker/app/local.Dockerfile
```diff
@@ -1,27 +1,25 @@
...
 WORKDIR /app
 COPY . .
 
-# Update setuptools by removing egg first, add other dependencies
-RUN rm -r /opt/bitnami/python/lib/python3.*/site-packages/setuptools* && \
-    pip install -U setuptools
-RUN pip install pipenv uwsgi
-RUN pipenv install --system --deploy
+# Install project dependencies
+RUN pip install -U pipenv pyuwsgi && pipenv install --system --deploy
```
MarkKoz: Would moving `RUN pip install -U pipenv pyuwsgi` above `COPY` help with layer caching?

scragly: Yeah likely. Good idea.

scragly: This is now actioned.

MarkKoz (approved):

scragly:
No downtime this time, and best of all, the entire CI for a merge to master took 3:54, down from an avg time of over 7 minutes.

image

PR #261, opened 2019-09-26
Migrate hidden warnings to notes and add additional validation

### LINKED ISSUE #260: Migrate hidden warnings to notes
The new `Infraction` model has a [specific type field option](https://github.com/python-discord/site/blob/master/pydis_site/apps/api/models/bot/infraction.py#L8-L19) for notes that we recently started using (python-discord/bot#452). This means that we're now storing notes as `"note"` in the database and no longer use the misnomer of "hidden warning".

However, there are still a lot of `"hidden warning"` infractions in the database that were set prior to merging the PR referenced above. I think it's a good idea to perform a data migration to turn these `"warning"` infractions into `"note"` infractions:

```py
from django.db import migrations


def migrate_hidden_warnings_to_notes(apps, schema_editor):
    """Migrates hidden warnings to notes."""
    Infraction = apps.get_model('api', 'Infraction')

    for infraction in Infraction.objects.filter(type="warning", hidden=True):
        infraction.type = "note"
        infraction.save()


class Migration(migrations.Migration):

    dependencies = [
        ('api', '0042_infraction_add_default_ordering'),
    ]

    operations = [
        migrations.RunPython(migrate_hidden_warnings_to_notes),
    ]
```

This will mean that notes will now consistently be called "note", both in the back-end as the front-end (bot).

---

In addition to migrating the data, I think it's a good idea to decouple the terms "warning" and "note" completely by adding validation rules for infractions that state that `"warnings"` can never have `hidden=True` and `"notes"` can never have `"hidden=False"`. This way, they truly become different entities, instead of having a "hidden warning" as some sort of implicit synonym for notes.

We already do this with for other incompatible combinations of field values, like a `"kick"` infraction with an expiration date or a `"superstarify"` infraction with `hidden=True`, so this would be merely an extension of the current validation rules instead of introducing validation completely to the infraction endpoint.

SebastiaanZ:
The Django rewrite introduced a new type of infraction: the `note` infraction; this new infraction type replaced the misnomer "hidden warnings". To retroactively apply this to historical infractions, the Django migration script stored all infractions that were stored as "hidden warnings" in the old database as notes in the new database. So far, so good.

Unfortunately, the new Django bot still sent notes as "hidden warnings" to the API, until this was corrected by python-discord/bot#452. This means that in the time between the migration and that PR, notes were still stored as warnings and not as notes. 

This pull request rectifies that by using a data migration to turn "hidden warning" infractions into "note" infractions. In addition, it adds validation to the API that makes sure we reject "hidden warnings" or "non-hidden notes". 

This closes #260


MrHemlock (approved):
Tested it on my env and it adjusted the hidden warning I had previously.  Also I don't believe we have to worry about any hidden warnings or the like as there isn't a way to force an infraction to be one way or the other besides via the Admin page on the site.  And the bot itself has been corrected so that Notes are now the proper type.

lemonsaurus (approved):

PR #265, opened 2019-09-27
Custom `manage.py` Entry Point Script

scragly:
I've done up a custom `manage.py` to suit a few custom purposes and to help aid a simplified development setup.

The SiteManager will handle preparing the server for both development and production environments, ensuring appropriate preparations are only run on the correct environments.

## Development Environment
```
manage.py run --debug
```
### Preparation steps
1. DEBUG env var added
1. Wait until PostgreSQL database is accepting connections
1. A super user is created for website with `admin:admin` as default.
   - User can set custom super user details with env vars:
     - `SUPER_USERNAME`
     - `SUPER_PASSWORD`
1. Run `migrate`
1. Run `collectstatic`
1. Run the Django development server on port 8000.

### Fixes
- Avoid `migrate` and `collectstatic` running twice on first launch due to development web server `StatReloader`.

## Production Environment
```
manage.py run
```
### Preparation steps
1. Run `migrate`
1. Run `collectstatic`
1. Run the uwsgi production server.

## Dockerfile & Compose
- `manage.py` is the entry point now for Dockerfile as almost any Django command would need to be run through it.
- `bot` docker will connect to the same project stack, so will use the internal docker domain `web` to reference the container on the network. Aliases have been added for relevant subdomains and the new domain aliases added to ALLOWED_HOSTS when in debug mode. 

## File Changes
- `Dockerfile` and `local.Dockerfile` were no longer distinct, so both environments use the one file.
- `docker/apps` directory is now the `docker` directory.


manage.py
```diff
@@ -1,21 +1,153 @@
...
+        # Setup a default bot token to connect with site API
+        from rest_framework.authtoken.models import Token
+        token, is_new = Token.objects.update_or_create(user=user)
+        if token.key != bot_token:
+            token.delete()
+        token, is_new = Token.objects.update_or_create(user=user, key=bot_token)
+        if is_new:
+            print(f"New bot token created: {token}")
+        else:
+            print(f"Existing bot token found: {token}")
```
MarkKoz, line 82: Won't it replace the token if the existing token does not match? Does this really mean "existing bot token found but it's been replaced anyway"?

manage.py
```diff
@@ -1,21 +1,153 @@
...
+        self.debug = "--debug" in args
+        self.silent = "--silent" in args
+
+        if self.silent:
+            self.verbosity = 0
+        else:
+            self.verbosity = 2 if "--verbose" in args else 1
+
+        if self.verbosity and self.debug:
+            os.environ.setdefault("DEBUG", "true")
```
MarkKoz, line 52: Why would setting this also be dependent on verbosity? Does this environment variable affect verbosity?

manage.py
```diff
@@ -1,21 +1,153 @@
...
+        database_url = os.environ["DATABASE_URL"]
+        match = re.search(r"@(\w+):(\d+)/", database_url)
+        if not match:
+            raise OSError("Valid DATABASE_URL environmental variable not found.")
+        domain = match.group(1)
+        port = int(match.group(2))
+
+        # Attempt to connect to the database socket
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        while True:
```
MarkKoz: What would happen if the container crashed? Would this just perpetually wait or would it raise an exception that isn't `socket.error`?

manage.py
scragly: `is_new` covers all cases where the default token doesn't match or doesn't exist in the first place.

`else` is only if the token actually is a match and truly wasn't recreated.

manage.py
scragly: If the PGSQL container failed to start, it would keep waiting and outputting `retrying` as the socket only doesn't error when the port is open, meaning the PGSQL container is fully ready to accept connections.

This also is only done during dev environments, so it's a good reminder that they may have forgotten to up the other container (dunno why they're running `docker-compose up site` in that case though), or they severely broke something.

manage.py
scragly: It's just ensuring that `--silent` isn't used when developing.

scragly: Now that I'm trying to remember the exact thought process that made me write it, I actually can't recall the detailed reason. You're right, that part could probs just be removed. I'm going to do a big more digging in the Django docs just in case though.

scragly: I think it might have been at a time when I was going to make debug the default. Adjusted now though anyhow, thanks!

manage.py
MarkKoz: Do you think it would be good to add a timeout and exit with return code 1?

scragly: Hmm, I'm not sure, as it's pretty clear what it's doing and it's simple to interrupt. I guess though if they run the container detached and it has that issue then it would just keep going and going I guess until they realise by manually checking logs, so for just that possible edge case, it might make sense.

scragly: On testing, I realised it's actually impossible to launch web alone via compose thanks to `depends_on` anyway, so the only case is if something is really really broken somehow. At any rate, I've added and tested the retry maximum.

MarkKoz (approved):

lemonsaurus (approved):
This looks excellent, my only comments are what @MarkKoz already caught and you fixed already.

PR #268, opened 2019-09-30
Monkeypatch in new HEADER_ID_PREFIX.

### LINKED ISSUE #267: Ugly prefix on all ID links.
Currently, all the headers that are created by the wiki will have id's that are prefixed with `wiki-toc`. As such, when you want to link a header, the link will look something like https://pythondiscord.com/pages/contributing/site/#wiki-toc-development-environment.

It would be better if this simply said `#development-environment`, so let's change that.

### LINKED ISSUE #276: updating/upgrading contributing tutorial
Not sure if it the right place, but:

This page: https://pythondiscord.com/pages/contributing/bot/ could be better (although it is already pretty good)

Rename `#bot-commands`and `#ot-offtopics` to `bot` and `off-topic-0` to actually meet with their names in `config.yml`. Explicit more that the `bot_token` has to be put in the `.env`file.

Precise somewhere that not all cogs are loaded in `DEBUG` mode, I lost one hour trying to figure why `verification.py` was not loading.

What do you think?

lemonsaurus:
Currently, all toc-links will have a wiki-toc prefix. This commit removes this prefix, so that we can header link with just something like `#environment` instead of `#wiki-toc-environment`.

There's no option for this in the `django-wiki`, so this monkeypatch is the best solution I could find. I'm not super happy with it, though, so if anyone has a better solution, do let me know.

Closes #267.

### DIFF
--- a/pydis_site/__init__.py
+++ b/pydis_site/__init__.py
@@ -1,0 +1,4 @@
+from wiki.plugins.macros.mdx import toc
+
+# Remove the toc header prefix. There's no option for this, so we gotta monkey patch it.
+toc.HEADER_ID_PREFIX = ''


SebastiaanZ:
I came to the same conclusion as you when I looked at it briefly yesterday. I'm not sure why the authors of the plugin did not make this configurable, since it seems to be just a single constant. I wonder if there's some hidden limitation we're overlooking, although there doesn't appear to be an obvious one.

SebastiaanZ (approved):
Looks good to me.

lemonsaurus:
:warning: Heads up! :warning: 

A bunch of these ID links have been hardcoded into our guides, so if you're going to merge this, these hardcoded URLs will stop working. It would therefore be great if you, whoeever you are, could take responsibility for getting these URLs updated after you hit that Merge button.

scragly (approved):

PR #269, opened 2019-09-30
Migrate nominations to the new Nomination model

### LINKED ISSUE #247: Django: Writing a migration that turns infractions into Nominations.
@jchristgit will write a migration to turn all "infraction nominations" into real Nomination objects. This has to be done after @scragly finishes https://github.com/python-discord/organisation/issues/134

SebastiaanZ:
Before the migration to Django, we stored information on a nomination, like the `reason` and `end_reason`, as note infractions with a special prefix in the infractions table. We've since decided that nominations shouldn't be categorized under infractions and have create a separate model for nominations in the new Django back-end. This PR migrates the historical nomination data still stored in the infractions table to the new nominations table by using a data migration.

The data migration takes the following steps:

1. Query all nomination data from the infractions table, ordered ascending by the insertion date.
2. Replay the queried data  in chronological order:
   - Add the start or the end of the nomination to new Nomination model.
   - Remove the now obsolete infraction object.

*Note: I've manually checked and prepared all the nomination data I could locate in the infractions table. However, very early nominations may not have had a `reason`, since we did not enforce that from the start; those may be "lost" to us. In addition, some nominations may not have received an `end_reason` when ending the nomination; these nominations will appear as `active` incorrectly. (I've checked, I think there's only one of those, but this is easy/easier to fix after the data migration.)*

This PR also adds a more useful string representation for nomination objects: Nominations objects will now be displayed using the target user's username and the current status of the nomination (active vs. ended). This makes it easier to locate the correct nomination in Django Admin.

This PR closes #247


MarkKoz:
> In addition, some nominations may not have received an end_reason when ending the nomination

Wouldn't that make all end reasons be skewed in the migration? As in, the wrong end infraction would be used to make a nomination inactive. Or is this not the case because the users would have to match? If so, then it would be really be an edge case for something like that to cause a skew, right?

pydis_site/apps/api/migrations/0044_migrate_nominations_from_infraction_to_nomination_model.py
```diff
@@ -0,0 +1,64 @@
...
+    Infraction = apps.get_model('api', 'Infraction')
+    Nomination = apps.get_model('api', 'Nomination')
+
+    all_nominations = (
+        Q(reason__startswith="Helper nomination:") | Q(reason__startswith="Unwatched (talent-pool):")
+    )
+
+    for infraction in Infraction.objects.filter(all_nominations).order_by('inserted_at'):
+        if infraction.reason.startswith("Helper nomination:"):
+            if Nomination.objects.filter(user=infraction.user, active=True).exists():
```
MarkKoz, line 26: Would this situation arise from this?

> In addition, some nominations may not have received an end_reason when ending the nomination; these nominations will appear as active incorrectly.

MarkKoz: On second though I don't think so since they wouldn't even have matched the reason filter, right?

SebastiaanZ: > Would this situation arise from this?
> 
> > In addition, some nominations may not have received an end_reason when ending the nomination; these nominations will appear as active incorrectly.

No, it does not arise from it. Currently, if someone has received a first nomination that was ended before we recorded "end reasons" and then received a new nomination, the new nomination would be rejected, since someone can't have two active nominations at the same time. If we'd then observe an "end reason", the first nomination would be ended with that reason. We'd end up in the correct *state* (no active nomination), but with some loss of intermediate information.

It's a bit of an edge case and it does not happen in the historical data we have, but another solution would be to close the first active nomination when we observe a new one with a `"(no reason recorded)"` and then add the second one, which means the end reason we then observe will matched to the correct reason. 

This has another edge case that may occur, though, and one that actually did happen in our historical data: For a brief period of time, if two people nominated someone more or less simultaneously, the second nomination would be recorded in the same way as the first (as a new active nomination), with no end reason between those two. With the above "solution", this would make it seem as if the individual had been nominated multiple times, being rejected for the first nominations, while in reality it was just one "nomination period" and multiple people were excited about them. (This mainly happened when a helper nominated someone that was already in talent-pool, since they had no way of knowing that.)

Ultimately, though, I've checked all historical data and we're talking about one or two individuals who's nominations resulted in the Helper role and the current migration just makes it look like they were nomination once and the end reason spells out they were accepted and offered the role. The only thing we're missing out on is that a second individual also nominated them during their nomination period.

SebastiaanZ:
> > In addition, some nominations may not have received an end_reason when ending the nomination
> 
> Wouldn't that make all end reasons be skewed in the migration? As in, the wrong end infraction would be used to make a nomination inactive. Or is this not the case because the users would have to match? If so, then it would be really be an edge case for something like that to cause a skew, right?

No, all nominations are user specific, so skews may occur, but only within the same individual. This does not happen in the historical data we have, though. What does happen is that one or two nominations were ended before we started recording end reasons, so they'll appear as open.

The only skew that may happen is this intra-individual one:

- "joseph" receives a nomination, recorded "start reason"
- "joseph" was not accepted, nomination stopped before we started recording end reason. This means that first nomination appears as open in our algorithm above, since we did not encounter an "end reason" in the data.
  - "joseph" receives a new nomination. The current migration says "that's a second one, we're going to skip this one, because they're still nominated"
- "joseph" second nomination was ended, this time we did record a reason, and it will be attached first, not the second, nomination, since the migration still has the first as active.

This could be solved by ending the first nomination with a "no reason recorded" when we encounter a second nomination, but this leads to the other edge case that does actually happen in our historical data described in the other comment.

pydis_site/apps/api/migrations/0044_migrate_nominations_from_infraction_to_nomination_model.py
MarkKoz: Do I understand correctly?

* If `"(no reason recorded)"` approach is used, it may end up in keeping a duplicate start nomination because of that bug you described with simultaneously nominating someone.
* If the current approach is used, it can result in a skew as you described in the other comment, but in practice, this won't happen given our historical data
* This check for if an active infraction already exists is not just there as a safety measure but actually has a use for the simultaneous nomination bug to discard the duplicate.

SebastiaanZ: Yes, each additional nomination while the first is still active would instead appear as a separate new nomination. For the two users this is relevant, it would appear as if they were nominated on three or four separate occasions instead of the one nomination period both actually have had.

MarkKoz (approved):
This looks good and makes sense. I can't test it but I trust in your methodology.

jchristgit (approved):
solid

PR #274, opened 2019-10-03
Django Allauth (#201)

### LINKED ISSUE #201: Django: Oauth integration

gdude2002:
This pull request implements Django Allauth as described in #201.

Fixes #201.

---

This is a fairly big PR which, I think, may have had a little scope creep. Here's what it changes:

* Add `django-allauth` dependency
* Implements Discord OAuth logins using `django-allauth`
* Exposes messages from the Django messages framework as slide-in Bulma notifications at the top right of the page, positioned just below the navbar
* Re-styles Bulma notifications when used in the context of the Django messages framework
* Redefines Bulma's bundled colours, in order to make them less loud and make Bulma calculate colours based on the primary blurple better (by redefining turquoise to blurple)
* Implements some logic to synchronise Discord roles and Django permissions groups based on signals emitted by various events

---

I would super love some reviews for this, particularly when it comes to the tests I wrote for the signal handlers, and the signal handling itself.

I'm not super happy with the signal handlers test - it's more setup than test! If anyone has any ideas on how to clean that up, I'd appreciate it.

---

Old TODO:

- [x] Set up signal listeners for Allauth events
- [x] Set up signal listeners for relevant Model save events
- [x] Document the signal system in comments/docstrings
- [x] Fix up message system
    - Move messages to a corner
    - Restyle messages
        - Discord embed-style messages with the left border?
        - More muted colours? Perhaps shades of the primary blurple.
- [ ] ~~Profile page with GH connect/disconnect and account delete~~
- [ ] ~~Create templates for the Allauth-provided account pages~~
- [x] Review the Allauth URLs to ensure we're only using what we need
    - ~~Ensure login page explains the data we're collecting and collects explicit consent~~
- [x] Create a user profile page (or just send them back to the homepage)
- [x] Figure out how to signpost/link to this system from the rest of the site


gdude2002:
Super happy to say that this PR is now **ready for review**! Here are some screenshots and videos that showcase the changes I've made.

**Login link in dropdown**

Login link in dropdown

**New Bulma colours, shown here as notification styles**

Bulma Colours

**Django admin, showing a role mapping**

Role mapping

**Django admin, showing my Discord account, sans email**

image

**[MP4 showing the login flow](https://cdn.discordapp.com/attachments/417755308891439124/630500812736561152/login-flow-2.mp4)**

pydis_site/apps/home/apps.py
```diff
@@ -1,7 +1,38 @@
...
+            data. We do this so that we never have to store it.
+
+            This is monkey-patched because most OAuth providers - or at least the ones we care
+            about - all use the function from the base Provider class. This means we don't have
+            to make a new Django app for each one we want to work with.
+            """
+            data["email"] = ""
+            return data
+
+        Provider.extract_extra_data = extract_extra_data
```
jchristgit, line 38: shouldn't this subclass, call super(), and then strip the email, in case allauth does something useful here?

pydis_site/apps/home/signals.py
```diff
@@ -0,0 +1,159 @@
...
+    which are essentially just any class at all, and Receivers can filter the Signals they listen
+    for by choosing a Sender, if required.
+
+    Signals themselves define a set of arguments that they will provide to Receivers when the
+    Signal is sent. They are always keyword arguments, and Django recommends that all Receiver
+    functions accept them as `**kwargs` (and will supposedly error if you don't do this),
+    supposedly because Signals can change in the future and your receivers should still work.
+
+    Signals do provide a list of their arguments when they're initially constructed, but this
+    is purely for documentation purposes only and Django does not enforce it.
```
jchristgit, line 32: It's nice that you document this, but I think it would be better to just provide a link to the Django documentation for these three paragraphs

pydis_site/apps/home/signals.py
```diff
@@ -0,0 +1,159 @@
...
+
+        if not isinstance(provider, DiscordProvider):
+            return
+
+        try:
+            user: DiscordUser = DiscordUser.objects.get(id=int(account.uid))
+        except DiscordUser.DoesNotExist:
+            return
+
+        self._apply_groups(user, account, True)
```
jchristgit: The kwarg is a bit hard to figure out without reading the function, would be nice to

```suggestion
        self._apply_groups(user, account, deletion=True)
```

pydis_site/apps/home/apps.py
gdude2002: Allauth actually does nothing at all here, just `return data`. I figured this would be better for future work, since you can't actually tell Allauth which class to use - although I guess I could replace the entire class? I feel like that'd break things, though. I dunno.

What do you think?

pydis_site/apps/home/signals.py
gdude2002: I agree with you, @lemonsaurus asked me to document it though.

pydis_site/apps/home/signals.py
```diff
@@ -0,0 +1,159 @@
...
+                try:
+                    new_groups.append(mappings.get(role=role).group)
+                except RoleMapping.DoesNotExist:
+                    continue  # No mapping exists
+
+            remove_groups = [
+                mapping.group for mapping in mappings if mapping.group not in new_groups
+            ]
+
+            add_groups = [group for group in new_groups if group not in current_groups]
```
jchristgit: Looking at the `.remove` call above, can't we have these as generators?

jchristgit: Django evaluates this as a list by itself:

```py
        def add(self, *objs, bulk=True):
            self._remove_prefetched_objects()
            objs = list(objs)
            # ...
```
via `django/db/models/fields/related_descriptors.py`

pydis_site/apps/home/signals.py
```diff
@@ -0,0 +1,159 @@
...
+
+            if remove_groups:
+                account.user.groups.remove(*remove_groups)
+
+            if add_groups:
+                account.user.groups.add(*add_groups)
```
jchristgit: Not fully convinced you need the `if` conditions here. I don't think it hurts to always run this and keep the code simpler

pydis_site/apps/home/tests/test_signal_listener.py
```diff
@@ -0,0 +1,319 @@
...
+
+            # Attempt to apply groups if everything checks out
+            user_logged_in.send(DjangoUser, user=self.django_user)
+            mock_obj.assert_called_with(self.discord_user, self.social_user)
+
+    def test_apply_groups_admin(self):
+        """Test application of groups by role, relating to an admin user."""
+        handler = SignalListener()
+
+        self.assertTrue(self.django_user_discordless.groups.all().count() == 0)
```
jchristgit: this should use `assertEqual(..., 0)`

pydis_site/apps/home/tests/test_signal_listener.py
```diff
@@ -0,0 +1,319 @@
...
+
+        # Apply groups based on admin role being present on Discord
+        handler._apply_groups(self.discord_admin, self.social_admin)
+        self.assertTrue(self.admin_group in self.django_admin.groups.all())
+
+        # Remove groups based on the user apparently leaving the server
+        handler._apply_groups(self.discord_admin, self.social_admin, True)
+        self.assertTrue(self.django_user_discordless.groups.all().count() == 0)
```
jchristgit: this should use `assertEqual(..., 0)`

pydis_site/apps/home/tests/test_signal_listener.py
```diff
@@ -0,0 +1,319 @@
...
+        handler._apply_groups(self.discord_unmapped, self.social_unmapped)
+        self.assertTrue(self.django_user_discordless.groups.all().count() == 0)
+
+        # No groups should be applied when there are only unmapped roles to match
+        handler._apply_groups(self.discord_unmapped, self.social_user)
+        self.assertTrue(self.django_user.groups.all().count() == 0)
+
+        # No groups should be applied when the user isn't in the guild
+        handler._apply_groups(self.discord_not_in_guild, self.social_user)
+        self.assertTrue(self.django_user.groups.all().count() == 0)
```
jchristgit: I'm not gonna comment on each, but you know the drill

pydis_site/apps/home/signals.py
lemonsaurus: There is a link to the Django docs, too. This is fine.

pydis_site/templates/base/navbar.html
```diff
@@ -83,6 +84,25 @@
...
+          {% if not user.is_authenticated %}
+            {% get_providers as socialaccount_providers %}
+
+            {% for provider in socialaccount_providers %}
+              {% if provider.id == "discord" %}
+                  <a class="navbar-item"
+                     href="{% provider_login_url provider.id process="login" scope=scope auth_params=auth_params %}"
+                  >Login with {{ provider.name }}</a>
+              {% endif %}
+            {% endfor %}
```
jchristgit, line 98: why not just display the single login url instead of iterating over each until you find discord?

jchristgit (changes requested):
This looks pretty solid so I will have to resort to notpicking :(

The role mapping model is just about the cleanest way I have seen to accomplish this. I think in my last implementation of this I just hardcoded that, but a nice admin frontend for it is pretty brilliant.

That said, if the role mappings update, shouldn't the signal handler fire?

pydis_site/apps/home/apps.py
jchristgit: Yeah, that makes sense.

pydis_site/templates/base/navbar.html
gdude2002: As far as I can tell, there are a couple reasons to do this:

* There might be more than one Discord OAuth app to login with
* Discord OAuth might not be configured yet

Those aren't necessarily real concerns for the time being, I guess. I do still need to get the URL from the provider though, and this is the official way Allauth provides to do it.

gdude2002:
Good catch on the missing signals. I'm leaving a short TODO list in this comment so I don't forget:

- [x] Handle RoleMapping deletions
- [x] Handle RoleMapping updates (use `pre_save` and check against object already in DB if possible, otherwise try to disallow updates)

pydis_site/apps/home/signals.py
gdude2002: This has proven to be not the case.

image

gdude2002:
Alright, I think that's all of your review addressed, @jchristgit!

jchristgit (approved):
very solid work

pydis_site/apps/home/signals.py
```diff
@@ -0,0 +1,192 @@
...
+    pre_social_login, social_account_added, social_account_removed,
+    social_account_updated)
+from django.contrib.auth.models import Group, User as DjangoUser
+from django.db.models.signals import post_save, pre_delete, pre_save
+
+from pydis_site.apps.api.models import User as DiscordUser
+from pydis_site.apps.staff.models import RoleMapping
+
+
+class SignalListener:
```
lemonsaurus: This class name is too generic. We might have other signal listener classes other places in the site, so it would probably be good if we make this class name more specific. something like `AllauthSignalListener` would be fine.

Besides, SignalListener sounds a bit like a Django internal, and it feels like it should take arguments that configure which signals to listen to, or something.

pydis_site/apps/home/signals.py
```diff
@@ -0,0 +1,192 @@
...
+
+        pre_social_login.connect(self.social_account_updated)
+        social_account_added.connect(self.social_account_updated)
+        social_account_updated.connect(self.social_account_updated)
+        social_account_removed.connect(self.social_account_removed)
+
+        user_logged_in.connect(self.user_logged_in)
+
+    def user_logged_in(self, sender: Type[DjangoUser], **kwargs) -> None:
+        """Handles signals relating to Allauth logins."""
```
lemonsaurus: I think you can probably improve these docstrings -- that goes for pretty much all the docstrings in this file. I understand that it handles these signals, but what does it actually _do_ when the user logs in?

pydis_site/apps/home/signals.py
```diff
@@ -0,0 +1,192 @@
...
+        try:
+            account: SocialAccount = SocialAccount.objects.get(
+                uid=str(instance.id), provider=DiscordProvider.id
+            )
+        except SocialAccount.DoesNotExist:
+            return  # User has never logged in with Discord on the site
+
+        self._apply_groups(instance, account)
+
+    def _apply_groups(
```
lemonsaurus, line 208: Don't we lint for docstrings on functions like this?

Either way I think there should be a docstring.

pydis_site/apps/home/tests/test_signal_listener.py
```diff
@@ -0,0 +1,392 @@
...
+
+from pydis_site.apps.api.models import Role, User as DiscordUser
+from pydis_site.apps.home.signals import SignalListener
+from pydis_site.apps.staff.models import RoleMapping
+
+
+class SignalListenerTests(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        """Executed when testing begins."""
```
lemonsaurus: Again, this docstring could stand to be improved, especially considering the complexity of this setup method. 

What are we setting up? Why?

pydis_site/apps/home/tests/test_signal_listener.py
```diff
@@ -0,0 +1,392 @@
...
+
+        with mock.patch.object(SignalListener, "_apply_groups", mock_obj):
+            _ = SignalListener()
+
+            post_save.send(
+                DiscordUser,
+                instance=self.discord_user,
+                raw=True,
+                created=None,  # Not realistic, but we don't use it
+                using=None,  # Again, we don't use it
```
lemonsaurus: can we line these inlines up? It'll probably look better, and PEP8 allows it.

pydis_site/apps/home/tests/test_signal_listener.py
```diff
@@ -0,0 +1,392 @@
...
+            )
+
+            mock_obj.assert_not_called()
+
+            post_save.send(
+                DiscordUser,
+                instance=self.discord_user,
+                raw=False,
+                created=None,  # Not realistic, but we don't use it
+                using=None,  # Again, we don't use it
```
lemonsaurus: same thing here

pydis_site/apps/home/tests/test_signal_listener.py
```diff
@@ -0,0 +1,392 @@
...
+            provider=DiscordProvider.id,
+            uid=cls.discord_moderator.id
+        )
+
+    def test_model_save(self):
+        """Test signal handling for when Discord user model objects are saved to DB."""
+        mock_obj = mock.Mock()
+
+        with mock.patch.object(SignalListener, "_apply_groups", mock_obj):
+            _ = SignalListener()
```
lemonsaurus: Would it work to just not store this at all? like just doing `SignalListener()`? I'm mostly curious, the way you've done it here doesn't really bother me, but it might be cleaner to just call it like a function if it actually works.

pydis_site/apps/home/tests/test_views.py
```diff
@@ -7,3 +7,17 @@ def test_index_returns_200(self):
         url = reverse('home')
         resp = self.client.get(url)
         self.assertEqual(resp.status_code, 200)
+
+
+class TestLoginCancelledReturns302(TestCase):
+    def test_login_cancelled_returns_302(self):
```
lemonsaurus, line 14: can we docstring these? It's nice because when the tests are run, it will display the docstrings instead of the method names as it lists them, and that looks nice. I know the function name is self-documenting, but I'd rather read a natural language sentence that describes the same thing.

No need for a long docstring here, just a one-liner is fine.

same with the method below.

pydis_site/apps/home/urls.py
```diff
@@ -1,14 +1,36 @@
...
+    path('accounts/', include('allauth.socialaccount.providers.discord.urls')),
+
+    path(
+        'accounts/login/cancelled', MessageRedirectView.as_view(
+            pattern_name="home", message="Login cancelled."
+        ), name='socialaccount_login_cancelled'
+    ),
+    path(
+        'accounts/login/error', MessageRedirectView.as_view(
+            pattern_name="home", message="Login failed due to an error, please try again.",
```
lemonsaurus: "Failed due to an error" is a bit of a pleonasm. Maybe "The Login process encountered an error" or something?

pydis_site/settings.py
```diff
@@ -278,9 +294,18 @@
...
 # Custom settings for django-simple-bulma
 BULMA_SETTINGS = {
-    "variables": {
-        "green": "#21c65c",  # Accessibility: Better contrast with the light text
-        "primary": "#7289DA",
+    "variables": {  # If you update these colours, please update the notification.css file
+        "primary": "#7289DA",  # Discord blurple
+
+        "orange": "#ffb39b",  # Bulma default, but at a saturation of 100
+        "yellow": "#ffea9b",  # Bulma default, but at a saturation of 100
```
lemonsaurus: can we line these up?

lemonsaurus (dismissed):
Very high quality PR, but I do have a few requests before I can approve.

pydis_site/apps/home/signals.py
gdude2002: We don't lint for docstrings on private methods - just public ones - but I'm happy to add one.

pydis_site/apps/home/tests/test_signal_listener.py
gdude2002: Do you mean, like, so they're not all on separate lines?

pydis_site/settings.py
gdude2002: Not sure what you mean by lining up here either

pydis_site/apps/home/tests/test_signal_listener.py
lemonsaurus: no no, just the inline comments.

```py
                created=None,  # Not realistic, but we don't use it
                using=None,    # Again, we don't use it
```

pydis_site/settings.py
lemonsaurus:
```py
thing()  # this
with.fire()  # looks
pull_chain()  # messy

thing()       # this
with.fire()   # looks
pull_chain()  # clean
```

pydis_site/apps/home/tests/test_signal_listener.py
gdude2002: I kind of just pre-emptively did this to make sure garbage collection didn't screw us over, but I can check.

gdude2002: Works fine without that, yep.

scragly (approved):
Absolutely sexy. LGTM

PR #275, opened 2019-10-03
Fix migrations exclude in flake8

MarkKoz:
They were not being excluded when flake8 was invoked via pre-commit.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,7 +3,7 @@ max-line-length=100
 docstring-convention=all
 import-order-style=pycharm
 application_import_names=pydis_site
-exclude=__pycache__, venv, .venv, **/migrations
+exclude=__pycache__, venv, .venv, **/migrations/**
 ignore=
     B311,W503,E226,S311,T000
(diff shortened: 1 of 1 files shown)


sco1 (approved):

SebastiaanZ (approved):
~~The fix works as expected, but I'm not entirely sure if that tells us that it's working. The problem is that I can't replicate the difference in behavior, as `precommit` excludes the migration files with both versions of `.flake8` in place on my machine. That makes me think that the behavior may be platform-dependent or dependent on "something" at least.~~

~~That said, if both versions work for some and the new version seems to work for everyone, the new version is definitely an improvement.~~

I've been able to replicate this now, but I don't see why it didn't previously. I did the same both times (trying to commit a change in a migration file that contained linting errors if not excluded) and this time it yelled at me without your fix.

SebastiaanZ:
I did some more testing and it consistently works with @MarkKoz's suggested change. I'll merge it now.

PR #277, opened 2019-10-07
Add jump_url to the reminder model for bot#466

Akarys42:
Hi, this PR add a jump_url field to the reminder model. 

This PR should be merged at the same time than python-discord/bot#466

### DIFF
--- a/pydis_site/apps/api/models/bot/reminder.py
+++ b/pydis_site/apps/api/models/bot/reminder.py
@@ -35,7 +35,7 @@ class Reminder(ModelReprMixin, models.Model):
     jump_url = models.CharFiels(
         help_text=("The jump url of the message "
                    "that created the reminder"
-                  )
+        )
     )
     content = models.CharField(
         max_length=1500,
--- a/pydis_site/apps/api/models/bot/reminder.py
+++ b/pydis_site/apps/api/models/bot/reminder.py
@@ -32,6 +32,11 @@ class Reminder(ModelReprMixin, models.Model):
             ),
         )
     )
+    jump_url = models.CharFiels(
+        help_text=("The jump url of the message "
+                   "that created the reminder"
+                  )
+    )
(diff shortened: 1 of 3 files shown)


SebastiaanZ:
As far as I can tell, your pull request does not include the migration files needed to migrate the database according to the new model specifications. You also need to think about what the database should do with existing entries: If we create a new `jump_url` column, which value should it store for existing entries that were added before this column existed?

In addition, the [serializer for the reminder model](https://github.com/python-discord/site/blob/d87226479a53708d6fdb5d39a8e87b12b10e065e/pydis_site/apps/api/serializers.py#L184-L193) that takes care of serializing and deserializing the model from and to json does not know your field yet. Depending on your solution for the values for existing rows, we may also want some kind of validation in the serializer to ensure new entries do include that row. That validator will also needs tests to make sure it's working correctly.

Akarys42:
Okay so the serializer should now know about the new field. I’ve done some research, but how I should do for the migration part?

(Oh and I will squash down some commit to make the commit history more proper when I will have finished)

SebastiaanZ:
If you want to contribute to the `site` repository, then it's a good idea to get familiar with Django and, if you working on something API-related, [Django REST framework](https://www.django-rest-framework.org/).  The official Django documentation has [a great tutorial](https://docs.djangoproject.com/en/2.2/intro/) that covers the basics of working with Django, including models and the related migrations that need to take place to get the database up-to-date with the latest specification of the models.

I also want to ask you to test and lint your code prior to opening pull requests. While we always have to additional layers of protection in the form of requiring two reviews and a CI pipeline, the proper functioning of our API is essential to the basic utilities we use to run the community, like our moderation tools. That means it's also essential that code is properly tested before it's submitted as a pull request.

I'm going to close this pull request for now, since it's not up to the standards we require for pull requests to this repository.

PR #278, opened 2019-10-07
Validation for active infractions

SebastiaanZ:
This pull request adds validation rules to the API to validate infractions based on the `active` field. This means that the API will now reject active infractions of types that should never be active (notes, warnings, and kicks) and only accept one active infraction per user for the other types of infractions (mute, ban, watch, superstar). 

In addition, this pull requests adds a data migration to make sure the database itself is consistent with the new validation rules. Since this is a fairly complex data migration, I have added tests to make sure the migration works as expected. (Read more about that below.)

Finally, I've introduced a database-level `UniqueConstraint` to prevent a user from having multiple active infractions of the same type. I have also added tests to make sure this constraint works as expected.

## Testing migrations

To test the effect of the migration file, I've created a subclass of `django.test.TestCase`, `MigrationsTestCase`, that reverts the test database to the migration just prior to the one of interest. We can then inject test data, apply the migration, and test if the migration had the desired effect. 

Since a migration is really snapshot in time of the migration history, this also allows us to work with the database models as they were at those two time points in the migration history (directly before and directly after applying this migration), since otherwise these tests will fail in the future if we decide to change the models.


pydis_site/apps/api/serializers.py
```diff
@@ -1,4 +1,5 @@
 """Converters from Django models to data interchange formats and back."""
+import logging
```
SebastiaanZ: I've accidentally left this logging import here; I'll remove it when pushing the commit that straightens out the migration conflict that #269 will cause.

pydis_site/apps/api/serializers.py
```diff
@@ -12,6 +13,8 @@
     Tag, User
 )
 
+log = logging.getLogger(__name__)
```
SebastiaanZ: I'll remove this then as well, obviously.

pydis_site/apps/api/models/bot/user.py
```diff
@@ -50,7 +50,7 @@ class User(ModelReprMixin, models.Model):
 
     def __str__(self):
         """Returns the name and discriminator for the current user, for display purposes."""
-        return f"{self.name}#{self.discriminator}"
+        return f"{self.name}#{self.discriminator:0>4}"
```
MarkKoz, line 53: What's this for?

Edit: it's for when discriminators are less than 4 digits, right? Cause they are stored as integers and are therefore not zero-padded by default?

pydis_site/apps/api/tests/test_infractions.py
```diff
@@ -305,6 +312,167 @@ def test_returns_400_for_non_hidden_required_hidden_type(self):
...
+        for infraction_type in restricted_types:
+            with self.subTest(infraction_type=infraction_type):
+                invalid_infraction = {
+                    'user': self.user.id,
+                    'actor': self.user.id,
+                    'type': infraction_type,
+                    'reason': 'Take me on!',
+                    'hidden': True,
+                    'active': True,
+                    'expires_at': '2019-10-04T12:52:00+00:00'
```
MarkKoz: This test is a bit fragile in the sense that it relies on the active validation logic preceding the validation for allowing expiration or being hidden. But I think I can live with that.

pydis_site/apps/api/serializers.py
```diff
@@ -110,6 +113,16 @@ def validate(self, attrs: dict) -> dict:
         """Validate data constraints for the given data and abort if it is invalid."""
         infr_type = attrs.get('type')
 
+        active = attrs.get('active')
+        if active and infr_type in ('note', 'warning', 'kick'):
+            raise ValidationError({'active': [f'{infr_type} infractions cannot be active.']})
+
+        user = attrs.get('user')
+        if active and Infraction.objects.filter(user=user, type=infr_type, active=True).exists():
```
MarkKoz: How come a [`UniqueValidator`](https://www.django-rest-framework.org/api-guide/validators/#uniquevalidator) wasn't used instead? Also, supposedly a `ModelSerializer` does this: 

> It will automatically generate validators for the serializer, such as unique_together validators.

I wonder, does it support generating validators for unique constraints on the model?

pydis_site/apps/api/models/bot/user.py
SebastiaanZ: That's right, yeah.

pydis_site/apps/api/tests/test_infractions.py
```diff
@@ -305,6 +312,167 @@ def test_returns_400_for_non_hidden_required_hidden_type(self):
...
+        with self.assertRaises(IntegrityError):
+            Infraction.objects.create(
+                user=self.user,
+                actor=self.user,
+                type="ban",
+                active=True,
+                reason="The second active ban"
+            )
+
+    def test_unique_constraint_accepts_active_infraction_after_inactive_infraction(self):
```
MarkKoz, line 429: Since this and the following test cases have no assertions, it means you're relying on `IntegrityError` not being raised? If it was raised (like any other exception), it'd fail the test, right? Sorry if that has an obvious answer, just making sure here.

pydis_site/apps/api/tests/test_infractions.py
```diff
@@ -305,6 +312,167 @@ def test_returns_400_for_non_hidden_required_hidden_type(self):
...
+                    'actor': self.user.id,
+                    'type': infraction_type,
+                    'reason': 'Take me on!',
+                    'hidden': True,
+                    'active': True,
+                    'expires_at': '2019-10-04T12:52:00+00:00'
+                }
+                response = self.client.post(url, data=invalid_infraction)
+                self.assertEqual(response.status_code, 400)
+                self.assertEqual(response.json(), {
```
MarkKoz: I don't believe this is consistent with our code style but honestly it currently looks better than what it's "supposed" to be:

```py
self.assertEqual(
    response.json(), 
    {
        'active': [f'{infraction_type} infractions cannot be active.']
    }
)
```

It wouldn't be as bad if the dict was on one line:

```py
self.assertEqual(
    response.json(), 
    {'active': [f'{infraction_type} infractions cannot be active.']}
)
```

MarkKoz: Oh, so I'm not sure if this should be changed or not. Thoughts?

pydis_site/apps/api/tests/test_infractions.py
```diff
@@ -305,6 +312,167 @@ def test_returns_400_for_non_hidden_required_hidden_type(self):
...
+                    'actor': self.user.id,
+                    'type': infraction_type,
+                    'reason': 'Take on me!',
+                    'active': True,
+                    'expires_at': '2019-10-04T12:52:00+00:00'
+                }
+                second_response = self.client.post(url, data=second_active_infraction)
+                self.assertEqual(second_response.status_code, 400)
+                self.assertEqual(second_response.json(), {
+                    'active': [f'This user already has an active {infraction_type} infraction']
```
MarkKoz: Same as [line 333](https://github.com/python-discord/site/pull/278/files#r333130421).

pydis_site/apps/api/tests/test_infractions.py
SebastiaanZ: I will change this to a more explicit variant. I've looked around for how to properly test a `assertNotRaises` situation, since there's no built-in way to do that with `unittest`, but I'm finding conflicting answers.

Some advise to just run it, like I've done here, and rely on the test resulting in `Error` if an exception occurs, but other argue, and I think I agree with that, that observing an exception where non should occur should result in a test `Failure` instead. 

One solution would be to do something like:

```py
try:
    # the model creation
except IntegrityError:
    self.fail("Raised IntegrityError unexpectedly!")
```

pydis_site/apps/api/migrations/0044_active_infractions_migration.py
```diff
@@ -0,0 +1,106 @@
...
+        for user in query:
+            infractions = sorted(user.infractions, key=ExpirationWrapper, reverse=True)
+            for infraction in infractions[1:]:
+                infraction.active = False
+                infraction.save()
+
+
+def reverse_migration(apps, schema_editor):
+    """There's no need to do anything special to reverse these migrations."""
+    return
```
MarkKoz: Shouldn't it still do something? Can Django already figure out how to, for example, change a field back to True from False on its own?

pydis_site/apps/api/tests/test_infractions.py
SebastiaanZ: I'll change that.

pydis_site/apps/api/migrations/0044_active_infractions_migration.py
SebastiaanZ: The reason we don't need to do anything is because we can't reverse the state of the actual data; we simply don't know what the previous value were. Reversal is necessary if the forward function of `RunPython` also changed the model itself, since you then have to restore the model to the state prior to the migration.

So, we can't undo the data migration and we don't have to undo structural changes, since we made none.

pydis_site/apps/api/migrations/0044_active_infractions_migration.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # Build an SQL query by chaining methods together
+
+    # Get users with active infraction(s) of the provided `infr_type`
+    query = user_model.objects.filter(
+        Q(infractions_received__type=infr_type, infractions_received__active=True)
+    )
+
+    # Prefetch their active received infractions of `infr_type` and attach `.infractions` attribute
+    query = query.prefetch_related(
```
MarkKoz: Minor efficiency thing: I think the prefetch could be done after the gte 2 filter and the distinct()

MarkKoz: Well it's just a migration I guess this sort of stuff doesn't even matter

SebastiaanZ: It's a simple change, but I wonder if it will make a difference in the SQL query that's going to be made. The actual query will not be run until after all the methods are chained together and we're actually trying to read from `query` in the original function.

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
```diff
@@ -0,0 +1,545 @@
...
+        Creates `infractions` for the `user` with the given `actor`.
+
+        The `infractions` dictionary can contain the following fields:
+         - `type` (required)
+         - `active` (default: True)
+         - `expires_at` (default: None; i.e, permanent)
+         - `hidden` (default: False).
+
+         The parameters `infraction_model` and `user_model` can be used to pass in an instance of
+         both model classes from a different migration/project state.
```
MarkKoz: These two lines seem to be indented a bit too far.

MarkKoz (changes requested):

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
```diff
@@ -0,0 +1,545 @@
...
+            ),
+            infraction_model=self.infraction_model,
+            user_model=self.user_model,
+        )
+        database_count = Infraction.objects.all().count()
+        self.assertEqual(3, database_count)
+
+    def test_infraction_factory_multiple_users(self):
+        """Does the test database hold as many infractions as we tried to create?"""
+        created_infractions = {}
```
MarkKoz: This dictionary doesn't seem to be used anywhere else.

MarkKoz (changes requested):

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
```diff
@@ -0,0 +1,545 @@
...
+                active=infraction['active'],
+                hidden=infraction['hidden'],
+                expires_at=infraction['expires_at'],
+            )
+            infraction_history.append(new_infraction)
+
+        return InfractionHistory(user_id=user_id, infraction_history=infraction_history)
+
+
+class InfractionFactoryTests(MigrationsTestCase):
```
MarkKoz, line 80: Why is this a `MigrationsTestCase`? The tests for the factory aren't concerned with what values are allowed for fields but rather just that if a given value is equal to the resulting value. Is it just so future migrations don't break this test (e.g. in the future a check constraint for a kick would be added that it must always be hidden, so line 105 would fail)?

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
```diff
@@ -0,0 +1,545 @@
...
+        # Moderator that serves as actor for all infractions
+        cls.user_moderator = cls.user_model.objects.create(
+            id=9999,
+            name="Olivier de Vienne",
+            discriminator=1040,
+            avatar_hash=None,
+        )
+
+        # User #1: clean user with no infractions
+        cls.created_infractions.update(
```
MarkKoz: Why use `.update()` over `cls.created_infractions[1] = InfractionFactory.create(...)`?

MarkKoz (changes requested):

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
```diff
@@ -0,0 +1,545 @@
...
+            {
+                1: InfractionFactory.create(
```
MarkKoz: Do you think using strings as keys instead of IDs would improve readability, or is it already pretty clear from the test case names which infractions are being used?

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
```diff
@@ -0,0 +1,545 @@
...
+        self.assertFalse(
+            self.infraction_model.objects.filter(user__id=user_id).exists()
+        )
+
+    def test_migration_left_user_with_inactive_note_untouched(self):
+        """Did the migration leave users with only an inactive note untouched?"""
+        user_id, infraction_history = self.created_infractions[2]
+        inactive_note = infraction_history[0]
+        self.assertTrue(
+            self.infraction_model.objects.filter(**model_to_dict(inactive_note)).exists()
```
MarkKoz, line 416: Do all these filter by infraction ID too? Or is that not necessary?

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
SebastiaanZ: > Is it just so future migrations don't break this test (e.g. in the future a check constraint for a kick would be added that it must always be hidden, so line 105 would fail)?

Yes, it's to pin these tests to the point in the migration history we are testing. For our non-migration tests, we always update the tests to reflect the current (i.e., latest) states of the models, but that would not make sense for something that is used specifically to generate data according to an older model.

In this case, I want to use `InfractionFactory` to specifically generate data that would violate the UniqueTogether constraint introduced in a later migration.

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
SebastiaanZ: Good point. Adaptation from an earlier version and it was late.

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
SebastiaanZ: I think strings like `"user 1"` would be clearer, so I'm going to change that.

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
SebastiaanZ: Yes, they do. The infraction object returned by the InfractionFactory will include the `id` (and all the other fields of the created `Infraction` object).

MarkKoz (commented):
* Migration testing is quite clever. I'm impressed. Nice work!
* The infractions tests don't strike me as very DRY, though many other tests for the site are probably also guilty of that. Would it be feasible to come up with some functions to make it more DRY, at least for the new infraction tests you've written?
* While your tests are already thorough, I noticed that `MigrationsTestCase` does not itself have any tests. Would it be feasible to come up with tests for that? Is it even worth it?

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
MarkKoz: I was thinking something along the lines of `"no_infractions"`, `"one_active_note"`, etc.

MarkKoz: I disliked scrolling back and forth in the module to see what the infractions are supposed to be but I understand the objects need to be created in the hook rather than in the test cases (which would be more ideal for readability).

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
MarkKoz: > In this case, I want to use InfractionFactory to specifically generate data that would violate the UniqueTogether constraint introduced in a later migration.

That makes perfect sense, I did not catch this use myself.

pydis_site/apps/api/migrations/0044_active_infractions_migration.py
MarkKoz: > we simply don't know what the previous value were.

Oh, right... that seems quite obvious now 😄

pydis_site/apps/api/migrations/0044_active_infractions_migration.py
```diff
@@ -0,0 +1,106 @@
...
+            return False
+        else:
+            return self.expiration_date == other.expiration_date
+
+
+def migrate_inactive_types_to_inactive(apps, schema_editor):
+    """Migrates infractions of non-active types to inactive."""
+    inactive_infraction_types = Q(type="note") | Q(type="warning") | Q(type="kick")
+    infraction_model = apps.get_model('api', 'Infraction')
+    infraction_model.objects.filter(inactive_infraction_types).update(active=False)
```
jchristgit: can't we use `type__in` here instead?

```py
.filter(type__in=('note', 'warning', 'kick'))
```

pydis_site/apps/api/migrations/0044_active_infractions_migration.py
```diff
@@ -0,0 +1,106 @@
...
+    The QuerySet will prefetch the infractions and attach them as an `.infractions` attribute to the
+    `User` instances.
+    """
+    active_infractions = infraction_model.objects.filter(type=infr_type, active=True)
+
+    # Build an SQL query by chaining methods together
+
+    # Get users with active infraction(s) of the provided `infr_type`
+    query = user_model.objects.filter(
+        Q(infractions_received__type=infr_type, infractions_received__active=True)
```
jchristgit: I am probably missing something, but what's the difference between `Q` and no `Q`?

pydis_site/apps/api/tests/migrations/base.py
```diff
@@ -0,0 +1,102 @@
...
+        are applied. This allows us to simulate a database state just prior to the migrations we are
+        trying to test.
+
+        To make sure we're creating objects according to the state the models were in at this point
+        in the migration history, use `apps.get_model(app_name: str, model_name: str)` to get the
+        appropriate model, e.g.:
+
+        >>> Infraction = apps.get_model('api', 'Infraction')
+        """
+        pass
```
jchristgit, line 90: isn't a docstring without method body valid on its own without `pass`?

pydis_site/apps/api/migrations/0044_active_infractions_migration.py
SebastiaanZ: Nothing. I changed my approach, copy-pasted it, and did not think about it properly. I'm working on an update of this PR, so I'll this here as well.

pydis_site/apps/api/migrations/0044_active_infractions_migration.py
SebastiaanZ: Yes, we can. I'll change that as well.

pydis_site/apps/api/tests/migrations/base.py
MarkKoz: I think so, but I don't mind the `pass` being there to more explicitly convey that it does nothing.

jchristgit: fair

pydis_site/apps/api/serializers.py
SebastiaanZ: I've changed this logic for the `UniqueTogetherValidator` with a custom message. It works well, although it gives slightly less specific feedback:

```py
'non_field_errors': ['This user already has an active infraction of this type.']
```

However, since this is generated for a specific infraction payload, that should not really be an issue and the message is unambiguous in the context of the query.

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
SebastiaanZ: I've removed it in my latest commit (not pushed atm, but soon)

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
SebastiaanZ: Yes, that can't really be helped unless we want to rollback, add data, apply the migration in each test case. That would increase the running time of the tests by a lot, though. This test is already on the heavy side (although we're talking one to two seconds in the pipeline, which is acceptable in my opinion).

I'll use sensible strings as keys to make it clearer, I think that's the best we can do here.

SebastiaanZ:
> * The infractions tests don't strike me as very DRY, though many other tests for the site are probably also guilty of that. Would it be feasible to come up with some functions to make it more DRY, at least for the new infraction tests you've written?

(I still have to look into this.)

> * While your tests are already thorough, I noticed that `MigrationsTestCase` does not itself have any tests. Would it be feasible to come up with tests for that? Is it even worth it?

I'm not quite sure how to do that. I've thought about it, but I haven't come up with a way of testing it that isn't just "pretending to thoroughly test it". That is, testing some parts of it is fairly straight forward (the initial checks for set class variables, whether or not the hook methods are getting called) and I can add that, but that's not really testing the core of the `TestCase` yet.

The idea I'm playing with in my head is patching the hook methods to see whether they get called with the apps in the state we expect them to be in at that point in the migration history. That would confirm the thing does what it should: Give us states at specified points in the migration history. It's a bit of a hack, though, cause I'd need to find a migration point for which I can test "We've gone to precisely this state". Not sure if there's a better way.

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
SebastiaanZ: I forgot about this. I still need to fix this.

pydis_site/apps/api/tests/test_infractions.py
SebastiaanZ: I can make the payload more dependent on the type, but it would mean a logically more complex test.

pydis_site/apps/api/tests/test_infractions.py
SebastiaanZ: I've changed it to the proposed solution.

pydis_site/apps/api/tests/migrations/test_active_infraction_migration.py
SebastiaanZ: I've changed this.

pydis_site/apps/api/serializers.py
MarkKoz: So you still ended up having to manually specify the validator. It did not generate it on its own?

pydis_site/apps/api/tests/test_infractions.py
```diff
@@ -417,27 +424,37 @@ def test_unique_constraint_raises_integrity_error_on_second_active_of_same_type(
...
+                actor=self.user,
+                type="ban",
+                active=True,
+                reason="The first active ban"
+            )
+        except IntegrityError:
+            self.fail("An unexpected IntegrityError was raised.")
+
+    @patch(f"{__name__}.Infraction")
+    def test_the_accepts_active_infraction_after_inactive_infractions_test(self, infraction_patch):
```
MarkKoz: This test name is confusing, perhaps unsurprisingly.

MarkKoz: I don't really feel like this test is necessary to be honest.

pydis_site/apps/api/tests/migrations/test_base.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def test_project_state_is_loaded_for_correct_migration_files_twice(self, migration_executor):
+        """The `project_state` should first be loaded with `migrate_from`, then `migrate_to`."""
+        self.test_case.setUpTestData()
+
+        expected_args = [call(self.test_case.migrate_from), call(self.test_case.migrate_to)]
+        self.assertEqual(migration_executor().loader.project_state.call_args_list, expected_args)
+
+    def test_loader_build_graph_gets_called_once(self, migration_executor):
+        """We should rebuild the before applying the second set of migrations."""
```
MarkKoz: This docstring seems to be missing a word.

MarkKoz (changes requested):

pydis_site/apps/api/tests/test_infractions.py
SebastiaanZ: It's necessary since we include tests in the coverage report and otherwise the `except`-block for the test we're testing here will never be run. 

We may be able to declare an exception to the coverage requirement for those specific lines, but I wrote to test to make sure it actually works like I intended it to, since it does have custom test logic.

pydis_site/apps/api/tests/migrations/test_base.py
SebastiaanZ: Ah, yes. I'll fix that.

pydis_site/apps/api/tests/test_infractions.py
MarkKoz: Oh right. I am not used to coverage being enabled for tests. I still think the name could be better by mentioning that it is specifically testing the exception being caught, but the name is already getting quite long 🤔 .

pydis_site/apps/api/serializers.py
SebastiaanZ: As far as I can tell, it does not automatically add a validator for the constraint we've added to the model. If I remove our "custom" validator, the serializer doesn't seem to catch double active infractions and we'll get an `IntegrityError` (fail late in the database) instead.

pydis_site/apps/api/tests/test_infractions.py
MarkKoz: Maybe not that much more complex. I'm wondering how it'd look if the expiration and hidden args could be put in the tuple alongside the types.

pydis_site/apps/api/serializers.py
MarkKoz: That's fine then. I looked at the [source](https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L1535) and it seems to only work if [`unique_together`](https://docs.djangoproject.com/en/2.2/ref/models/options/#unique-together) is used.

pydis_site/apps/api/tests/test_infractions.py
SebastiaanZ: Better, I think.

MarkKoz (approved):

SebastiaanZ:
Since other pull requests with migrations for the api app have since been merged, care should be taken when merging this. I'll resolve the created conflicts later, since it involves a bit more for this PR. In the mean time, reviews are still welcome.

lemonsaurus (approved):

PR #279, opened 2019-10-08
Add favicons; workaround for gitattributes

ByteCommander:


sco1 (approved):

jchristgit:
Most of these have colours but the Safari one is just.. weird. is that intentional, or is that some Apple thing?

Kitty can't open it at all:

image

In Safari it's black and white

image

lemonsaurus:
I think it's some kind of Apple thing, since these were generated by a website that I've used for a bunch of sites with no issue. it probably works.

lemonsaurus (approved):

ByteCommander:
@jchristgit This Safari Pinned Tab icon must be a monochrome and transparent SVG, according to https://realfavicongenerator.net/. It offers several ways to convert the uploaded picture to a fitting format, including to upload a separate transparent PNG and convert it to a solid "shadow", which is what I did. 
Probably that caused these small dot artifacts around the shape where we'd have a drop shadow in the original. Maybe there could have been a better approach by preparing a different icon from scratch or from the original SVG, but it seemed good enough this way, especially as it's going to be displayed only as small tab icon.
Btw not sure what kitty is, but eog opens it without any complaints.

PR #281, opened 2019-10-11
Allow viewing log entries in the Django Admin.

jchristgit:
This PR adds log entries to the Django Admin (under the API section) to allow viewing them there.

The default overview displays log entries in chronologically descending order...

image

... and individual log entries can be inspected by clicking on them, with editing disabled to reduce clutter ...

image

... multiline tracebacks are displayed properly:

image

As I don't want to put tons of business logic in the admin this doesn't customize the HTML / CSS of it manually or anything like that.


pydis_site/apps/api/models/log_entry.py
```diff
@@ -1,3 +1,5 @@
+import textwrap
```
MarkKoz: This is not being used.

MarkKoz (changes requested):

pydis_site/apps/api/models/log_entry.py
jchristgit: woops, forgot to run flake8 after moving to the django builtin stuff

MarkKoz (commented):
I think showing both the module and the logger name is redundant in most cases though technically they don't have to be the same name. I suppose it's fine to include it. In any case, the description for the `module` field is incorrect since clearly that is not a fully qualified name.

MarkKoz:
Does the list support filtering?

jchristgit:
> In any case, the description for the module field is incorrect since clearly that is not a fully qualified name.

Yeah, my original assumption was that it was, but for some reason getting the full module out of logging seems to be pretty hard. I can correct that in a separate PR.

jchristgit:
> Does the list support filtering?

Now it does. I (well, Django) added filtering by app / timestamp / level and searching by message in f158422.

image

MarkKoz (commented):
Just tried this out. Filtering works well.

My concern now: can the ability to delete log entries be removed from the django admin? I feel like we would never want to do that.

MarkKoz:
Also, a nice feature to have is the ability to link to source code on our repo based on the line number and module name (and/or link in the tracebacks). Perhaps that can be saved for another PR though.

lemonsaurus:
I think we will probably eventually write a real frontend for this instead of using the Django admin, and that would be a good place to put nice to have features like advanced filtering, links to relevant source code, maybe even for restricting deletion.

jchristgit:
> My concern now: can the ability to delete log entries be removed from the django admin? I feel like we would never want to do that.

Done.

MarkKoz (changes requested):
The merge introduced a conflicting migration that needs to be fixed:

```
django.core.management.base.CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0044_migrate_nominations_from_infraction_to_nomination_model, 0044_add_plural_name_for_log_entry in api).
```

jchristgit:
fixed in 673cc3b

lemonsaurus (approved):

MarkKoz (changes requested):
Sorry. Now I noticed the "add log entry" button which I think should also be removed. I could have sworn that button wasn't there before!

jchristgit:
I don't mind it at all, no worries

Removed in 9424ae0

docker-compose.yml
```diff
@@ -11,7 +11,7 @@
 version: "3.6"
 services:
   postgres:
-    image: postgres:11-alpine
+    image: postgres:12-alpine
```
MarkKoz: Surely this is a mistake?

MarkKoz (changes requested):

docker-compose.yml
jchristgit: I was already wondering where that change went...

Amended to the latest commit

MarkKoz (approved):
Splendid

PR #288, opened 2019-10-19
Add a jump_url field to the reminder model

Akarys42:
Hi, python-discord/bot#466 requires to add a `jump_url` field to the `reminder` model, so here it is!


pydis_site/apps/api/migrations/0046_reminder_jump_url.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('api', '0045_add_plural_name_for_log_entry'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='reminder',
+            name='jump_url',
+            field=models.CharField(default='invalid', help_text='The jump url to the message that created the reminder', max_length=88),
```
SebastiaanZ: Why don't we use an empty string as the value for existing entries? I think that would make the logic at the side of the bot more beautiful:

```py
if jump_url:
    # We know we have a jump_url
```

instead of

```py
if jump_url != "invalid":
    # Do something
```

```suggestion
            field=models.CharField(default='', help_text='The jump url to the message that created the reminder', max_length=88),
```

pydis_site/apps/api/models/bot/reminder.py
```diff
@@ -15,6 +15,12 @@ class Reminder(ModelReprMixin, models.Model):
             "If not, it has been sent out to the user."
         )
     )
+    jump_url = models.CharField(
```
SebastiaanZ: Is there a specific reason for not using an [URLField](https://docs.djangoproject.com/en/2.2/ref/models/fields/#urlfield) here? Using an `URLField` will mean that URLs in requests will automatically validated by the [URLValidator](https://docs.djangoproject.com/en/2.2/ref/validators/#django.core.validators.URLValidator). Seems like a useful addition to make sure the data we receive is valid.

pydis_site/apps/api/models/bot/reminder.py
```diff
@@ -15,6 +15,12 @@ class Reminder(ModelReprMixin, models.Model):
...
+        max_length=88,
```
SebastiaanZ, line 19: What's your reason for `88` here? Is it based on existing URLs Discord generates? If so, this means that this may break in the future: snowflake IDs are integer representations of 64-bit numbers. In base10 form, they're currently one digit shorter than they could theoretically become. I'd rather have us have some margin here. (Although I haven't computed if they will get larger realistically given the time frame we live in.)

SebastiaanZ (changes requested):
Some small changes

pydis_site/apps/api/models/bot/reminder.py
Akarys42: From my calculations the first bit will be 1 when we will reach the year 71745, so I think we are pretty safe :) but do you still want to me to add a little margin of something like 2 characters?

SebastiaanZ: Yeah, that probably doesn't really matter then. (Although I think it's the year 54525, since it's not the last bit, but another digit in base10, but same thing.)

Akarys42:
Updated.

MarkKoz (approved):

SebastiaanZ (approved):

PR #294, opened 2019-10-20
Remove breadcrumb truncation

bsoyka:
As per discussion in #meta in regards to the Code of Conduct page.


gdude2002:
This looks good to me, I'm about 90% sure this'll work, but it needs to be tested. If you're able to test it yourself, would you mind dropping a screenshot? Otherwise, I'll try to test when I have time.

bsoyka:
@gdude2002 Tried to test it, but I can't get Docker to work right on my machine. Sorry! 😬

lemonsaurus (approved):
Well, the approach looks sound but we do need it tested before we can merge this. 

I also ask myself how this will affect pages with unreasonably long names. Will it take up the entire page or do weird stuff like give us horizontal scrollbars? 

I also wonder whether a better solution might've been smarter truncation instead of no truncation at all. For example, are breadcrumbs really worth having if there's only one level? Would the truncation perhaps look better if it didn't cut off in the middle of a word?

But anyway, this is probably an improvement for now at any rate.

gdude2002:
@bsoyka DM'd me last night to say they were having trouble setting up Docker and the site, so I'm testing it now.

It might be worth noting that the wiki is also forcibly limiting the slug for wiki pages to 50 chars - although that's not the entire url, it's just the slug for that specific article.

Here's how it looks. 

image

SebastiaanZ (approved):

Issue #290, opened 2019-10-20
Move tests to a single tests/ folder


gdude2002:
As discussed on Discord, we currently have tests for each app placed within the app's directory - but this leaves us without a clean place to test things like utils, which aren't unique to the app.

Instead, all tests should be moved to `pydis_site/tests`, with a subfolder for each app - much like we did with the templates and static files.

ju-sh:
I'll give a shot at this.

scragly:
This is still in active discussion and we're not yet decided if this is something we're going to do. 

As a note for future @ju-sh, it's best to wait for a reply and to be assigned by an org member before going ahead with any PRs.

gdude2002:
Sorry, that's my bad. I was given the go-ahead to apply the `good first issue` label before a discussion prompt came up, and I forgot to apply the `discussion` label at that point.

Xithrius:
The stance on this as of now is that there doesn't seem to be a large pushing reason to do this change, even after 2 years.

If someone really wants to implement this, then this issue can be re-opened.

PR #295, opened 2019-10-21
organize tests of all apps into one directory

ju-sh:
Fixes issue https://github.com/python-discord/site/issues/290


scragly:
As explained in #290, this PR is a bit premature so will be marked as invalid and closed for now until internal discussion has finished.

PR #300, opened 2019-10-25
Create offensive message model for python-discord/bot#617

### LINKED ISSUE #222: Offensive word filtering
Once the bot's word filters are triggered, we will need to call a site endpoint to store the ID of the offending message, along with a "delete time" of one week after the authored time.

The bot will pull a list of offending message IDs daily, and delete the messages that have reached the allotted delete time.

This feature will be worked on after the Django rewrite is complete.

Akarys42:
Hi, this PR add a new offensive message model needed for python-discord/bot#617. 

This PR closes #222.


pydis_site/apps/api/migrations/0046_offensivemessage.py
```diff
@@ -0,0 +1,25 @@
...
+from django.db import migrations, models
+import pydis_site.apps.api.models.bot.offensive_message
+import pydis_site.apps.api.models.utils
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0045_add_plural_name_for_log_entry'),
+    ]
```
SebastiaanZ: This will conflict with the already merged reminder jump_url migration.

pydis_site/apps/api/migrations/0046_offensivemessage.py
```diff
@@ -0,0 +1,25 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='OffensiveMessage',
+            fields=[
+                ('id', models.BigIntegerField(help_text='The message ID as taken from Discord.', primary_key=True, serialize=False, validators=[django.core.validators.MinValueValidator(limit_value=0, message='Message IDs cannot be negative.')])),
+                ('channel_id', models.BigIntegerField(help_text='The channel ID that the message was sent in, taken from Discord.', validators=[django.core.validators.MinValueValidator(limit_value=0, message='Channel IDs cannot be negative.')])),
+                ('delete_date', models.DateField(help_text='The date on which the message will be auto-deleted.', validators=[pydis_site.apps.api.models.bot.offensive_message.future_date_validator])),
```
SebastiaanZ: Why did you choose to only save the date instead of a datetime precisely a week after the original message? We already have a `Scheduler` base class in the `bot` repo that can be used to create scheduled deletions, just like how it's used to remove infractions after a certain period. I think that's going to save an additional custom implementation of the removal-after-delay logic.

pydis_site/apps/api/viewsets/bot/offensive_message.py
```diff
@@ -0,0 +1,66 @@
...
+class OffensiveMessageViewSet(
+    CreateModelMixin, ListModelMixin, DestroyModelMixin, UpdateModelMixin, GenericViewSet
+):
+    """
+    View providing CRUD access to reminders.
+
+    ## Routes
+    ### GET /bot/offensive-message
+    Returns all offensive messages in the database.
+
```
SebastiaanZ: Does this route take any parameters or are we always returning all the entries in the database?

pydis_site/apps/api/viewsets/bot/offensive_message.py
```diff
@@ -0,0 +1,66 @@
...
+    ### DELETE /bot/offensive-message/<id:int>
+    Delete the offensive message object with the given `id`.
+
+    #### Status codes
+    - 204: returned on success
+    - 404: if a offensive message object with the given `id` does not exist
+
+    ## Authentication
+    Requires an API token.
```
SebastiaanZ: Are you planning to use this endpoint to delete the entries after the corresponding messages have been successfully removed from the guild? I don't think I've seen it in use in the `bot` PR, but I've only taken a cursory view at that one. If we're not going to use it, I don't think we should expose a delete method.

pydis_site/apps/api/viewsets/bot/offensive_message.py
```diff
@@ -0,0 +1,66 @@
...
+    ListModelMixin,
+    UpdateModelMixin
+)
+from rest_framework.viewsets import GenericViewSet
+
+from pydis_site.apps.api.models.bot.offensive_message import OffensiveMessage
+from pydis_site.apps.api.serializers import OffensiveMessageSerializer
+
+
+class OffensiveMessageViewSet(
```
SebastiaanZ, line 12: I'd like to see tests for this model that guarantee the API we want for it. This means that if we ever change anything in the settings later, the tests will reveal if it had any undesirable side-effects on the endpoint.

SebastiaanZ (changes requested):
I think it would also be a good idea to add the model to the admin back-end, so we can inspect it if needed.

Akarys42:
Updated.

pydis_site/apps/api/tests/test_offensive_message.py
```diff
@@ -0,0 +1,61 @@
...
+    def test_returns_400_on_non_future_date(self):
+        url = reverse('bot:offensivemessage-list', host='api')
+        delete_at = datetime.datetime.now() - datetime.timedelta(days=1)
+        data = {
+            'id': '602951077675139072',
+            'channel_id': '291284109232308226',
+            'delete_date': delete_at.isoformat()[:-1]
+        }
+        response = self.client.post(url, data=data)
+        self.assertEqual(response.status_code, 400)
```
SebastiaanZ, line 42: There are several reasons for why the API could return a `400` response; I think would be a good idea to assert that you're getting the appropriate response here by also asserting the message the API returns. See this infraction test, for an example:

https://github.com/python-discord/site/blob/8cc1f03297598629b27fb0b1adab82160505e879/pydis_site/apps/api/tests/test_infractions.py#L215-L218

This also holds for the other test methods, but I think you get the idea.

pydis_site/apps/api/migrations/0047_merge_20191025_1920.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0046_reminder_jump_url'),
+        ('api', '0046_offensivemessage'),
+    ]
+
+    operations = [
+    ]
```
SebastiaanZ: I don't think it's necessary to include a merge migration, since the diverging migration paths are not actually a part of master yet. To keep a clean history, I would just make sure that your migration nicely follows the current migrations we have. (In other words, this is only necessary for your local set-up since other branches have been merged in the mean time; I think it's better to clean up the local migration history instead of making it more complex.)

pydis_site/apps/api/migrations/0048_auto_20191025_1921.py
```diff
@@ -0,0 +1,19 @@
...
+        ('api', '0047_merge_20191025_1920'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='offensivemessage',
+            name='delete_date',
+            field=models.DateTimeField(help_text='The date on which the message will be auto-deleted.', validators=[pydis_site.apps.api.models.bot.offensive_message.future_date_validator]),
+        ),
+    ]
```
SebastiaanZ: This is another migration that's necessary for you locally, since you're migrating from one local database state to another by altering the `delete_date` field from a `DateField` to a `DateTimeField`, but is not that relevant for public history. Your pull request basically introduces a single, atomic change in the database: a new model; ideally, this is just captured in a single migration that handles that.

pydis_site/apps/api/viewsets/bot/offensive_message.py
```diff
@@ -0,0 +1,62 @@
...
+    CreateModelMixin, ListModelMixin, DestroyModelMixin, UpdateModelMixin, GenericViewSet
```
SebastiaanZ: Is there ever a case where we use PUT or PATCH methods to update the entries in the table? I don't see it mentioned in the docstring and there are no tests for those methods in `test_offensive_message`, but you did include the [`UpdateModelMixin`](https://www.django-rest-framework.org/api-guide/generic-views/#updatemodelmixin) here, which will implement these (partial/full) update methods for existing entries.

This is related to the next comment as well, but I don't really see a use-case for updating existing entries; we mainly want to be able to add them, list them, and delete them when we're done with them.

Using these mixins:
```suggestion
    CreateModelMixin, ListModelMixin, DestroyModelMixin, GenericViewSet
```
will make sure we're able to create, list, and destroy entries, but nothing else.

All these actions should be confirmed by tests and, preferably, we should test that using a method that's not allowed (PUT/PATCH) results in a `405` response.

pydis_site/apps/api/viewsets/bot/offensive_message.py
```diff
@@ -0,0 +1,62 @@
...
+):
+    """
+    View providing CRUD access to reminders.
```
SebastiaanZ: Probably a mistake from copy-pasting, but "reminders" does not fit here. I also don't think we should provide an `update` interface, because I don't really see a use-case for it. We either add, list, or delete the offensive messages here; the message id, channel id, and deletion date should probably not change.

pydis_site/apps/api/viewsets/bot/offensive_message.py
```diff
@@ -0,0 +1,62 @@
...
+    #### Status codes
+    - 201: returned on success
+    - 400: if the body format is invalid
+
+    ### DELETE /bot/offensive-message/<id:int>
+    Delete the offensive message object with the given `id`.
+
+    #### Status codes
+    - 204: returned on success
+    - 404: if a offensive message object with the given `id` does not exist
```
SebastiaanZ: I don't see tests for this API interface, but it's an important API method to have.

SebastiaanZ (changes requested):

SebastiaanZ (approved):

pydis_site/apps/api/urls.py
```diff
@@ -33,6 +34,10 @@
     'nominations',
     NominationViewSet
 )
+bot_router.register(
+    'offensive-message',
```
MarkKoz: I think this should be plural to be consistent with other routes. Docs would need to be updated with the correct route too.

MarkKoz (changes requested):

MarkKoz (approved):

PR #303, opened 2019-10-28
Show attachments in staff log

### LINKED ISSUE #549: Message filtering did not account for message with only attachments
There was a `mute` issued to an user after posting 5 messages with 5 attachments. The bot issued a `mute` and treated accordingly against message spamming behaviour, however in this case there are two problems:
- They were messages with empty message contents, only attachments.
- The attachments were completely deleted, and there is no way to know what they were.

The log of the case looks like this
```
Deleted by: Python#4329
Date: Oct. 23, 2019, 6:54 a.m.
A̷k̴a̴s̵h̶#4646Oct. 23, 2019, 6:54 a.m. | User ID: 330256174098284554
A̷k̴a̴s̵h̶#4646Oct. 23, 2019, 6:54 a.m. | User ID: 330256174098284554
A̷k̴a̴s̵h̶#4646Oct. 23, 2019, 6:54 a.m. | User ID: 330256174098284554
A̷k̴a̴s̵h̶#4646Oct. 23, 2019, 6:54 a.m. | User ID: 330256174098284554
A̷k̴a̴s̵h̶#4646Oct. 23, 2019, 6:54 a.m. | User ID: 330256174098284554
```

The following codes might be of interest:
```py
    """Detects duplicated messages sent by a single user."""
    relevant_messages = tuple(
        msg
        for msg in recent_messages
        if (
            msg.author == last_message.author
            and msg.content == last_message.content
        )
    )

    total_duplicated = len(relevant_messages)

    if total_duplicated > config['max']:
        ...
```
Which is in [duplicates.py](https://github.com/python-discord/bot/blob/master/bot/rules/duplicates.py)

Akarys42:
Hi, this PR show the attachments along with the original message in staff logs (https://staff.pythondiscord.com/bot/logs/...). python-discord/bot#630 need to be merged after this one for the feature to work.

This PR closes python-discord/bot#549

### Before 
before

### After 
after


SebastiaanZ:
Did you test what happens to these attachment links when a message has been deleted (for a while)? I can imagine Discord not keeping attachments around if they're attached to a message anymore.

Akarys42:
No discord does not keep them for very long, but I don't think we can do
something against it, because of privacy.

SebastiaanZ:
I've just tested it and the attachments are basically inaccessible as soon as the message gets deleted. That means I don't really see the point in adding the URLs to the database. I did not consider that the other day when I commented on the issue, but I don't think this is a suitable solution.

Akarys42:
Actually that depend on discord decide to react, the proxy_url isn't
disabled immediately in most of the cases

Akarys42:
I'm not a lawyer, but I looked through discord's privacy policy and it seem
like we can save the images as long as we "use such information only to
provide the SDK/API functionality within their applications and/or
services.".

Akarys42:
I did more research and here is what I found (again, I’m not a lawyer).

# [Discord developer ToS](https://discordapp.com/developers/docs/legal)
### 2.1 API or SDK licence
> Subject to the terms and conditions of this Agreement, Discord grants you a limited, non-exclusive, revocable, non-transferable license to access and use the API or SDK solely to as necessary to (i) develop, test, display and execute your Applications with the API or SDK functionality included; and (ii) distribute to third parties, or permit third parties to access, the API or SDK as integrated into your Applications.
### 2.4 End User data
> “End User Data” means all data associated with the content within the functionality enabled by the API or SDK, including but not limited to message content, message metadata, voice data and voice metadata. You acknowledge and agree that Discord and its end users, as applicable, retain all worldwide right, title and interest in and to the End User Data. If you have access to End User Data through the API or the SDK, you shall ensure that your Applications do not collect, use and/or disclose End User Data except (a) as necessary to exercise your rights under this Agreement, (b) in accordance with Discord’s Privacy Policy. If you have access to End User Data through the SDK, you additionally agree to get express permission from the End Users.

> You shall not sell, license or otherwise commercialize any End User Data. You may not use End User Data to target End Users for marketing or advertising purposes. You may not disclose End User Data to any ad network, data broker or other advertising or monetization related service.

> You shall delete all End User Data upon Discord’s or the End User’s request and within seven (7) days following the End User’s termination of the End User’s account. You may only retain chat logs as necessary for the operation of your Applications. Discord’s offering of an API or SDK that allows you to reproduce or display End User message content is not an express or implied license to such content from Discord.

# My point of view 
From what I understand, we can’t use the end user data, including attachments, except for exercising our rights (2.4a). Using the API functionalities to produce our services are part of our rights (2.1), so we should be able to store attachments. 

# Solutions
We can either
* Store the attachments in the database, but attachments can be up to 100Mo, which is a lot to store (although if the database is on a block storage, that’s not a problem)
* Make the bot repost the attachments in a private channel, and use the new link. I think this option is also more safe from a legal point, because all the data stay on discord’s servers.

ikuyarihS:
Hmm, how about reuploading the attachments to the mod log, then use that link instead? That way we can access it much later as well, and discord will maintain it.

Akarys42:
@SebastiaanZ said some mods wouldn’t want all the attachments to appear in mod-log or mod-alert, because that take too much place (tbh I kind of agree with them), this is why I suggested to create a special bot-only channel.

MarkKoz (changes requested):
There should be a test for the attachment images being in the HTML, similar to the embed tests.

SebastiaanZ:
@Akarys42 Another channel just for the occasional image attachment relay seems a bit like overkill to me. That means we'd have a mostly empty channel and there are already a lot of channels in that log category for moderators.

Since we're deleting messages, maybe we should make an exception and have the attachments relayed to `message-change-log`. We normally suppress bulk deletions (like the antispam deletions), but having those attachments relayed there wouldn't be too bad. I think (I hope) most mods have that channel muted since it has a huge volume, so the occasional attachment being relayed there probably wouldn't hurt anyone.

SebastiaanZ:
Okay, maybe we should just open another channel. I'm going to discuss this with the admins/mods to gauge everyone's preferences.

MarkKoz (approved):

jchristgit (changes requested):
Your migration conflicts with `0049_offensivemessage.py`. Can you update to the latest master?

Akarys42:
Done! @jchristgit

jchristgit (approved):
:shipit: :shipit: :shipit: :shipit: :shipit:

PR #304, opened 2019-10-31
Make newlines visible in deleted messages

### LINKED ISSUE #302: Make newlines visible in the deleted messages front-end
The deleted messages front-end currently doesn't display newlines if they're consecutive, i.e., if the lines are otherwise empty (`"\n\n\n\n\n\n\n\n\n"`). This makes it difficult to interpret what actually happened in chat. A good solution to this would be to make newlines character visible using the `↵` character, printed in a light gray color (similar to how IDEs do it).

**before**

**proposed after**

SebastiaanZ:
This pull requests makes the newlines in deleted messages visible using the `↵` symbol. This should make it easier to correctly judge the deleted messages that show up in our deleted messages front-end. 

**Screenshot**

linebreak_example

---

In addition, I've kaizened the colour filter that translates integer representations of colours to their RGB hex-value. The Discord dark theme shows black colours (int: 0; hex: #000000) as white (#FFFFFF) to make reading text having that color easier against the dark background. This pull request makes sure our front-end displays the same behavior.

---

This pull request closes #302


jchristgit (approved):

pydis_site/apps/staff/tests/test_deletedmessage_filters.py
```diff
@@ -18,16 +18,49 @@ class Colour(enum.IntEnum):
...
+            (Colour.RED, "#FF0000"),
+            (Colour.WHITE, "#FFFFFF"),
+
+            # Since we're using a "Discord dark theme"-like front-end, show black text as white.
+            (Colour.BLACK, "#FFFFFF"),
+        )
+
+        for colour, hex_value in test_values:
+            with self.subTest(colour=colour, hex_value=hex_value):
+                self.assertEqual(deletedmessage_filters.hex_colour(colour), hex_value)
```
aeros, line 34:
```python
         for colour, hex_value in test_values:
             actual_hex = deletedmessage_filters.hex_colour(colour)
             with self.subTest(colour=colour, actual_hex=actual_hex, 
                                             expected_hex=hex_value):
                 self.assertEqual(actual_hex, expected_hex)
```

This has no functional difference, but it would make it easier to tell at a glance what the issue was if/when the tests fail. I find that making use of the var names also helps to self-document the exact purpose of the tests.

pydis_site/apps/staff/tests/test_deletedmessage_filters.py
```diff
@@ -18,16 +18,49 @@ class Colour(enum.IntEnum):
...
+            ),
+            (
+                "Consecutive linebreaks\n\n\nwork, too",
+                f"Consecutive linebreaks{html_br}{html_br}{html_br}work, too"
+            )
+        )
+
+        for input_, expected_output in test_values:
+            with self.subTest(input=input_, expected_output=expected_output):
+                self.assertEqual(deletedmessage_filters.visible_newlines(input_), expected_output)
```
aeros, line 66:
```python
        for text, expected_output in test_values:
            actual_output = deletedmessage_filters.visible_newlines(text)
            with self.subTest(text=text, actual_output=actual_output, 
                                            expected_output=expected_output):
                self.assertEqual(actual_output, expected_output)
```

aeros (commented):
I definitely approve of the unit test output changes, particularly with organizing them into subtests, and iterating over a tuple of actual and expected results instead of a wall of `self.assertEqual()`. Here's a couple of minor suggested improvements along the same lines (making the tests more readable and easier to debug), feel free to rename the vars or reformat:

lemonsaurus (approved):
I think we should get this merged.

pydis_site/apps/staff/tests/test_deletedmessage_filters.py
SebastiaanZ: Wouldn't the `AssertionError` message for an `assertEqual` test already contain both values?

eivl (approved):
Approved on the basis that we are sure that we only need to check for pure black.

PR #308, opened 2019-11-11
Add pipenv run start command

Akarys42:
Hi, 
This PR just add a new `pipenv run start` command that would be _very_ useful to people that run the site without docker.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -43,6 +43,7 @@ unittest-xml-reporting = "~=2.5.1"
 python_version = "3.7"
 
 [scripts]
+start = "python manage.py run"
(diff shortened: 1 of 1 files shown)


scragly:
Requires `--debug` flag when using `run` during development.

Akarys42:
@scragly I know, this command allows you to do `pipenv run start --debug` too.

scragly:
We aren't about to use `pipenv run` during production, so there's no reason to omit it from the command in it's entirety.

Akarys42:
@scragly flag added :heavy_check_mark:

MarkKoz (approved):

scragly (approved):

PR #317, opened 2019-12-19
Fix Infraction UniqueTogetherValidator

MarkKoz:
**Must merge python-discord/bot/pull/705 first**

The active infractions queryset only gets filtered by the fields specified. This meant that if the same user and type had another infraction instance which was active, the validator would fail. The validator assumes failure if it sees any items still in the queryset after filtering.

By including the active field in the validator, the queryset will be filtered by the active field too. In the case described above, the queryset would end up empty because a no infractions which are active will ever match an active=False filter.

A consequence of doing this is that the active field is now explicitly required - it no longer has a default value.


MarkKoz:
@SebastiaanZ The side affect of adding `active` to the validator's fields is that it is now a required field. `ModelSerializer` does not "inherit" the `default` value specified in the model, despite `active` having a default. Setting it in `InfractionSerializer` like so

```
class InfractionSerializer(ModelSerializer):
    """A class providing (de-)serialization of `Infraction` instances."""
    active = BooleanField(default=True)
```

does not work because not all infraction types are allowed to be active. In fact, it's another bug (!!!!) that the validator will incorrectly allow such types to be created as active if the `active` field is simply not specified in the payload.

The simplest fix is to explicitly require `active`. This may make sense from a design point of view anyway because it is otherwise confusing that some infraction types need to default to `True` while others to `False`. Alternatively, can look into advanced customisation of the serializer to set a default based on the `type`. The model also currently doesn't have any validation let alone setting the default value correctly based on type, so that's a larger issue.

SebastiaanZ:
Okay, that makes sense. I think I'm in favor of:

> The simplest fix is to explicitly require active. This may make sense from a design point of view anyway because it is otherwise confusing that some infraction types need to default to `True` while others to `False`.

Using a default value of `True` when we do not even want to allow that value for some entries in the first place doesn't sound like it passes the principle of least surprise to me. We could, in theory, use a default value of `False`, since that's a valid value for all infraction types, and have active infractions specify that explicitly (turning the current situation upside down) or just require an explicit value for this field.

Since the `active` field is fundamental to an infraction's nature, I think that requiring an explicit value makes the most sense: If you accidentally forget to set it, the API complains instead of assuming a value for you.

SebastiaanZ (approved):
Nice work.

GhostofGoes (approved):

Issue #244, opened 2019-09-14
Django: Write tests for the ReminderViewSet


lemonsaurus:
We currently don't have any tests for this ViewSet. It still has 100% coverage because it doesn't implement any custom methods, but we should still be testing if the endpoint provides the functionality that we expect.

bendiller:
I'd love to work on this if it's open. Seems like a good introductory issue for me.

MarkKoz:
Sure

Den4200:
Already done.

PR #322, opened 2020-01-13
Reminder tests

bendiller:
I believe this is ready for review. I'm not very familiar with Django (let alone the REST framework module), nor am I very familiar with testing overall, but I tried to model my efforts after the examples already present for other ViewSets.

I'll be happy to make any changes necessary, thanks for your time!


pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -0,0 +1,139 @@
...
+        self.data = {
+            'author': self.author.id,
+            'content': 'Remember to...wait what was it again?',
+            'expiration': datetime.utcnow().isoformat(),
+            'jump_url': "https://www.google.com",
+            'channel_id': 123,
+        }
+        url = reverse('bot:reminder-list', host='api')
+        response = self.client.post(url, data=self.data)
+        self.assertEqual(response.status_code, 201)
```
MarkKoz: This should be in a test function instead of in the setup fixture.

pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -0,0 +1,139 @@
...
+
+class ReminderCreationTests(APISubdomainTestCase):
+    def setUp(self):
+        super().setUp()
+        self.author = User.objects.create(
+            id=1234,
+            name='Mermaid Man',
+            discriminator=1234,
+            avatar_hash=None,
+        )
```
MarkKoz: It'd be better to create the user in `setUpTestData` since it runs once for the entire test case instead of creating a user once per test.

pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -0,0 +1,139 @@
...
+    def test_reminder_in_full_list(self):
+        url = reverse('bot:reminder-list', host='api')
+        response = self.client.get(url)
+
+        self.assertEqual(response.status_code, 200)
+
+        self.data['expiration'] += 'Z'  # Massaging a quirk of the response time format
+        self.data['active'] = True
+        self.data['id'] = 1
+        self.assertEqual(response.json(), [self.data])
```
MarkKoz: This isn't really a creation test. If you want to ensure the object exists, I suggest doing an assertion for it in the actual creation test:

```py
self.assertIsNotNone(Reminder.objects.filter(id=...).first())
```

pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def test_patch_updates_record(self):
+        url = reverse('bot:reminder-detail', args=(self.reminder.id,), host='api')
+        response = self.client.patch(url, data=self.data)
+        self.assertEqual(response.status_code, 200)
+
+        url = reverse('bot:reminder-list', host='api')
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json()[0]['content'], self.data['content'])
```
MarkKoz: Use `Reminder.objects.get(id=...)` and compare the fields of that object instead of using the list endpoint.

pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -0,0 +1,139 @@
...
+        url = reverse('bot:reminder-detail', args=('something',), host='api')
+        response = self.client.delete(url)
+
+        self.assertEqual(response.status_code, 404)
+
+    def test_delete_known_reminder_returns_204(self):
+        url = reverse('bot:reminder-detail', args=(self.reminder.id,), host='api')
+        response = self.client.delete(url)
+
+        self.assertEqual(response.status_code, 204)
```
MarkKoz, line 110: I suggest an assertion to ensure the object no longer exists in the database.

pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -0,0 +1,139 @@
...
+        self.assertEqual(response.json(), [])
+
+    def test_delete_returns_404(self):
+        url = reverse('bot:reminder-detail', args=('1234',), host='api')
+        response = self.client.delete(url)
+
+        self.assertEqual(response.status_code, 404)
+
+
+class ReminderCreationTests(APISubdomainTestCase):
```
MarkKoz, line 49: This lacks tests for invalid data.

MarkKoz (changes requested):
Needs a test case for the list endpoint to test retrieval of existing objects and also test searching and filtering.

pydis_site/apps/api/tests/test_reminders.py
bendiller: That's a much better approach, thanks.

pydis_site/apps/api/tests/test_reminders.py
bendiller: Will do!

pydis_site/apps/api/tests/test_reminders.py
bendiller: True. Would you suggest a separate test for each field that can be invalid? I'm particularly weak on testing and trying to learn best practices.

pydis_site/apps/api/tests/test_reminders.py
bendiller: Good point.

pydis_site/apps/api/tests/test_reminders.py
bendiller: I thought that was odd too. I used a couple existing files created by respected contributors as models for mine, and this "assert within setUp()" appeared in a couple places. Would it be appropriate to move this into the `test_reminder_in_full_list` test? Or is it better to test creation using a `post` against the list endpoint in one test, and use another test to check that a `get` for the list endpoint returns expected reminder instances?

pydis_site/apps/api/tests/test_reminders.py
bendiller: This is another case of using examples from the existing code base. I like your suggestion better - I didn't consider using the models directly rather than trying to shoe-horn everything into the API endpoints.

pydis_site/apps/api/tests/test_reminders.py
MarkKoz: I think a single test for _anything_ invalid is adequate to ensure the API reacts appropriately to erroneous data.

Here's my perspective on why:

Testing each field may be more applicable to model tests (which we don't do anywhere I believe, but maybe we should...), but these are API/viewset tests. The viewset is using standard mixins/base classes from DRF without any overrides. Those mixins are tested by DRF themselves. If the viewset used custom mixins and/or custom method handlers, then it would be a different situation regarding testing.

pydis_site/apps/api/tests/test_reminders.py
MarkKoz: Put it in a new test function. As per [this](https://github.com/python-discord/site/pull/322#discussion_r373799373) comment, put that assertion in this new function too and `test_reminder_in_full_list` can be removed completely.

MarkKoz: Actually, `test_reminder_in_full_list ` may be moved to another test case since retrieval tests are needed. It just doesn't belong with the creation tests at least.

> Needs a test case for the list endpoint to test retrieval of existing objects and also test searching and filtering.

pydis_site/apps/api/tests/test_reminders.py
bendiller: That's very helpful, I lost sight of that context regarding the DRF mixins and their existing tests.

bendiller:
> Needs a test case for the list endpoint to test retrieval of existing objects and also test searching and filtering.

I think I've got most of these issues sorted, but I'm not certain I've done what you asked regarding searching and filtering. I adapted what looked like the requested behavior from `test_infractions.test_filter_search()`. If what I've done is correct, can you help me understand when this searching is actually used in practice? Is this the mechanism by which the bot lists the reminders for a particular user, when commanded?

I'm going to push what I've got for your further comments - thanks for your feedback and help!

Oh, and because I didn't mention it in the original PR, this addresses issue #244

pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -46,35 +46,36 @@ def test_delete_returns_404(self):
...
-
-        self.assertEqual(response.status_code, 200)
-
-        self.data['expiration'] += 'Z'  # Massaging a quirk of the response time format
-        self.data['active'] = True
-        self.data['id'] = 1
-        self.assertEqual(response.json(), [self.data])
+        response = self.client.post(url, data=data)
+        self.assertEqual(response.status_code, 400)
+        self.assertRaises(Reminder.DoesNotExist, Reminder.objects.get, **{'id': 1})
```
MarkKoz: Wouldn't this be the same thing? Not sure.

```suggestion
        self.assertRaises(Reminder.DoesNotExist, Reminder.objects.get, id=1)
```

pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -46,35 +46,36 @@ def test_delete_returns_404(self):
...
             'content': 'Remember to...wait what was it again?',
             'expiration': datetime.utcnow().isoformat(),
             'jump_url': "https://www.google.com",
             'channel_id': 123,
         }
         url = reverse('bot:reminder-list', host='api')
-        response = self.client.post(url, data=self.data)
+        response = self.client.post(url, data=data)
         self.assertEqual(response.status_code, 201)
+        self.assertIsNotNone(Reminder.objects.get(id=1))
```
MarkKoz: My concern is `get()` raises an error rather than returning `None`. A test failure due to an uncaught exception may be less clear than an assertion failing.

pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -107,6 +107,7 @@ def test_delete_known_reminder_returns_204(self):
         response = self.client.delete(url)
 
         self.assertEqual(response.status_code, 204)
+        self.assertRaises(Reminder.DoesNotExist, Reminder.objects.get, **{'id': self.reminder.id})
```
MarkKoz: As mentioned in another suggestion, I think this can just be a regular kwarg.

```suggestion
        self.assertRaises(Reminder.DoesNotExist, Reminder.objects.get, id=self.reminder.id)
```

pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -110,6 +111,42 @@ def test_delete_known_reminder_returns_204(self):
...
+        self.assertEqual(response.json(), [self.rem_dict])
+
+    def test_filter_search(self):
+        url = reverse('bot:reminder-list', host='api')
+        response = self.client.get(f'{url}?search={self.author.name}')
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), [self.rem_dict])
+
+
```
MarkKoz, line 167: A test for filtering on fields is still missing. [Here](https://github.com/python-discord/site/blob/master/pydis_site/apps/api/tests/test_infractions.py#L90)'s an example. It may be more effective to have two reminder objects created so that a filter can be used to only return one of the two objects.

MarkKoz (changes requested):
>  If what I've done is correct, can you help me understand when this searching is actually used in practice? Is this the mechanism by which the bot lists the reminders for a particular user, when commanded?

I don't know - I didn't make the viewset. I would have expected `search_fields` to be specified so it only searches on a subset of the fields. It doesn't seem to have practical use to me in this case considering the viewset also supports filtering by fields. At least for infractions it was useful because reasons could be searched with regex.

pydis_site/apps/api/tests/test_reminders.py
bendiller: Yep, yours works! Not sure where I got the idea that mine needed to be that way.

pydis_site/apps/api/tests/test_reminders.py
bendiller: Ah, I didn't consider that, thanks. I did wonder why your initial suggestion used `filter(...).first()`, and now I get it.

pydis_site/apps/api/tests/test_reminders.py
bendiller: Okay, I'm with you now (and this has helped me understand the `ReminderViewSet` code better, too).

pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -0,0 +1,196 @@
...
+        cls.rem_dict_one['expiration'] += 'Z'  # Massaging a quirk of the response time format
+        cls.rem_dict_two = model_to_dict(cls.reminder_two)
+        cls.rem_dict_two['expiration'] += 'Z'  # Massaging a quirk of the response time format
+
+    def test_reminders_in_full_list(self):
+        url = reverse('bot:reminder-list', host='api')
+        response = self.client.get(url)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), [self.rem_dict_one, self.rem_dict_two])
```
SebastiaanZ: I think this test is a bit fragile. Formally, the model has no default `ordering` defined, which makes the order in which the objects are returned "unspecified":

> If a query doesn’t have an ordering specified, results are returned from the database in an unspecified order. A particular ordering is guaranteed only when ordering by a set of fields that uniquely identify each object in the results. For example, if a name field isn’t unique, ordering by it won’t guarantee objects with the same name always appear in the same order. ([source](https://docs.djangoproject.com/en/3.0/ref/models/options/#ordering))

I'm not sure what the best approach is to making these tests robust against ordering issues, as I don't think they're sortable by default. (Otherwise we could sort both lists and compare them after a sort.)

A way around that could be to use [`TestCase.assertCountEqual`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual) which tests if the lists contain the same elements with the same counts regardless of order.

```suggestion
        self.assertCountEqual(response.json(), [self.rem_dict_one, self.rem_dict_two])
```

pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def test_filter_search(self):
+        url = reverse('bot:reminder-list', host='api')
+        response = self.client.get(f'{url}?search={self.author.name}')
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), [self.rem_dict_one, self.rem_dict_two])
```
SebastiaanZ: Same as above
```suggestion
        self.assertCountEqual(response.json(), [self.rem_dict_one, self.rem_dict_two])
```

SebastiaanZ (changes requested):

pydis_site/apps/api/tests/test_reminders.py
bendiller: Whoops, I've never really thought too much about list equality comparison in Python. Considering how convenient everything else tends to be, I just assumed it'd be dealt with automagically. `TestCase.assertCountEqual` seems like a good solution!

MarkKoz (approved):

SebastiaanZ (approved):

PR #339, opened 2020-03-05
Add more pre-commit hooks

MarkKoz:
## Relevant Issues

python-discord/organisation#138

## Description

New hooks were added for pre-commit and they will run in CI too. The `pipenv run lint` script will now run all the new hooks, including flake8. As a side effect of running in CI, flake8's output will now be shown in stdout in addition to the XML that's published.

### Hooks added

A couple of these hooks automatically apply fixes. However, they still report failure and leave any changes they make uncommitted. Therefore, the user has to commit the automatic fixes.

* `check-merge-conflict` - Check for files that contain merge conflict strings.
* `check-toml` - Attempts to load all toml files to verify syntax. 
* `check-yaml` - Attempts to load all yaml files to verify syntax. 
* `end-of-file-fixer` - Makes sure files end in a newline and only a newline.
* `mixed-line-ending` - Replaces mixed line endings with LF. 
* `trailing-whitespace` - Trims trailing whitespace.
* `python-check-blanket-noqa` - Enforce that `noqa` annotations always occur with specific codes

## Reasoning

* Dev dependencies were updated and re-pinned because why not? They were getting old enough. I'm just following along with the other repos.
* The pre-commit environment is cached to speed up builds.
* The reasoning for the use of the hack to mock the pipenv binary is detailed in the commit message.

## Additional Details

The pre-commit venv is [cached](https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops). There should only be a cache miss if the `.pre-commit-config.yaml` file changes or if the location of the Python interpreter changes (more realistically, if the Python version changes). _Maybe_ Azure wipes the cache sometimes too, but in that case pre-commit will simply re-create the environment.


SebastiaanZ (approved):
Looks good. There's one diff I don't quite get, the last line of CONTRIBUTING.MD, but that's just because I don't see a difference; I'm probably overlooking something and there's nothing wrong with the changed version.

scragly:
> the last line of CONTRIBUTING.MD

It's removed an empty line from the end of file.

SebastiaanZ:
> > the last line of CONTRIBUTING.MD
> 
> It's removed an empty line from the end of file.

That's what I suspected; the GitHub diff just isn't showing that. It shows that difference for other removed lines at the end of files, but for this one it shows up as if the line the diff is referencing has been removed and inserted unchanged.

This:


vs:


Guess it's a GH diff glitch.

scragly:
Yep. I had to download the raw files to confirm it just in case; github wasn't showing the correct diff for it as you said.

.gitattributes
```diff
@@ -8,4 +8,3 @@
 # See e.g. https://stackoverflow.com/a/38588882/4464570
 *.png binary
 *.whl binary
-
```
lemonsaurus, line 11: Is this a result of running the `end-of-file-fixer`?

If so, I don't really understand why this file doesn't get a newline at the end of the file. I thought every kind of file was supposed to have one but maybe there's more to it?

pydis_site/static/css/home/index.css
```diff
@@ -85,4 +85,3 @@ span.repo-language-dot.javascript {
         max-width: none;
     }
 }
-
```
lemonsaurus, line 88: Another special rule for css files? 

I guess there must be files then that should not have newlines to the end of the file.

lemonsaurus (approved):
Yep, this all looks good. I guess I don't fully understand the newline at end fixer, I thought it would ensure newlines at the end of any kind of file but it seems this is not the case. Can anyone shed some light on this? 

Anyway, it looks good enough to merge, and I'm sure the fixer is doing what it should be doing.

eivl (approved):
Looks fine, I do not really understand the azure-pipelines configuration well enough to comment anything other than what it looks like.

.gitattributes
MarkKoz: This was an _extra_ newline at the end of the file. It had two final newlines. Now it has one. Some things, like GitHub, git, and neovim, aren't showing the final newline as a separate line, while others, like PyCharm, are.

lemonsaurus: ahaaa

PR #348, opened 2020-03-30
Off-topic channel names non random selecting

### LINKED ISSUE #309: Track usage of off topic names to better distribute name usage.
Currently off topic names are used randomly, however randomness is pretty far from evenness. It would be nice to make use of names more evenly, to avoid recently used names being used multiple times.

One method would be to have a column tracking usage in the current iteration, a bool of True/False. `used=False` would be the default for new names. The pool of available names to pick from will be all tags with `used=False`. When a names it picked, it'll be given `used=True`. Once no names are returned for `used=False`, we just update the used column for all to be set to False again and start over again.

ks129:
I don't know, is this best way to do this, but I'm not super good with Django. Bot side implementation is simple: Under task, for request, add a new query param: `mark_used`. I'm not gonna add anything in bot side before this way to do this get approved. Should fix #309


pydis_site/apps/api/migrations/0051_offtopicchannelname_used.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('api', '0050_remove_infractions_active_default_value'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='offtopicchannelname',
+            name='used',
+            field=models.BooleanField(default=False, help_text='Show is channel already used as channel name in this round.'),
```
lemonsaurus:
```suggestion
            field=models.BooleanField(default=False, help_text='Whether or not this name has already been used during this rotation'),
```

pydis_site/apps/api/models/bot/off_topic_channel_name.py
```diff
@@ -16,6 +16,11 @@ class OffTopicChannelName(ModelReprMixin, models.Model):
         help_text="The actual channel name that will be used on our Discord server."
     )
 
+    used = models.BooleanField(
+        default=False,
+        help_text="Show is channel already used as channel name in this round."
```
lemonsaurus:
```suggestion
        help_text="Whether or not this name has already been used during this rotation",
```

pydis_site/apps/api/tests/test_off_topic_channel_names.py
```diff
@@ -21,6 +21,12 @@ def test_cannot_read_off_topic_channel_name_list_with_random_item_param(self):
 
         self.assertEqual(response.status_code, 401)
 
+    def test_cannot_read_off_topic_channel_name_list_with_random_item_and_mark_used_param(self):
```
lemonsaurus: Please add docstrings to your test functions, it makes them display better when we run them.

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -21,6 +21,10 @@ class OffTopicChannelNameViewSet(DestroyModelMixin, ViewSet):
     If the `random_items` query parameter is given, for example using...
         $ curl api.pythondiscord.local:8000/bot/off-topic-channel-names?random_items=5
     ... then the API will return `5` random items from the database.
+    If the `mark_used` query parameter is given like...
+        $ curl api.pydis.local:8000/bot/off-topic-channel-names?random_items=5&mark_used=true
+    ... then the API will mark returned `5` items `used`.
+    When running out of names, API will mark all names to not used and start new round.
```
lemonsaurus: Since you're removing this part, just move some of this information about starting new rounds into the docstring for the `list` function itself.

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -106,6 +110,31 @@ def list(self, request: HttpRequest) -> Response:
                     'random_items': ["Must be a positive integer."]
                 })
 
+            if 'mark_used' in request.query_params and request.query_params['mark_used']:
+                queryset = self.get_queryset().order_by('?').exclude(used=True)[:random_count]
```
lemonsaurus: what does `.order_by('?')` do here?

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -106,6 +110,31 @@ def list(self, request: HttpRequest) -> Response:
...
+                self.get_queryset().filter(
+                    name__in=(query.name for query in queryset)
+                ).update(used=True)
+
+                # When client request more channel names than non-used names is available, start
+                # new round of names.
```
lemonsaurus:
```suggestion
                # When the client requests more channel names than are available, 
                # we reset all names to used=False and start a new round of names.
```

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -106,6 +110,31 @@ def list(self, request: HttpRequest) -> Response:
...
+                if len(queryset) < random_count:
+                    # Get how much names still missing and don't fetch duplicate names.
```
lemonsaurus:
```suggestion
                    # Figure out how many additional names we need, and don't fetch duplicate names.
```

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -106,6 +110,31 @@ def list(self, request: HttpRequest) -> Response:
...
+                    need_more = random_count - len(queryset)
```
lemonsaurus: `need_more` is not a great variable name.

```suggestion
                    names_needed = random_count - len(queryset)
```

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -106,6 +110,31 @@ def list(self, request: HttpRequest) -> Response:
...
+                    ext = self.get_queryset().order_by('?').exclude(
```
lemonsaurus: Don't use names like `ext`. Just give it a full name.

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -106,6 +110,31 @@ def list(self, request: HttpRequest) -> Response:
...
+                        name__in=(query.name for query in queryset)
+                    )[:need_more]
+
+                    # Set all names `used` field to False except these that we just used.
```
lemonsaurus:
```suggestion
                    # Reset the `used` field to False for all names except the ones we just used.
```

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -106,6 +110,31 @@ def list(self, request: HttpRequest) -> Response:
...
+                    self.get_queryset().exclude(name__in=(
+                        query.name for query in ext)
+                    ).update(used=False)
+                    # Join original queryset (that had missing names)
+                    # and extension with these missing names.
```
lemonsaurus:
```suggestion
                    ).update(used=False)

                    # Join the original queryset (that had missing names)
                    # and the extension with these missing names.
```
Always put a newline before a block comment, if possible.

Also, I'm not sure what you're doing here, can you explain it better?

lemonsaurus (changes requested):
So, I think the overall approach (tracking names that are used with a boolean and setting everything to used=False when you run out) is the right approach. However, I don't think I see any need for this to be a parameter. This should just be the way that it always works. Why would we ever want to return names in any other way than this?

Additionally, making this change would require a change on the bot side to make use of this new optional param.

Please refactor this so that this new rotation method is the only way this works, and that way the bot doesn't need to change anything to make use of it.

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
ks129: I'm pretty sure that this should `used=False` because then name is again available.

lemonsaurus: oh, yeah. my bad.

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
ks129: This makes that ORM will return queries in random order. Removing this will make that we get always the same names.

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
ks129: This joins these missing names to the original list of names.

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
lemonsaurus: ah, nice.

lemonsaurus (approved):
These changes seem to have addressed my requests.

lemonsaurus:
This has grown very stale. Can someone on the @python-discord/core-developers team review this so we can get it merged?

@ks129 sorry for the long wait. Can you fix the failing builds? We will try to get this merged already.

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -106,7 +108,29 @@ def list(self, request: HttpRequest) -> Response:
...
+                # Figure out how many additional names we need, and don't fetch duplicate names.
+                names_needed = random_count - len(queryset)
+                other_names = self.get_queryset().order_by('?').exclude(
+                    name__in=(query.name for query in queryset)
+                )[:names_needed]
+
+                # Reset the `used` field to False for all names except the ones we just used.
+                self.get_queryset().exclude(name__in=(
+                    query.name for query in other_names)
+                ).update(used=False)
```
MarkKoz: Won't this reset the names from the first query back to unused? The exclude is only filtering by `other_names`. I think it needs to filter by both `queryset` and `other_names`.

MarkKoz (changes requested):
There's a migration conflict you need to fix. @lemonsaurus has pointed out previously that merges can be avoided by simply changing the dependency of your migration to the latest migration (assuming there are no _actual_ conflicts). Correct me if I'm wrong.

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -106,7 +108,29 @@ def list(self, request: HttpRequest) -> Response:
...
+
+                # Join original queryset (that had missing names)
+                # and extension with these missing names.
+                queryset = list(queryset) + list(other_names)
```
MarkKoz: I really had a feeling there must be a simpler way to do it. I played around with a bunch of ideas and ultimately came up with this:

```py
queryset = self.get_queryset().order_by('used', '?')[:random_count]

if any(obj.used for obj in queryset):
    self.get_queryset().update(
        used=Case(
            When(name__in=(obj.name for obj in queryset), then=Value(True)),
            default=Value(False)
        )
    )
else:
    self.get_queryset().filter(
        name__in=(obj.name for obj in queryset)
    ).update(used=True)
```

It's shorter and has only two queries rather than four. However, I don't know if it's strictly better. Maybe it's harder to follow. I'm not sure about performance since it does extra sorting and has an `any()` which iterates results. However, I imagine this to be negligible in practice given the amount of names we request. Furthermore, it is likely made up by the removal of the extra operations compared to the previous approach.

Let me know what you think. @lemonsaurus, maybe you have some comments since you reviewed it too?

pydis_site/apps/api/tests/test_off_topic_channel_names.py
```diff
@@ -10,12 +10,14 @@ def setUp(self):
         self.client.force_authenticate(user=None)
 
     def test_cannot_read_off_topic_channel_name_list(self):
+        """Test does this return 401 response code when not authenticated."""
```
MarkKoz: It's redundant to write "test" in the docstring of a test function. It's also awkward to use "does". If I were to use "test", I'd write it as 

> Test **if** this return**s** **a** 401 response when not authenticated.

Can also use "that" instead of "if". Anyway, as I said, I'd really opt for something along the lines of

> Should return a 401 response code when not authenticated.

 or

> Returned a 401 response code because of no authentication.

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
lemonsaurus: I welcome any simplification - I shared your gut feeling that there had to be a simpler way, but didn't have a great suggestion.

I think your approach looks smart, but it would require a few comments to not be scary. Especially the if clause.

pydis_site/apps/api/tests/test_off_topic_channel_names.py
lemonsaurus: I sort of agree with @MarkKoz, although docstrings should be imperative mood - so something like `Return a 401 response when not authenticated.` seems more in line with PEP257 to me.

MarkKoz (approved):

PR #352, opened 2020-04-21
News IDs storage in bot settings

ks129:
Added `news` to allowed setting names, made migration for it. Part of Python bot upcoming Python News feature.


jb3 (approved):
Works fine locally when tested.

lemonsaurus (approved):

PR #356, opened 2020-05-27
No longer accept or track `avatar_hash`

lemonsaurus:
This should completely remove avatar_hash from the site - both in our tests, in the model itself, and from the database (as a result of the migration).

This should before https://github.com/python-discord/bot/pull/962, although both should be merged roughly at the same time.

Please review this carefully.


MarkKoz (approved):
The `User` model is also created in `test_active_infraction_migration.py` with an `avatar_hash` argument. Presumably this is fine since the tests pass. I think it's working with old migrations which use an older model for the `User`.

lemonsaurus:
> The `User` model is also created in `test_active_infraction_migration.py` with an `avatar_hash` argument. Presumably this is fine since the tests pass. I think it's working with old migrations which use an older model for the `User`.

Yes, since this uses an older migration, it would actually crash without it. So that's by design.

SebastiaanZ (approved):
This will be a critical merge that is interdependent with the bot.

PR #357, opened 2020-06-02
Change roles field on user model to Postgres array

jb3:
After observing incredibly high response times for the `GET /bot/users` we investigated how Django was querying the Postgres tables relevant and found that for every sync we make 59k queries to the database across tables with over 100,000 rows. We experienced errors since uWSGI was killing workers after 30 seconds and the bot was unable to sync until we removed user data in a similar fashion to our monthly data reviews.

This PR removes the `api_user_roles` table and converts the roles field of `User` from a many-to-many field (requiring a separate table) to a postgres [ArrayField](https://docs.djangoproject.com/en/3.0/ref/contrib/postgres/fields/#arrayfield).

All relevant code that makes use of roles (primarily the OAuth2 login system) should have been altered. The `top_role` attribute of the `User` model has also been converted to return the relevant `Role` object.


lemonsaurus (approved):
Looks :ok_hand: to me.

pydis_site/apps/api/models/bot/user.py
```diff
@@ -1,10 +1,22 @@
...
 
 
+def _validate_existing_role(value: int) -> None:
+    """Validate that a role exists when given in to the user model."""
+    role = Role.objects.filter(id=value)
+
+    if not role:
+        raise ValidationError(
+            f"Role with ID {value} does not exist"
+        )
```
SebastiaanZ: You're a bit generous with the line endings here (it fits on a single line that will only reach column 70), but I don't particularly mind.

```suggestion
        raise ValidationError(f"Role with ID {value} does not exist")
```

SebastiaanZ (approved):

PR #365, opened 2020-06-19
Implement forms app

jb3:
This PR introduces a forms application which will allow for collection of survey and form results for data such as events sign ups, insights and so on, effectively replacing Google Forms which we use now.


jb3:
Lint is failing because no tests have been written. I'll sort that out once I've built the feature.

jb3:
The forms app will be implemented as a separate service for easier development. It was becoming a big thing to tie into the already complex site.

PR #346, opened 2020-03-21
Update to Django 3.0

MarkKoz:
Resolves #344

Still needs to be tested extensively. If you have ideas for what should be tested, leave a comment.

PR #368, opened 2020-07-14
Removes django-crispy-bulma as a dependency.

lemonsaurus:
django-crispy-bulma is not being maintained, and looks like it has a bleak future. I'm considering just archiving it, to be honest. 

As such, it has no place in our site. This PR removes it as a dependency, which should help unblock #346.


MarkKoz (changes requested):
Got an error when logging in as admin and then trying to delete the account

bild

lemonsaurus:
Yeah, I can see why that was happening @MarkKoz. The `django-crispy-forms` dependency was configured in settings.py to use bulma as its template folder, which doesn't exist without `django-crispy-bulma`. 

Turns out `django-crispy-forms` was only being used on that one page, so I refactored it to just use pure Django forms, and removed the `django-crispy-forms` dependency from site as well.

MarkKoz (approved):

kosayoda (approved):

PR #370, opened 2020-07-16
Add mentions field to Reminders model

kosayoda:
To facilitate [bot/#1026](https://github.com/python-discord/bot/issues/1026), this PR adds a field to store the IDs of Member and Role objects.

No validation is done other than ensuring the ID is greater than 0, since we just ignore the ID if it doesn't convert to a valid Member/Role object on the bot side.


pydis_site/apps/api/tests/test_reminders.py
```diff
@@ -54,6 +54,18 @@ def setUpTestData(cls):
...
+            name='Fish Dude',
+            discriminator=5678,
+        )
+        cls.role = Role.objects.create(
+            id=555,
+            name="Random role",
+            colour=2,
+            permissions=0b01010010101,
+            position=10,
+        )
```
MarkKoz: It isn't necessary to create these objects since there's no FK being used. Just use some integer literals for the IDs in the test.

pydis_site/apps/api/viewsets/bot/reminder.py
```diff
@@ -43,15 +50,34 @@ class ReminderViewSet(
     #### Request body
     >>> {
     ...     'author': int,
+    ...     'mentions': List[int],
     ...     'content': str,
-    ...     'expiration': str  # ISO-formatted datetime
+    ...     'expiration': str,  # ISO-formatted datetime
+    ...     'channel_id': int,
+    ...     'jump_url': Optional[str]
```
MarkKoz: I believe `jump_url` isn't optional according to the model.

MarkKoz (changes requested):
Is there any compelling reason we'd still need to track the author? Can we just lump their ID in with the rest of the mentions?

kosayoda:
> Is there any compelling reason we'd still need to track the author? Can we just lump their ID in with the rest of the mentions?

Currently, the `author` field is a FK to a `User`. Getting all the reminders set by a particular `User` is trivial, and although we could filter by an index to the array, I don't think it's ideal.

More importantly, since I'm using an `ArrayField` to store the list of IDs, there's also no way to cascade delete reminders when the author leaves the server (assuming we go through with storing the author ID in the `mentions` field).

I'll address your other changes.

MarkKoz:
Ah right we need to know the author cause we have a list reminders command.

MarkKoz (approved):

Den4200 (approved):
It looks good to me.

PR #371, opened 2020-07-16
FilterList model and endpoints

### LINKED ISSUE #305: Manage whitelisted guilds and file formats in DB.
This is part 2 of https://github.com/python-discord/bot/issues/639

In order to manage whitelist items via a command, we will need to store them in our database.

- Create models for the whitelist items
- Create a migration that adds these models, ~~and also automatically adds the currently whitelisted items, which you can find in `default-config.yml` on the bot repo. We don't wanna have to add these manually.~~ Let's just add the items manually, so that we can merge the site-side changes independently of the bot changes. It'll take like, 2 minutes.
- Lastly, ~~create API endpoints with all CRUD actions.~~ let Django Restful Framework handle all the CRUD actions.

lemonsaurus:
This pull request introduces a new model, FilterList, and the prerequisite viewset, URL endpoint, migrations, serializer and tests.

The intention of this model is to store things that we will either whitelist or blacklist. For example, we'll be storing file formats, guild invite IDs, and other things that are currently in the bot config file, but that we want to be able to modify during runtime.

Eventually, we may set up automation where this table is synchronized against stuff like the [Awesome Discord Programming List](https://github.com/mhxion/awesome-discord-communities).

**Review style: commit-by-commit.**

Closes #305.

### GET /bot/filter-lists
Returns all filterlist items in the database.

#### Response format
```js
[
    {
        'id': "2309268224",
        'created_at': "01-01-2020 ...",
        'updated_at': "01-01-2020 ...",
        'type': "file_format",
        'allowed': 'true',
        'content': ".jpeg",
        'comment': "Popular image format.",
    },
    ...
]
```

#### Status codes
- 200: returned on success
- 401: returned if unauthenticated

### GET /bot/filter-lists/\<id:int\>
Returns a specific FilterList item from the database.

#### Response format
```js
{
    'id': "2309268224",
    'created_at': "01-01-2020 ...",
    'updated_at': "01-01-2020 ...",
    'type': "file_format",
    'allowed': 'true',
    'content': ".jpeg",
    'comment': "Popular image format.",
}
```

#### Status codes
- 200: returned on success
- 404: returned if the id was not found.

### GET /bot/filter-lists/get-types
Returns a list of valid list types that can be used in POST requests.

#### Response format
```js
[
    ["GUILD_INVITE","Guild Invite"],
    ["FILE_FORMAT","File Format"],
    ["DOMAIN_NAME","Domain Name"],
    ["FILTER_TOKEN","Filter Token"]
]
```

#### Status codes
- 200: returned on success

### POST /bot/filter-lists
Adds a single FilterList item to the database.

#### Request body
```js
{
   'type': str,
   'allowed': bool,
   'content': str,
   'comment': Optional[str],
}
```

#### Status codes
- 201: returned on success
- 400: if one of the given fields is invalid

### DELETE /bot/filter-lists/\<id:int\>
Deletes the FilterList item with the given `id`.

#### Status codes
- 204: returned on success
- 404: if a tag with the given `id` does not exist


MarkKoz:
An `allowed` field doesn't make sense to me because it facilitates the simultaneous maintenance of both a blacklist and whitelist for the same type. This means every possible value would have to exist in the database as explicitly allowed or denied, which is of course infeasible. 

It makes more sense to enforce all records of a certain type to all be whitelist or all be a blacklist. Each type should have a default state (i.e. allowed or denied), and if an item of that type is present in the DB, then assume the opposite state. It can be up to the client to decide what that default state is. If you really need to have it in the DB, then it should be stored in a separate table as a mapping between type and allowed/denied rather than stored as a field for individual items in the list.

lemonsaurus:
@MarkKoz 
> An allowed field doesn't make sense to me because it facilitates the simultaneous maintenance of both a blacklist and whitelist for the same type.

Some types need both a blacklist and a whitelist, for example the Guild Invite IDs. Here the idea is that some guild invites will be blacklisted and will never be permitted, some will be whitelisted and will always be permitted, and everything else will be permitted based on certain criteria (whether they are Partnered or Verified, in this case).
 
The UniqueConstraint makes it impossible for a blacklist and a whitelist for the same item to simultaneously exist, so every item is always exclusively on one of the two lists.

MarkKoz:
Okay, that makes more sense. I was initially viewing the union of the blacklist and whitelist as the set of all possibilities.

Pipfile.lock
```diff
@@ -666,23 +645,20 @@
...
             ],
-            "markers": "python_version >= '3.5'",
             "version": "==5.0.2"
         },
         "pyflakes": {
             "hashes": [
                 "sha256:0d94e0e05a19e57a99444b6ddcf9a6eb2e5c68d3ca1e98e90707af8152c90a92",
                 "sha256:35b2d75ee967ea93b55750aa9edbbf72813e06a66ba54438df2cfac9e3c27fc8"
             ],
-            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'",
```
MarkKoz: What's with all of these being removed? Maybe it's something added by a more recent version of pipenv than what you used to lock?

pydis_site/apps/api/viewsets/bot/allow_deny_list.py
```diff
@@ -0,0 +1,83 @@
...
+    #### Status codes
+    - 204: returned on success
+    - 404: if a tag with the given `id` does not exist
+    """
+
+    serializer_class = AllowDenyListSerializer
+    queryset = AllowDenyList.objects.all()
+
+    @action(detail=False, methods=["get"])
+    def get_types(self, _: Request) -> Response:
```
MarkKoz: This endpoint does not appear to be documented above.

pydis_site/apps/api/tests/test_allowlists.py
```diff
@@ -0,0 +1,112 @@
...
+        }
+        no_content_json = {
+            "allowed": True,
+            "type": "FILE_FORMAT"
+        }
+        cases = [{}, no_type_json, no_allowed_json, no_content_json]
+
+        for case in cases:
+            response = self.client.post(URL, data=case)
+            self.assertEqual(response.status_code, 400)
```
MarkKoz: This should use a subtest. I assume Django supports those.

pydis_site/apps/api/tests/test_allowlists.py
```diff
@@ -0,0 +1,112 @@
...
+        self.assertEqual(response.status_code, 404)
+
+    def test_deleting_known_id_returns_204(self):
+        response = self.client.delete(f"{URL}/{self.jpeg_format.id}")
+
+        self.assertEqual(response.status_code, 204)
+
+    def test_name_gets_deleted(self):
+        response = self.client.delete(f"{URL}/{self.png_format.id}")
+        self.assertEqual(response.status_code, 204)
```
MarkKoz: These seem redundant.

Pipfile.lock
lemonsaurus: hmm. interesting. I'm not sure. I can try to update pipenv and relock?

MarkKoz (dismissed):

pydis_site/apps/api/serializers.py
```diff
@@ -97,6 +97,16 @@ class Meta:
...
 
 
+class AllowListSerializer(ModelSerializer):
+    """A class providing (de-)serialization of `AllowList` instances."""
+
+    class Meta:
+        """Metadata defined for the Django REST Framework."""
+
+        model = AllowList
+        fields = ('created_at', 'updated_at', 'type', 'allowed', 'content')
```
MarkKoz: I'm rusty with DRF. Will it allow `created_at` and `updated_at` to be set/changed even though they're auto fields? If yes, we should prevent that. Also, why are they listed here, but not the `id`?

Pipfile.lock
lemonsaurus: Fixed in 7f00c53

pydis_site/apps/api/serializers.py
lemonsaurus: `id` is added in a later commit.

I've tried to POST requests with `updated_at` set to a datetime ten years in the past, but it seems to just ignore it and set it to what it should be anyway, so I don't think it's possible to set them. I think it's overridden by the auto field logics, or that the auto stuff happens later in the lifecycle.

pydis_site/apps/api/tests/test_allowlists.py
lemonsaurus: resolved in 99fc3cd

pydis_site/apps/api/tests/test_allowlists.py
lemonsaurus: resolved in 99fc3cd

pydis_site/apps/api/viewsets/bot/allow_deny_list.py
lemonsaurus: Resolved in 53e4dfe

MarkKoz (approved):

pydis_site/apps/api/serializers.py
SebastiaanZ: I think that this has less to do with DRF and more to do with Django: According to the documentation, `auto_now` and `auto_now_add` fields just ignore any value that you try to pass to it. This is what `auto_now_add` says:

> Note that the current date is always used; it’s not just a default value that you can override. So even if you set a value for this field when creating the object, it will be ignored. 

The comment for `auto_add` is similar, but does not include the `creating the object` line: It always just sets the current datetime when updating the model and will ignore any custom value you try to pass to it.

pydis_site/apps/api/migrations/0025_allow_custom_inserted_at_infraction_field.py
```diff
@@ -14,6 +14,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='infraction',
             name='inserted_at',
-            field=models.DateTimeField(default=datetime.datetime.utcnow, help_text='The date and time of the creation of this infraction.'),
+            field=models.DateTimeField(default=timezone.now(), help_text='The date and time of the creation of this infraction.'),
```
SebastiaanZ: You should provide the callable so Django can use it to generate a default when it needs it. If you call it in the definition, it will be a single, static default value.

```suggestion
            field=models.DateTimeField(default=timezone.now, help_text='The date and time of the creation of this infraction.'),
```

SebastiaanZ (dismissed):
Alright, that's quite a lot of commits and files, but it looks solid. I've also ran some manual tests locally by polling the endpoints; looks good. One requested change to prevent us from using a wrong default value somewhere.

pydis_site/apps/api/migrations/0025_allow_custom_inserted_at_infraction_field.py
lemonsaurus: good catch.

SebastiaanZ (approved):
Nice work. Adding in the rework/fix/kaizen of the social account tests made it a slightly less focused PR, but I doubt fixing that would have otherwise been a major priority. Better to get it done.

lemonsaurus:
Now that we're gonna merge this pretty soon, I'll write a migration to get all the data ready.

SebastiaanZ:
Is this now ready for merge?

MarkKoz:
Best to wait until the bot PR is ready too, in case a review there requires changes here.

PR #376, opened 2020-08-21
Add deployment info and set up Sentry releases

### LINKED ISSUE #372: Deployment info
Sometimes when we run a deployment of a new version of the site, it is hard to know whether the deployment succeeded. Maybe we only changed minor things in an API endpoint, and it's hard to verify without making a bunch of API requests or SSHing into the server.

To remedy this, I'd like to make two minor changes:
- Include a comment in the base template which has the current git hash. This will allow us to then go to https://pythondiscord.com and look at the source to determine precisely which version of the site is currently running on our server.
- Add a webhook to #dev-log which lets us know when the site boots up successfully. If this webhook didn't get sent, the deployment probably failed.

### LINKED ISSUE #377: Set up Sentry releases on site
This will allow us to see which commits introduce a specific Sentry event.

lemonsaurus:
Fixes #372 
Fixes #377

This adds two small changes related to git SHA.

## Add deployment info to all views
image

This is a useful bit of information that we can use to validate our assumptions about whether a deployment was successful.

## Set up Sentry Releases
This will allow us to pinpoint exactly which commit has introduced a new bug in Sentry, and should provide valuable context to errors we experience.

image


pydis_site/utils/utils.py
```diff
@@ -0,0 +1,10 @@
+import git
+
+# Git SHA
+repo = git.Repo(search_parent_directories=True)
+GIT_SHA = repo.head.object.hexsha
+
+
+def get_git_sha() -> str:
+    """Get the Git SHA for this repo."""
+    return GIT_SHA
```
MarkKoz: This seems redundant. Is this for coverage or something?

pydis_site/templates/wiki/base.html
```diff
@@ -19,6 +19,7 @@
 {% endblock %}
 
 {% block content %}
+  <!-- Git hash for this release: {{ git_sha }} -->
```
MarkKoz: Is this redundant? It causes the comment to show up twice for wiki pages

```html
<body class="site">
  <!-- Git hash for this release: 52d7ae050d32df7ceda6888f8a60590ceb0e37dc -->

<main class="site-content">
  

  
  <!-- Git hash for this release: 52d7ae050d32df7ceda6888f8a60590ceb0e37dc -->
  
    


<nav class="navbar is-primary" role="navigation" aria-label="main navigation">
```

MarkKoz (changes requested):

pydis_site/utils/utils.py
lemonsaurus: Yep.

pydis_site/templates/wiki/base.html
lemonsaurus: oh. my mistake, I was sure the wiki base didn't extend the other base. Removing it.

pydis_site/utils/utils.py
MarkKoz: Ignoring the line with `# pragma: no cover` seems better than writing a function to get it to stop complaining. I don't know how to write a proper test for this anyway (like testing that the templates use it or something).

lemonsaurus: sorry, I misread this part. you're saying why not just import the constant

Well, we could do that, I guess I mostly made this because it feels misleading to call it a util if it's not even a function, but that's not much of a justification.

MarkKoz: How about calling it `constants.py` then?

lemonsaurus: originally this function had all three lines of code, but then I realized it would be super wasteful to check the SHA every time this was called, so I moved it up to the top of the file. And now we're just left with, well, this.

pydis_site/constants.py
```diff
@@ -0,0 +1,5 @@
+import git
+
+# Git SHA
+repo = git.Repo(search_parent_directories=True)
```
MarkKoz, line 4: Maybe this should be "private" but I'm not too fussed at this point.

MarkKoz (approved):

jb3 (approved):

PR #378, opened 2020-08-26
Add pagination and bulk update support for User(discord user) serializer.

### LINKED ISSUE #375: Make the user endpoint more scalable
Right now, we're syncing users in a hugely inefficient way. When we request the users, we get a giant 90,000 item reply, and when we send users we're sending them one at a time with 90,000 requests.

Let's not do either of these things.

This will be a cross-platform change, so see the corresponding bot issue here.

## Use DRF pagination
On the User endpoint, enable [DRF pagination](https://www.django-rest-framework.org/api-guide/pagination/) and send something like 10000 items at a time.

This will break the bot requests, so on the bot side, we'll need to implement proper handling of this pagination.

## Allow batch updates
If we POST a batch of users to this endpoint, we the endpoint should add all of these users as a single query. This is also supported by DRF, [so have a look at this part of the documentation](https://www.django-rest-framework.org/api-guide/serializers/#dealing-with-multiple-objects) for instructions on how to implement it.

RohanJnr:
closes #375 

#### Pagination
- Using built in PageNumberPagination with the default page size being 10,000 user objects(dicts) per page.
- 2 query parameteres to control page number and page size.

#### Bulk Update
- New endpoint: /bot/users/bulk_patch
- A **UserListSerializer** for `UserSerializer` has been implemented to support bulk updates.(ref to docs: [drf-Customizing multiple update](https://www.django-rest-framework.org/api-guide/serializers/#customizing-multiple-update))

## Additional

- Replace restframework_bulk package by implementing bulk creation of User models.

#### Tests for new features.
- complete


ghost:
Thank you for contributing to Python Discord!

Please check out the following documents:
- [Join the Discord server!](https://discord.gg/python) - If you have not already, join the Discord server and introduce yourself in our #dev-contrib channel.
- [Python Discord Contributor Wiki](https://pythondiscord.com/pages/contributing/) - A set of resources about setting up our services
- [Python Discord Code of Conduct](https://pythondiscord.com/pages/code-of-conduct) - Code of conduct for our community
- [CONTRIBUTING.md](https://github.com/python-discord/site/blob/master/CONTRIBUTING.md) - Guidelines on contributing to this repository

lemonsaurus:
Okay, before I really jump into this review, I need you to take a look at the `BulkSerializerMixin` and figure out whether we still need it. It looks like it was implemented specifically to solve bulk input, and we even have a dependency just for that. So I'd like you to justify your approach over that approach, and consider whether this PR should remove the dependency that includes the `BulkSerializerMixin`.

lemonsaurus:
To be clear, I'm not saying you should necessarily be using `BulkSerializerMixin` instead, but you need to at least figure out what it does and how it does it so we can make an informed decision about whether to keep it or get rid of it.

RohanJnr:
I managed to simplify the code in the bulk update PR, I no longer need to override the `to_internal_value` function, I just had to define an `id` field explicitly as mentioned in the docs(which I missed before).

Coming to the `drf-bulk` package, we are only using it for multiple creations, this can be easily achieved by overriding the `create` method in `ModelViewSet` as mentioned in the docs(I tested it and works fine, takes around 7-10 lines of code).

So, I would say that we can get rid of `drf-bulk`(by this, I mean **BulkSerailizerMixin** and **BulkCreateModelMixin** which we are using) package and implement bulk features ourselves which is fairly easy.

lemonsaurus:
Okay @RohanJnr - Then I think you should get rid of the bulk plugin.

RohanJnr:
This PR is ready for review.

RohanJnr:
The bulk update method needs optimization. It calls the database on every instance that needs to be updated.
Currently working on a fix.

RohanJnr:
here are some rough benchmarks of the API changes coming in this PR
Creating 1000 users(bulk).
0.39 seconds

Creating 10000 users(bulk).
3.14 seconds

Updating 1000 users all in 1 SQL query.
0.74 seconds

Updating 10000 users all in 1 SQL query.
14.67 seconds

Updating 10000 users by updating 1000 users in 1 SQL query.
7.12 seconds

Updating 10000 users by updating 2000 users in 1 SQL query.
6.78 seconds

Looks like 2000 users in 1 SQL query is the sweet spot.

MrHemlock (approved):
Everything seems to be in order.

ks129 (approved):
Looks good for me.

RohanJnr:
Note: This PR should not be merged until the related bot PR is ready to be merged. The bot PR depends on the changes made in this and hence, they should be merged together.

pydis_site/apps/api/serializers.py
```diff
@@ -249,15 +255,57 @@ class Meta:
...
 
-class UserSerializer(BulkSerializerMixin, ModelSerializer):
+class UserListSerializer(ListSerializer):
+    """List serializer for User model to handle bulk updates."""
+
+    def create(self, validated_data: list) -> list:
+        """Override create method to optimize django queries."""
+        present_users = User.objects.all()
+        new_users = []
+        present_user_ids = [user.id for user in present_users]
```
MarkKoz: This should be a set.

pydis_site/apps/api/serializers.py
```diff
@@ -249,15 +255,57 @@ class Meta:
...
+
+        fields_to_update.remove("id")
+        User.objects.bulk_update(updated, fields_to_update)
+        return updated
+
+
+class UserSerializer(ModelSerializer):
     """A class providing (de-)serialization of `User` instances."""
 
+    id = IntegerField(min_value=0)
```
MarkKoz, line 323: Is this explicitly specified so you can set a custom `min_value`? I feel like this should have been a DB-level constraint to begin with.

pydis_site/apps/api/serializers.py
```diff
@@ -249,15 +255,57 @@ class Meta:
...
+
+    def create(self, validated_data: list) -> list:
+        """Override create method to optimize django queries."""
+        present_users = User.objects.all()
+        new_users = []
+        present_user_ids = [user.id for user in present_users]
+
+        for user_dict in validated_data:
+            if user_dict["id"] in present_user_ids:
+                raise ValidationError({"id": "User already exists."})
```
MarkKoz: Have you considered using SQL instead to check which users exists? It may be faster. It could be a single query: `select id from users where id in (1, 2, 3, ...) limit 1` and then check if the returned length is 1. The Django way to do is with a filter and a slice.

MarkKoz: In fact, I'm pretty sure there's already a DB-level unique constraint (because the ID is the PK). This means we can rely on the constraint to fail and propagate an error.

Django's transaction control regarding `bulk_create` needs to be investigated. It says it "generally" uses one query, but what happens if it doesn't? Would it still do multiple queries in a single transaction? It's important because it needs to roll back _all_ created users if _any_ user fails to be created.

pydis_site/apps/api/serializers.py
```diff
@@ -249,15 +255,57 @@ class Meta:
...
+        Override update method to support bulk updates.
+
+        ref:https://www.django-rest-framework.org/api-guide/serializers/#customizing-multiple-update
+        """
+        instance_mapping = {user.id: user for user in instance}
+        data_mapping = {item['id']: item for item in validated_data}
+
+        updated = []
+        fields_to_update = set()
+        for user_id, data in data_mapping.items():
```
MarkKoz: It's inefficient to create the dictionary and iterate it separately. Instead, iterate `validated_data` directly.

pydis_site/apps/api/serializers.py
MarkKoz: Apparently it's because of this

> You will need to add an explicit id field to the instance serializer. The default implicitly-generated id field is marked as read_only. This causes it to be removed on updates. Once you declare it explicitly, it will be available in the list serializer's update method.

Worth adding a comment to specify why it's explicitly defined.

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -118,4 +139,75 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
...
+            return Response(resp, status=status.HTTP_400_BAD_REQUEST)
+
+        filtered_instances = queryset.filter(id__in=object_ids)
+
+        if filtered_instances.count() != len(object_ids):
+            # If all user objects passed in request.body are not present in the database.
+            resp = {
+                "Error": "User object not found."
+            }
+            return Response(resp, status=status.HTTP_404_NOT_FOUND)
```
MarkKoz: If you handle this in the serialiser instead, you'll be able to provide the ID of the object in the error message. Can be done by catching the `KeyError` at `user = instance_mapping.get(user_id)`.

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -118,4 +139,75 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
...
+            instance=filtered_instances,
+            data=request.data,
+            many=True,
+            partial=True
+        )
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
MarkKoz: Same thing but with less code. `raise_exception` will raise a 400 for invalid data. `Response` defaults to a 200.

```suggestion
        serializer.is_valid(raise_exception=True):
        serializer.save()
        
        return Response(serializer.data)
```

MarkKoz (changes requested):
From `bulk_update` docs

> If objs contains duplicates, only the first one is updated.

This is currently implicit, so you need to detect and respond with an error for duplicate IDs when doing a bulk patch.

pydis_site/apps/api/serializers.py
```diff
@@ -249,15 +255,57 @@ class Meta:
...
+            for key in data:
+                fields_to_update.add(key)
+            user = instance_mapping.get(user_id)
+            user.__dict__.update(data)
+            updated.append(user)
+
+        fields_to_update.remove("id")
+        User.objects.bulk_update(updated, fields_to_update)
```
MarkKoz, line 315: I believe this should catch `IntegrityError` and raise a `ValidatioError` since DRF doesn't handle it by default. This may be something that can be handled at a more global scale with a custom error handler (which I think is ideal) if you want to look into it further.

pydis_site/apps/api/serializers.py
```diff
@@ -249,15 +255,57 @@ class Meta:
...
+        present_users = User.objects.all()
+        new_users = []
+        present_user_ids = [user.id for user in present_users]
+
+        for user_dict in validated_data:
+            if user_dict["id"] in present_user_ids:
+                raise ValidationError({"id": "User already exists."})
+            new_users.append(User(**user_dict))
+
+        return User.objects.bulk_create(new_users)
```
MarkKoz: Same thing about `IntegrityError` applies to this.

MarkKoz (changes requested):

pydis_site/apps/api/serializers.py
RohanJnr: The bulk_create() method actually can take a keyword argument called `ignore_conflicts`, if set to True, then it will ignore errors such as `ID already exists` error(basically, ignore duplicates) and only create new users.

RohanJnr: from docs:
`On databases that support it (all but Oracle), setting the ignore_conflicts parameter to True tells the database to ignore failure to insert any rows that fail constraints such as duplicate unique values. Enabling this parameter disables setting the primary key on each model instance (if the database normally supports it).`

RohanJnr: should I use this instead? what do you think?
Then the code will be reduced to just 1 line.
```python
        return User.objects.bulk_create(
            [User(**user_dict) for user_dict in validated_data],
            ignore_conflicts=True
        )
```

The only drawback is that the client will not be notified if he sent a duplicate User in his request.

MarkKoz: That's convenient, but no. I think it is best for the client to be notified of duplicates via an error response. Otherwise, the client may not realise that only a partial update occurred due to some being skipped (e.g. in the case there are duplicate users but the fields they modify are different).

MarkKoz: Oh, this is for create. Well, I'm a bit conflicted. On one hand, I don't like the idea of silently skipping things. On the other hand, it'd allow the syncer to sync as many users as it can rather than completely failing just cause 1 user is duplicated.

Therefore, I now think that if the user is in the DB already, then ignore duplicates. However, if the request data contains the user multiple times, then that's an error. Does that sound like a reasonable compromise?

RohanJnr: right, so if the number of duplicates is more than 1, then send an error message and also stop the creation process?
How about we ignore the duplicates and instead send a warning message, notifying the client that duplicates were found and skipped?
note: The only client we have now is the python bot, and I don't think there will be a problem with multiple duplicates.

pydis_site/apps/api/serializers.py
RohanJnr: I found that, if no fields are provided, a validation error is thrown.
How does the IntegrityError get triggered?

MarkKoz: When a DB constraint is violated.

pydis_site/apps/api/serializers.py
```diff
@@ -278,16 +278,20 @@ def update(self, instance: QuerySet, validated_data: list) -> list:
...
+            for key in user_data:
                 fields_to_update.add(key)
-            user = instance_mapping.get(user_id)
-            user.__dict__.update(data)
-            updated.append(user)
+
+                try:
+                    user = instance_mapping[user_data["id"]]
+                except KeyError:
+                    raise ValidationError({"id": f"User with id {user_data['id']} not found."})
```
MarkKoz: Is there a way for this to still be a 404? Will raising `NotFound` work? Is that idiomatic?

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -16,6 +18,30 @@ class UserListPagination(PageNumberPagination):
...
+        page_number = self.page.previous_page_number()
+        return page_number
+
+    def get_paginated_response(self, data: list) -> Response:
+        """Override method to send modified response."""
+        return Response(OrderedDict([
+            ('count', self.page.paginator.count),
+            ('next_page_no', self.get_next_page_number()),
+            ('previous_page_no', self.get_previous_page_number()),
+            ('results', data)
```
MarkKoz, line 43: Don't forget to update the view docs since you've changed the paginator.

pydis_site/apps/api/serializers.py
MarkKoz: It's not about whether there are 1 or 2 or 3 duplicates. I'm suggesting it should depend on whether they are in the request or in the DB. Example: a request's data has two users with the ID 1. This should be an error. Another example: a requests data has one user with ID 1 and the DB has a user with ID 1. This should be skipped.

I don't know if this is worth the added complexity. I thought it'd make the API more robust. But it's also an option to just document that "duplicates will be skipped" and call it a day. Let's make this ignore them for now and see what anyone else has to say.

I am a bit concerned about `ignore_conflicts`. The documentation words it like it will ignore if _any_ constraint is failed, not just unique ones. Though it has a warning specifically for MariaDB/MySQL regarding this point, so maybe PostreSQL is fine?

pydis_site/apps/api/viewsets/bot/user.py
RohanJnr: right :+1:

pydis_site/apps/api/serializers.py
RohanJnr: I can write a custom validation class to return a 404 response which can be used wherever needed or I could simply do
```python
err = ValidationError({"id": f"User with id {user_data['id']} not found."})
err.status_code = 404
raise err
```

MarkKoz: It's kinda complicated according to this https://github.com/encode/django-rest-framework/issues/6555

pydis_site/apps/api/serializers.py
```diff
@@ -260,16 +261,16 @@ class UserListSerializer(ListSerializer):
...
         new_users = []
-        present_user_ids = [user.id for user in present_users]
+        request_user_ids = [user["id"] for user in validated_data]
 
         for user_dict in validated_data:
-            if user_dict["id"] in present_user_ids:
-                raise ValidationError({"id": "User already exists."})
+            if request_user_ids.count(user_dict["id"]) > 1:
+                raise ValidationError({"id": f"User with ID {user_dict['id']} "
+                                             f"given multiple times."})
```
MarkKoz: The way you broke this into multiple lines doesn't follow the established style.
```suggestion
                raise ValidationError(
                    {"id": f"User with ID {user_dict['id']} given multiple times."}
                )
```

pydis_site/apps/api/serializers.py
```diff
@@ -260,16 +261,16 @@ class UserListSerializer(ListSerializer):
...
         """Override create method to optimize django queries."""
-        present_users = User.objects.all()
         new_users = []
-        present_user_ids = [user.id for user in present_users]
+        request_user_ids = [user["id"] for user in validated_data]
 
         for user_dict in validated_data:
-            if user_dict["id"] in present_user_ids:
-                raise ValidationError({"id": "User already exists."})
+            if request_user_ids.count(user_dict["id"]) > 1:
```
MarkKoz: This is not as efficient as it could be:

```py
seen = set()

for user_dict in validated_data:
    if user_dict["id"] in seen:
        raise ValidationError

    seen.add(user_dict["id"])
    new_users.append(...)

File: pydis_site/apps/api/serializers.py [hunk 13]
```diff
@@ -260,16 +261,16 @@ class UserListSerializer(ListSerializer):
...
+                raise ValidationError({"id": f"User with ID {user_dict['id']} "
+                                             f"given multiple times."})
             new_users.append(User(**user_dict))
 
-        return User.objects.bulk_create(new_users)
+        return User.objects.bulk_create(new_users, ignore_conflicts=True)
```
MarkKoz: The behaviour of ignoring conflicts and raising errors for duplicates in the request should be documented in the viewset.

MarkKoz (changes requested):

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -118,4 +168,69 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
...
+            kwargs['many'] = True
+
+        return super().get_serializer(*args, **kwargs)
+
+    @action(detail=False, methods=["PATCH"], name='user-bulk-patch')
+    def bulk_patch(self, request: Request) -> Response:
+        """
+        Update multiple User objects in a single request.
+
+        ## Route
```
MarkKoz: This detailed endpoint info should be in the class docstring rather than in the method's docstring.

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -118,4 +168,69 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
...
+        ### PATCH /bot/users/bulk_patch
+        Update all users with the IDs.
+        `id` field is mandatory, rest are optional.
```
MarkKoz: Not quite accurate because it raises a validation error (400) if there are no fields besides the ID. Therefore, the ID and at least one other field are mandatory.

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -56,6 +104,7 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
     ### POST /bot/users
     Adds a single or multiple new users.
     The roles attached to the user(s) must be roles known by the site.
+    User creation process will be skipped if user is already present in the database.
```
MarkKoz: The way you worded it is unclear in the sense that it may imply the entire operation will abort.
```suggestion
    Users that already exist in the database will be skipped.
```

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -26,8 +69,13 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
...
     ...         'in_guild': True
-    ...     }
-    ... ]
+    ...     },
+    ...     ]
+    ... }
+
+    #### Query Parameters
+    - page_size: Number of Users in one page.
+    - page: Page number
```
MarkKoz:
```suggestion
    - page_size: number of Users in one page
    - page: page number
```

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -72,6 +121,7 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
     #### Status codes
     - 201: returned on success
     - 400: if one of the given roles does not exist, or one of the given fields is invalid
+    - 400: if multiple user objects with the same id are given.
```
MarkKoz:
```suggestion
    - 400: if multiple user objects with the same id are given
```

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -118,4 +168,69 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
...
+        ...     'discriminator': int,
+        ...     'roles': List[int],
+        ...     'in_guild': bool
+        ...     },
+        ... ]
+
+        #### Status codes
+        - 200: returned on success.
+        - 400: if the request body was invalid, see response body for details.
+        - 404: if the user with the given id does not exist.
```
MarkKoz:
```suggestion
        - 200: returned on success
        - 400: if the request body was invalid, see response body for details
        - 404: if the user with the given id does not exist
```

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -118,4 +168,69 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
...
+        """
+        queryset = self.get_queryset()
+        try:
+            object_ids = [item["id"] for item in request.data]
+        except KeyError:
+            # user ID not provided in request body.
+            resp = {
+                "Error": "User ID not provided."
```
MarkKoz: For consistency with the rest of DRF, it should be named `"detail"`.

```suggestion
                "detail": "The id field is missing from at least one object."
```

pydis_site/apps/api/serializers.py
```diff
@@ -249,15 +256,69 @@ class Meta:
...
+                    raise NotFound({"id": f"User with id {user_data['id']} not found."})
+
+                user.__dict__.update(user_data)
+                updated.append(user)
+
+        fields_to_update.remove("id")
+
+        if not fields_to_update:
+            # Raise ValidationError when only id field is given.
+            raise ValidationError({"data": "Insufficient data provided."})
```
MarkKoz: By convention, using `"data"` as a key name implies that there's something wrong with a field named `data`. If you need to raise `ValidationErrors` that aren't specific to a field, use the configured [`NON_FIELD_ERRORS_KEY`](https://www.django-rest-framework.org/api-guide/settings/#non_field_errors_key) as the key for consistency. 

Furthermore, maybe the value should be in a list since this seems to be DRF convention. I think the idea is that there could be several errors for the same field/key, hence the list.

pydis_site/apps/api/viewsets/bot/user.py
MarkKoz: Alternatively, you can raise a validation error with 

```py
raise ValidationError({"id": ["This field is required."]})
```

Which may or may not be more idiomatic here. It's in a list cause that's how DRF normally formats such errors. I am leaning more towards the `ValidationError` approach.

pydis_site/apps/api/serializers.py
```diff
@@ -249,15 +256,69 @@ class Meta:
...
+        updated = []
+        fields_to_update = set()
+        for user_data in validated_data:
+            for key in user_data:
+                fields_to_update.add(key)
+
+                try:
+                    user = instance_mapping[user_data["id"]]
+                except KeyError:
+                    raise NotFound({"id": f"User with id {user_data['id']} not found."})
```
MarkKoz: Only `ValidationErrors` use the convention of the field name as the key. 404s raised by DRF elsewhere use `detail` as the key.
```suggestion
                    raise NotFound({"detail": f"User with id {user_data['id']} not found."})
```

MarkKoz: I'm adding this comment for posterity: the reason it's okay to raise a 404 here and risk failing the entire request is because the syncer shouldn't ever run into a race condition where it's trying to update a user that has been deleted (or not created yet). The former is impossible because it never deletes users; it only sets `in_guild` as false. The latter is impossible because if it sees the user doesn't exist in the DB, it will try to create rather than update.

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -118,4 +168,69 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
...
+        ...     'discriminator': int,
+        ...     'roles': List[int],
+        ...     'in_guild': bool
+        ...     },
+        ...     {
+        ...     'id': int,
+        ...     'name': str,
+        ...     'discriminator': int,
+        ...     'roles': List[int],
+        ...     'in_guild': bool
```
MarkKoz: The dict items should be indented.
```suggestion
        ...         'id': int,
        ...         'name': str,
        ...         'discriminator': int,
        ...         'roles': List[int],
        ...         'in_guild': bool
        ...     },
        ...     {
        ...         'id': int,
        ...         'name': str,
        ...         'discriminator': int,
        ...         'roles': List[int],
        ...         'in_guild': bool
```

pydis_site/apps/api/viewsets/bot/user.py
MarkKoz: It should also be noted these are optional parameters and that the default page size is 10k.

MarkKoz (changes requested):
Apologies for so many incremental reviews. I've finally done a comprehensive review and test of the API. Along with my request changes, here are the issues I encountered:

* Posting a single user (i.e. no bulk_create) that already exists results in a 500 response because of an `IntegrityError`. This should be re-raises as 400 error instead.
* Posting multiple users which exist with `bulk_create` works but the response still includes objects for users which were duplicates. It's just a copy of the data given in the request rather than what's in the database. 

    For example, if you try to set `in_guild` from false to true for an existing user, the response will return the object with `in_guild: true` even though it never changed and is still false in the DB. 

    This isn't too important since we never check the return value, but if you know of an easy fix then you should go for it. In fact, you can respond with nothing since we don't rely on the response. This was profiled not too long ago, and it turns out that serialisation takes a significant amount of time. Granted, it won't have as much of a performance benefit at this scale (IIRC we benchmarked serialisation of all users in a single response).
* `bulk_patch` returns duplicate objects in the response. More specifically, the amount of duplicates it returns depends on the number of fields that were given in the request. Again, not too important, but fix it if you can. 
* Checking for duplicate IDs in the request doesn't work because `filtered_instances = queryset.filter(id__in=object_ids)` effectively filters out duplicates before they ever reach the serialiser. This is also indicative of a hole in your tests; there should be a test that catches this issue.

    I suppose you could merge the duplicate check with the creation of `[item["id"] for item in request.data]` if you use a normal loop instead of a comprehension and use the same concept with a set as before.

pydis_site/apps/api/serializers.py
```diff
@@ -293,7 +293,7 @@ def update(self, instance: QuerySet, validated_data: list) -> list:
                 try:
                     user = instance_mapping[user_data["id"]]
                 except KeyError:
-                    raise NotFound({"id": f"User with id {user_data['id']} not found."})
+                    raise NotFound({"detail": [f"User with id {user_data['id']} not found."]})
```
MarkKoz: Using a list is a `ValidationError` convention only
```suggestion
                    raise NotFound({"detail": f"User with id {user_data['id']} not found."})
```

pydis_site/apps/api/serializers.py
```diff
@@ -274,7 +274,8 @@ def create(self, validated_data: list) -> list:
             seen.add(user_dict["id"])
             new_users.append(User(**user_dict))
 
-        return User.objects.bulk_create(new_users, ignore_conflicts=True)
+        users = User.objects.bulk_create(new_users, ignore_conflicts=True)
+        return User.objects.filter(id__in=[user.id for user in users])
```
MarkKoz: You misunderstood what I meant by duplicates. I mean that it still returns objects which weren't created because the ID is already in the database. What you're trying to solve is removing duplicate objects in the response, which is impossible.

To fix this, you'd have to know which users are already in the DB before you do a `bulk_create`. This sounds relatively slow, so it may not be worth fixing (or return nothing) since we don't even read the response.

pydis_site/apps/api/serializers.py
```diff
@@ -235,15 +244,79 @@ class Meta:
...
+
+    def create(self, validated_data: list) -> list:
+        """Override create method to optimize django queries."""
+        new_users = []
+        seen = set()
+
+        for user_dict in validated_data:
+            if user_dict["id"] in seen:
+                raise ValidationError(
+                    {"id": f"User with ID {user_dict['id']} given multiple times."}
```
MarkKoz:
```suggestion
                    {"id": [f"User with ID {user_dict['id']} given multiple times."]}
```

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -118,4 +196,45 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
...
+                        {"id": [f"User with ID {data['id']} given multiple times."]}
+                    )
+            except KeyError:
+                # If user ID not provided in request body.
+                raise ValidationError(
+                    {"id": ["This field is required."]}
+                )
+            object_ids.add(data["id"])
+
+        filtered_instances = queryset.filter(id__in=object_ids)
```
MarkKoz: Since it's performing validation, I think it makes more sense for this code to be in the serialiser. In the hypothetical case that the serialiser needs to be used elsewhere, this code would have to be duplicated.

MarkKoz (changes requested):

MarkKoz:
I also got this warning:
```
web_1       | /usr/local/lib/python3.7/site-packages/rest_framework/pagination.py:200: UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'pydis_site.apps.api.models.bot.user.User'> QuerySet.
web_1       |   paginator = self.django_paginator_class(queryset, page_size)
```

pydis_site/apps/api/serializers.py
RohanJnr: alright, I will return an empty list then.

MarkKoz: Okay, don't forget to document that.

MarkKoz (approved):
Excellent work. Thanks for putting up with my scrutiny.

PR #402, opened 2020-10-02
Reduce Discord OAuth2 scopes

### LINKED ISSUE #341: Replace Allauth anti-email monkey-patch with proper settings.
The allauth extension we use for the discord login and connection to a github account for any of our site accounts.

We have a [monkeypatch](https://github.com/python-discord/site/blob/master/pydis_site/apps/home/apps.py#L17-L38) to avoid saving in our database any email details that we may get from oauth authorisations. This does not avoid the request for emails showing in the auth request page though, as the scope is still requested.

Instead, we can define in settings.py the appropriate provider settings, particularly any required scopes to be requested. If we only provide `identify` for the discord one, it won't add the `email` scope in the auth request page in future, making it a cleaner and more appropriate solution.

The setting would look like:
```py
SOCIALACCOUNT_PROVIDERS = {
    'discord': {
        'SCOPE': [
            'identify',
        ],
    }
}
```

The relevant scope setting for github can be given also, it just needs to be looked up as to what scopes we should restrict it to in order to avoid unnecessary sensitive data being stored.

jb3:
Closes #341 

By default, AllAuth requests both identify + email from Discord on an attempted login.

We don't have any use for this email and so we discard it immediately, but there is no reason why we need to request this scope at all.

This PR will change the Discord authenticate screen to only state that it is requesting the basic user details of a user.

Screenshot 2020-10-02 at 23 49 17

Additionally it passes the `prompt=none` parameter to the authorise URL. This means that if the user has previously authenticated they will not be asked to click confirm and will instead be redirected directly to the application, making the process much more seamless. Users can de-authenticate with us through the "Authorized Apps" section in the Discord client.

### DIFF
--- a/pydis_site/settings.py
+++ b/pydis_site/settings.py
@@ -401,3 +401,10 @@ ACCOUNT_USERNAME_VALIDATORS = "pydis_site.VALIDATORS"
 
 LOGIN_REDIRECT_URL = "home"
 SOCIALACCOUNT_ADAPTER = "pydis_site.utils.account.SocialAccountAdapter"
+SOCIALACCOUNT_PROVIDERS = {
+    'discord': {
+        'SCOPE': [
+            'identify',
+        ],
+    }
+}
--- a/pydis_site/settings.py
+++ b/pydis_site/settings.py
@@ -402,9 +402,10 @@ ACCOUNT_USERNAME_VALIDATORS = "pydis_site.VALIDATORS"
 LOGIN_REDIRECT_URL = "home"
 SOCIALACCOUNT_ADAPTER = "pydis_site.utils.account.SocialAccountAdapter"
 SOCIALACCOUNT_PROVIDERS = {
-    'discord': {
-        'SCOPE': [
-            'identify',
+    "discord": {
+        "SCOPE": [
+            "identify",
         ],
+        "AUTH_PARAMS": {"prompt": "none"}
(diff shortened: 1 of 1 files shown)


MarkKoz (approved):
For other reviewers, see these for references:

#### Scopes

https://django-allauth.readthedocs.io/en/latest/advanced.html?highlight=scope#changing-provider-scopes

https://discord.com/developers/docs/topics/oauth2#shared-resources-oauth2-scopes

#### Prompt

https://discord.com/developers/docs/topics/oauth2#authorization-code-grant

> prompt controls how the authorization flow handles existing authorizations. If a user has previously authorized your application with the requested scopes and prompt is set to consent, it will request them to reapprove their authorization. If set to none, it will skip the authorization screen and redirect them back to your redirect URI without requesting their authorization. For passthrough scopes, like bot and webhook.incoming, authorization is always required.

To test this, you need to go to Django admin and add a social application named "discord". You need to specify a client ID and secret. You also need to move `pythondiscord.local:8000` (or whatever the name is) to the right box named "chosen sites." Finally, you need to add `http://pythondiscord.local:8000/accounts/discord/login/callback/` as redirect in the Discord Developer Portal under the OAuth2 settings for your app. This process should be documented somewhere for future reference.

scragly (approved):

PR #403, opened 2020-10-03
Logging cleanup

### LINKED ISSUE #347: Remove custom logging handler and log storage
Since we have solved our logging with the integration of Sentry we have no need for log aggregation on pydis site now.

The custom log API, the methods the Django site uses to store logs in this and the actual log table should be removed from the API.

ks129:
Remove all logging-related things from site, because PyDis use now Sentry. Removed URL, viewset, model, tests. Also removed site internal logging into DB. Closes #347 .


lemonsaurus:
@scragly I think you're the only one familiar with this stuff. Can you review this? If you approve, just merge it.

scragly:
No problem. It looks absolutely fine, but I'll do a quick deploy to ensure it runs without errors too.

scragly:
Looks great, runs great. Thanks!

PR #405, opened 2020-10-04
Dewikification: Remove django-wiki from the stack

### LINKED ISSUE #382: Dewikification: Remove django-wiki from the site
**Remember to target the `dewikification` branch with this pull request**

We need to completely remove `django-wiki` from the site, as a step in our journey towards complete dewikification.

It will be useful to look at https://github.com/python-discord/site/pull/202 to see what kinds of changes we made when introducing it - this issue is to essentially do the opposite of that. However, we can't simply revert this because we've made many changes to the wiki stuff since then, so this will need to be done manually.

This will probably also require a migration, since the `django-wiki` app introduces several models which are now in our database, and we're getting rid of all this data as well.

lemonsaurus:
This PR simply annihilates `django-wiki` and everything related to it.
- All 263 tests are still passing
- Smoke tested OK, website still seems to work fine.

This does _not_ remove all the related tables from the database in production. A seperate issue exists to do this, see https://github.com/python-discord/admin-tasks/issues/37


Akarys42 (changes requested):
You still have references to `django-nyt` here

https://github.com/python-discord/site/blob/92c0627860b586e1a6414b93835cf10e27399d20/pydis_site/settings.py#L105

and there

https://github.com/python-discord/site/blob/92c0627860b586e1a6414b93835cf10e27399d20/pydis_site/apps/home/urls.py#L36

which cause the app to not start. Other than that, it looks good.

Akarys42 (approved):

Issue #406, opened 2020-10-04
Dewikification: Remove OAuth from the stack


lemonsaurus:
**Note: Remember to target `dewikification` with this PR**

After some discussion in #dev-core, we've decided we're going to continue to disentangle and simplify the site, and that the dewikification project provides a good opportunity to do this. Let's remove everything related to OAuth and the login flow from the stack, too. 

The primary purpose these served was to allow us to use a members pydis role to determine what they could access. This is a fundamentally good idea, but without the wiki in the stack it doesn't really make much sense anymore. The only thing they'd be able to authenticate into would be the admin pages, and there's no real use for granular ACL in there. It might as well just be a simple login.

On the positive, removing these two dependencies significantly reduces complexity in the application. We'd really only have two more threads to disentangle at that point - **the API** (which will eventually be broken off into a separate app), and the staff logs (which we should find a different solution for).

ks129:
This issues is resolved by merged PR #407.

PR #407, opened 2020-10-04
Dewikification: Remove django-allauth

lemonsaurus:
This strips out everything related to `django-allauth`. This means we can no longer log in, which makes sense because without the wiki, there's really not much of a reason to log in anyway. 

Our plan is to continue to push the site towards being more or less just a simple static page, and this is another step on that road.

This may be easier to review commit-by-commit.

Closes https://github.com/python-discord/site/issues/406


MarkKoz (approved):
This looks fine. Besides creating a migration, I don't think anything has been missed.

lemonsaurus:
Don't merge this yet - I need to make a migration.

PR #409, opened 2020-10-06
Added metricity db connection and user bot API

dementati:
- Adding metricity as a secondary database. 
- A script for creating a test database in the postgres docker container for local development.
- A bot API endpoint for retrieving a user's metricity data, currently including verified date and total message count.
- An abstraction for interacting with the metricity database using raw SQL queries rather than the ORM.


ghost:
Thank you for contributing to Python Discord!

Please check out the following documents:
- [Join the Discord server!](https://discord.gg/python) - If you have not already, join the Discord server and introduce yourself in our #dev-contrib channel.
- [Python Discord Contributor Wiki](https://pythondiscord.com/pages/contributing/) - A set of resources about setting up our services
- [Python Discord Code of Conduct](https://pythondiscord.com/pages/code-of-conduct) - Code of conduct for our community
- [CONTRIBUTING.md](https://github.com/python-discord/site/blob/master/CONTRIBUTING.md) - Guidelines on contributing to this repository

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -119,3 +137,15 @@ class UserViewSet(BulkCreateModelMixin, ModelViewSet):
...
+    @action(detail=True)
+    def metricity_data(self, request: Request, pk: str = None) -> Response:
+        """Request handler for metricity_data endpoint."""
+        user = self.get_object()
+        with connections['metricity'].cursor() as cursor:
+            data = {}
+            cursor.execute("SELECT verified_at FROM users WHERE id = '%s'", [user.id])
+            data["verified_at"], = cursor.fetchone()
+            cursor.execute("SELECT COUNT(*) FROM messages WHERE author_id = '%s'", [user.id])
+            data["total_messages"], = cursor.fetchone()
```
lemonsaurus: I'm not a fan of this way of unpacking the data. `data["verified_at"],` is a very surprising way to do this, in my opinion. My gut instinct was that it had to be a typo, until I realized that `fetchone()` probably returns _two_ things instead of just one, in spite of its crystal clear method name.

An immediate fix for this would just be to do `[0]` instead of the weird comma, but that's also kind of ugly. Maybe there's a beautiful way to do this?

lemonsaurus (dismissed):
Thanks for making this PR! 

I think the solution looks great, and you clearly know what you're doing. I have a few readability nitpicks, and I'd also like to like to see a PR description, and for you to set up some commit message rules in your IDE.

The first line in a commit should be limited to 50 characters, in order to prevent this:
image

In order to ensure you don't need to make style commits and linting commits, you should set up `pre-commit` if you haven't already done so -  it will catch style errors and other problems before you're allowed to commit at all.

Lastly, it looks like our builds are failing because the coverage was reduced on the user endpoint - can you write a unit test for the new endpoint?

dementati:
> The first line in a commit should be limited to 50 characters, in order to prevent this:

Will do! I did set up precommit after the fact, but I'm not sure it would've caught the thing I fixed. Anyway, I'll try to look future PRs over an extra time before I push them.

> Lastly, it looks like our builds are failing because the coverage was reduced on the user endpoint - can you write a unit test for the new endpoint?

I'm already on that as I discussed with Joe last night, I just had to stop for the night before I could get around to it.

pydis_site/apps/api/viewsets/bot/user.py
dementati: It's because it returns a tuple or list with one element. I agree that it doesn't look great. I plan to wrap the database calls in an abstraction with a nicer API, so that might help. I'll let you know when I push it and you can see what you think.

lemonsaurus:
@dementati sounds good! Looking forward to seeing the next commits. :slightly_smiling_face: 

Nice work so far!

jb3 (approved):
This looks good, works perfectly with production metricity data, is well documented and tested and is ready to go.

ks129 (approved):
Looks amazing! 🤩

PR #411, opened 2020-10-10
Implement voice bans to API

ks129:
- Allow `voice_ban` infraction type.
- Include voice banned status to metricity endpoint for better performance + cover it with test.


pydis_site/apps/api/models/bot/infraction.py
```diff
@@ -15,7 +15,8 @@ class Infraction(ModelReprMixin, models.Model):
         ("mute", "Mute"),
         ("kick", "Kick"),
         ("ban", "Ban"),
-        ("superstar", "Superstar")
+        ("superstar", "Superstar"),
+        ("voice_ban", "Voice Ban")
```
jb3: Maybe we should put a trailing comma here to stop diffs like this in future
```suggestion
        ("voice_ban", "Voice Ban"),
```

scragly:
```suggestion
        ("voice_ban", "Voice Ban"),
```

I know it wasn't there before, but future cases could be avoiding the same two-line diffs with it.

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -240,10 +242,21 @@ def bulk_patch(self, request: Request) -> Response:
     def metricity_data(self, request: Request, pk: str = None) -> Response:
         """Request handler for metricity_data endpoint."""
         user = self.get_object()
+
+        try:
+            Infraction.objects.get(user__id=user.id, active=True, type="voice_ban")
+        except ObjectDoesNotExist:
+            voice_banned = False
+        except MultipleObjectsReturned:
```
jb3: Will there ever be multiple active voice bans for a user?

These cannot be in the DB due to the one active infraction type per user constraint.

```
ERROR:  duplicate key value violates unique constraint "unique_active_infraction_per_type_per_user"
```

jb3 (changes requested):
Looking good ks!

This work fine locally, I've inserted a record and queried the endpoint, it returns the expected result.

I've added a slight comment because I think the error handling section in the viewset is handling for an impossible case.

scragly (changes requested):
Everything looks great to me, just a tiny suggested change that isn't your fault at all.

scragly (approved):

jb3 (approved):
Great stuff!

PR #413, opened 2020-10-19
Allow requesting API data from internal DNS

scragly:
Currently, we're experiencing some issues requesting API data via the public domain, likely caused by some Cloudflare weirdness. 

This change puts the internal domain `api.site`, referenceable through the internal Docker DNS only, in the whitelist of ALLOWED_HOSTS.

This should be merged before making further changes, so we can confirm the internal domain works as expected after a deployment and **before** we move change the bot to use it.


fadedmax (approved):
LGTM

jb3 (approved):
This is a good start. We'll hook up some stuff for this soon.

Eventually we'll likely use Kubernetes cluster-local DNS which may reuqire some changes, but this is a good for now.

PR #418, opened 2020-10-27
Add activity blocks to Metricity data endpoint

jb3:
Voice gate has been a success (in general)! :tada:

One downside we have observed is users spamming in order to meet the 50 messages threshold, this is dissatisfactory and will not be tolerated!

This PR introduces a new response in the metricity data endpoint to show how many 10 minute intervals the requested user has been active in. This means how many blocks of 10 minutes has the user sent at least one message in.

A threshold for this has not been decided on for this yet, but this will allow us to mandate that the user gets their messages over a period of time.

If we wanted we could use this instead of a message constraint and say that the user must sent at least one (or any number of) messages per 10 minutes for 6 ten minute blocks.

### DIFF
--- a/pydis_site/apps/api/models/bot/metricity.py
+++ b/pydis_site/apps/api/models/bot/metricity.py
@@ -54,11 +54,7 @@ class Metricity:
               COUNT(*)
              FROM (
                SELECT
-                 to_timestamp(
-                    floor((extract('epoch' from created_at) / %d ))
-                    * %d
-                 )
-                 AT TIME ZONE 'UTC' AS interval
+                 (floor((extract('epoch' from created_at) / %d )) * %d) AS interval
                FROM messages
                WHERE
                  author_id='%s'
--- a/pydis_site/apps/api/models/bot/metricity.py
+++ b/pydis_site/apps/api/models/bot/metricity.py
@@ -1,5 +1,7 @@
 from django.db import connections
 
+BLOCK_INTERVAL = 10 * 60  # 10 minute blocks
+
 
 class NotFound(Exception):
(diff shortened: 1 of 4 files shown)


MarkKoz (changes requested):
I think this file needs to be updated for the sake of local testing, right? https://github.com/python-discord/site/blob/master/postgres/init.sql

thomaspet:
Don't know enough about this to give anything more than a code-review

MarkKoz (approved):
I'm only doing a code review but it looks correct. I'm basing that on the assumption that all the other metricity code in master works since the code you added is following the established patterns.

pydis_site/apps/api/models/bot/metricity.py
```diff
@@ -43,3 +45,28 @@ def total_messages(self, user_id: str) -> int:
             raise NotFound()
 
         return values[0]
+
+    def total_message_blocks(self, user_id: str) -> int:
+        """Query number of 10 minute blocks the user has been active during."""
```
lemonsaurus: Don't end a sentence with a preposition - especially not the preposition "during", that's just weird, dude.

lemonsaurus: Also, maybe you can describe some of the context for this? We should document some context _somewhere_, and I think this is the best option we have.

lemonsaurus (approved):
I think the implementation looks fine here, but you should document the intent a little better.

l33tlinuxh4x0r (approved):
Looks good to me, except for the indents. Mentioned in discord and it's fixed now.

PR #426, opened 2020-11-23
Restructure Build & Deploy sequence in CI

SebastiaanZ:
I've restructured the build & deploy sequence in our GitHub Actions CI workflow:

- There's now a separate workflow to deploy to our cluster

- The Build workflow has an additional `if` check to ensure we're only starting the build & deploy sequences for pushes to master

- We now checkout the python-discord/kubernetes repo to fetch the deployment.yaml file located in there


MarkKoz (approved):
The file name would likely be fine as just `deploy.yml` unless you foresee additional workflows for other deployments. Even in that case, a name with "deploy" as the prefix would be better for sorting purposes, like `deploy-kubernetes.yml` or just `deploy-k8s.yml`.

.github/workflows/kubernetes-deployment.yaml
```diff
@@ -0,0 +1,45 @@
...
+          echo "::set-output name=tag::$tag"
+
+      # Check out the private Kubernetes repository for the
+      # deployment.yaml file using a GitHub Personal Access
+      # Token to get access.
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          repository: python-discord/kubernetes
+          token: ${{ secrets.REPO_TOKEN }}
```
MarkKoz: Is the default token inadequately scopes to access private repos? By the way, don't forget to make sure this secret is actually set.

SebastiaanZ: Yes, it's inadequate; it's scoped to the current repository. The secret is currently set at an organization level.

SebastiaanZ:
> The file name would likely be fine as just `deploy.yml` unless you foresee additional workflows for other deployments. Even in that case, a name with "deploy" as the prefix would be better for sorting purposes, like `deploy-kubernetes.yml` or just `deploy-k8s.yml`.

Ah, yeah, I meant to change that back. Thanks for catching it.

PyCharm's validator picks ups `deploy.yaml` as a special kind of YAML file (haven't found out which one yet) and starts screaming that it doesn't follow the schema. I temporarily renamed it, but then did not notice it anymore in the change set.

Den4200 (approved):
This looks good to me.

PR #431, opened 2020-12-05
Set max-requests-jitter parameter to gunicorn

jb3:
Set the max requests parameter for gunicorn to restart our workers between 1 and 1000 requests to help control the damage of memory leaks. This comes after we observed over 1GB RAM usage from our site container.

A bonus feature of this PR is that it integrates the statsd client inside gunicorn to report to our graphite server so we should be able to measure metric data.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -163,7 +163,10 @@ class SiteManager:
             "-b", "0.0.0.0:8000",
             "pydis_site.wsgi:application",
             "--threads", "8",
-            "-w", "4"
+            "-w", "4",
+            "--max-requests-jitter", "1000",
+            "--statsd_host", "graphite.default.svc.cluster.local:8125",
+            "--statsd_prefix", "site",
         ]
 
         # Run gunicorn for the production server.
--- a/manage.py
+++ b/manage.py
@@ -164,7 +164,8 @@ class SiteManager:
             "pydis_site.wsgi:application",
             "--threads", "8",
             "-w", "4",
-            "--max-requests-jitter", "1000",
+            "--max-requests", "1000",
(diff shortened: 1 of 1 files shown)


jerbob (approved):

manage.py
```diff
@@ -163,7 +163,10 @@ def run_server(self) -> None:
             "-b", "0.0.0.0:8000",
             "pydis_site.wsgi:application",
             "--threads", "8",
-            "-w", "4"
+            "-w", "4",
+            "--max-requests-jitter", "1000",
```
SebastiaanZ: The jitter value is only applied to the `-max-request` option if `max-requests` is greater than `0`. It's meant to jitter your max-requests value and I don't think it works if you only set the jitter value. From the gunicorn source code:

```py
        if cfg.max_requests > 0:
            jitter = randint(0, cfg.max_requests_jitter)
            self.max_requests = cfg.max_requests + jitter
        else:
            self.max_requests = sys.maxsize
```

(Note: this is applied to each worker when it's initiated, so the `randint` really does jitter; it's not a global randint.)


[The author of this article](https://adamj.eu/tech/2019/09/19/working-around-memory-leaks-in-your-django-app/) states that they used a max-request of `1000` and a jitter of `50`:

```suggestion
            "--max-requests", "1000",
            "--max-requests-jitter", "50",
```

SebastiaanZ (changes requested):

manage.py
jb3: Ah! That makes more sense, misunderstanding of the docs on my behalf.

SebastiaanZ (approved):

PR #432, opened 2020-12-05
Include DestroyModelMixin to infractions view for DELETE method

### LINKED ISSUE #430: Support deleting infractions
The bot tries to delete infractions that fail to apply on the Discord side of things. This is cause it sends the infraction to our API before applying it on Discord. However, our API doesn't actually support the delete method so it returns a 405.

ks129:
Closes #430 

Added this mixin and documented this in doctoring.

### DIFF
--- a/pydis_site/apps/api/tests/test_infractions.py
+++ b/pydis_site/apps/api/tests/test_infractions.py
@@ -512,6 +512,36 @@ class CreationTests(APISubdomainTestCase):
             )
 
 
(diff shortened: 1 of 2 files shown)


MarkKoz (changes requested):
Needs some tests written.

MarkKoz (approved):

Den4200 (approved):
Looks good, thanks!

PR #435, opened 2020-12-13
Hotfix: Add some sanity to frontpage projects.

lemonsaurus:
If no API data is available, the front page shouldn't crash and burn.

This commit will ensure the front page will just render without the projects it can't get API data for instead of crashing the whole thing.

image


Akarys42 (approved):
The fix looks good to me and works fine locally.

PR #436, opened 2020-12-13
Stale RepositoryMetadata cleanup

lemonsaurus:
We don't want to keep lots of old metadata for repos we no longer need to track on the frontpage. That can lead to this weird side-effect:
image

This happens because we never removed the old RepositoryMetadata from the database after changing the repos we wanted to appear on the front page. After the recent refactor, we're now displaying all RepositoryMetadata objects when we fetch from cache, and then all of these would then be shown on the front page.


ks129 (approved):

pydis_site/apps/home/views/home.py
```diff
@@ -29,6 +29,12 @@ class HomeView(View):
         "python-discord/django-simple-bulma",
     ]
 
+    def __init__(self):
+        """Clean up stale RepositoryMetadata."""
+        for cached_repo in RepositoryMetadata.objects.all():
+            if cached_repo.repo_name not in self.repos:
+                cached_repo.delete()
```
Den4200: I think it would be much more efficient to filter and delete the repositories in SQL, using Django's ORM:
```suggestion
        RepositoryMetadata.objects.exclude(repo_name__in=self.repos).delete()
```

Den4200 (changes requested):
One small thing!

pydis_site/apps/home/views/home.py
lemonsaurus: Sure, I don't mind this. The efficiency is pointless here because there's only ever like 6 rows in this table, but I think I like your code better and it's fewer lines.

Den4200 (approved):
Looks good!

PR #443, opened 2021-01-26
Apply LimitOffsetPagination in GET Infraction with default Page Size 100

### LINKED ISSUE #442: Limit results when querying infractions
## Abstract

We had some issues that caused an outage, where someone would search the infractions for single characters, which resulted in an OOM after some time. We need to limit the number of infractions returned by the API

## Proposed Implementation

Add a [`LimitOffsetPagination`](https://www.django-rest-framework.org/api-guide/pagination/#limitoffsetpagination) to the viewset in `site/pydis_site/apps/api/viewsets/bot/infraction.py`, limiting the output to 50 infractions. 

Note that we need to make sure that Django still won't try to fetch all the infractions in order to sort them, or something similar.

kunal097:
Add Default Pagination class **LimitOffsetPagination** with Page Size **100**  in GET Infraction


pydis_site/settings.py
```diff
@@ -230,7 +230,9 @@
     'DEFAULT_PERMISSION_CLASSES': (
         'rest_framework.permissions.DjangoModelPermissions',
     ),
-    'TEST_REQUEST_DEFAULT_FORMAT': 'json'
+    'TEST_REQUEST_DEFAULT_FORMAT': 'json',
+    'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.LimitOffsetPagination',
```
Akarys42: We don’t want to use pagination everywhere, only in the infraction route. Can you please revert this change? This also cause some tests to fail.

pydis_site/apps/api/viewsets/bot/infraction.py
```diff
@@ -3,20 +3,15 @@
 from rest_framework.decorators import action
 from rest_framework.exceptions import ValidationError
 from rest_framework.filters import OrderingFilter, SearchFilter
-from rest_framework.mixins import (
-    CreateModelMixin,
-    DestroyModelMixin,
-    ListModelMixin,
-    RetrieveModelMixin
-)
+from rest_framework.mixins import (CreateModelMixin, DestroyModelMixin,
```
Akarys42: This doesn’t conform with the style used in the repository, can you please revert it?

pydis_site/apps/api/viewsets/bot/infraction.py
```diff
@@ -3,20 +3,15 @@
...
+from rest_framework.pagination import LimitOffsetPagination
 from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 
 from pydis_site.apps.api.models.bot.infraction import Infraction
-from pydis_site.apps.api.serializers import (
-    ExpandedInfractionSerializer,
-    InfractionSerializer
-)
+from pydis_site.apps.api.serializers import (ExpandedInfractionSerializer,
```
Akarys42: Same as above

Akarys42 (changes requested):
Hello, thank you for your contribution! There are a couple of things that needs to be reverted.

pydis_site/settings.py
kunal097: Sure, I'll revert this one

pydis_site/apps/api/viewsets/bot/infraction.py
kunal097: Sure, I'll update that part also.

kunal097 (commented):
Changes have been done as discussed. Please review the changes.

pydis_site/apps/api/viewsets/bot/infraction.py
```diff
@@ -133,10 +134,27 @@ class InfractionViewSet(
...
+        DRF method for listing Infraction entries.
+
+        Called by the Django Rest Framework in response to the corresponding HTTP request.
+        """
+        queryset = self.filter_queryset(self.get_queryset())
+        page = self.paginate_queryset(queryset)
+        if page:
+            serializer = self.get_serializer(page, many=True)
+            result = self.get_paginated_response(serializer.data)
+            return Response(result.data.get('results'))
```
MarkKoz: If I understand correctly, you overwrote the `list` method just to customise the response. Specifically, you want to only return the raw data from the paginated response and not any of the pagination metadata. If that's the case, a better way is to [create a custom paginator](https://www.django-rest-framework.org/api-guide/pagination/#custom-pagination-styles) which overwrites `get_paginated_response`. This approach achieves the same result with less redundant code since it can rely on the `list` implementation provided by the mixin.

In any case, why is this being done? Is it to maintain the old API response format? How is the client supposed to paginate without knowing the final page?

MarkKoz (changes requested):

pydis_site/apps/api/viewsets/bot/infraction.py
kunal097: Yes, the sole purpose to override this method is to maintain the old response.
If the response has been changed then any other services which is using the older response will break. So before updating the response we'll need to update those services as well.

MarkKoz: They'd have to be updated anyway unless it is decided that providing more than the 100 most recent infractions for any query is unnecessary.

kunal097: So do I need to send updated response with pagination metadata?

MarkKoz: That is contingent on what moderators decide they need. You'll have to wait for a response from them.

kunal097: Ok

kunal097: Any update on this?

HassanAbouelela: I believe it was decided a single page is fine, and we can maintain the old result format. Mark's other comment on custom paginators is still unresolved though.

kunal097: Cool then I'll work on the custom paginator.

MarkKoz: Please update the documentation on the endpoint too. It should mention that the results are paginated but only the actual data is returned.

kunal097:
Please review the required changes.

pydis_site/apps/api/viewsets/bot/infraction.py
kunal097: Changes have been done and pushed, take a look at the new commit.

pydis_site/apps/api/viewsets/bot/pagination.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework.pagination import LimitOffsetPagination
+from rest_framework.response import Response
+
+
+class LimitSetPagination(LimitOffsetPagination):
+    """Extend LimitOffsetPagination."""
```
MarkKoz: The documentation doesn't have to be extremely detailed, but what you currently have is too vague. It doesn't really provide any information the reader doesn't already know.

pydis_site/apps/api/viewsets/bot/pagination.py
```diff
@@ -0,0 +1,12 @@
...
+
+    default_limit = 100
+
+    def get_paginated_response(self, data: list) -> Response:
+        """Override default response."""
```
MarkKoz: Too vague; same as above.

pydis_site/apps/api/viewsets/bot/pagination.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework.pagination import LimitOffsetPagination
+from rest_framework.response import Response
+
+
+class LimitSetPagination(LimitOffsetPagination):
```
MarkKoz: This name is unclear to me. What's the reasoning behind this name?

pydis_site/apps/api/viewsets/bot/pagination.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework.pagination import LimitOffsetPagination
```
MarkKoz: A better place for this file would be `pydis_site/apps/api/pagination.py`. It shouldn't be under `viewsets` since it's not a viewset.

pydis_site/apps/api/viewsets/bot/infraction.py
```diff
@@ -46,6 +49,7 @@ class InfractionViewSet(
     Invalid query parameters are ignored.
 
     #### Response format
+    - Response are paginated but only the actual data is returned
```
MarkKoz: Let's clarify what "actual data" means.
```suggestion
    The response is paginated but only the actual data is returned.
    No pagination-specific metadata is included.

```

pydis_site/apps/api/viewsets/bot/pagination.py
```diff
@@ -0,0 +1,12 @@
...
+from rest_framework.response import Response
+
+
+class LimitSetPagination(LimitOffsetPagination):
+    """Extend LimitOffsetPagination."""
+
+    default_limit = 100
+
+    def get_paginated_response(self, data: list) -> Response:
```
MarkKoz: It's technically not limited to list data. The type annotation should be `typing.Any` since it can be a wide range of different types.

pydis_site/apps/api/viewsets/bot/infraction.py
```diff
@@ -38,6 +39,8 @@ class InfractionViewSet(
     - **active** `bool`: whether the infraction is still active
     - **actor__id** `int`: snowflake of the user which applied the infraction
     - **hidden** `bool`: whether the infraction is a shadow infraction
+    - **limit** `int`: default limit is 100
+    - **offset** `int`: default is 0
```
MarkKoz: The documentation should state what these actually do in addition to their defaults.

MarkKoz (changes requested):

pydis_site/apps/api/viewsets/bot/infraction.py
kunal097: I tried to resolve all the required changes. Please take a look.

pydis_site/apps/api/pagination.py
```diff
@@ -0,0 +1,49 @@
...
+                "reason": null,
+                "hidden": false
+            }]
+
+    """
+
+    default_limit = 100
+
+    def get_paginated_response(self, data: typing.Any) -> Response:
+        """Override to skip metadata i.e. `count`, `next` and `previous`."""
```
MarkKoz:
```suggestion
        """Override to skip metadata i.e. `count`, `next`, and `previous`."""
```

pydis_site/apps/api/viewsets/bot/infraction.py
```diff
@@ -46,6 +49,8 @@ class InfractionViewSet(
     Invalid query parameters are ignored.
 
     #### Response format
+    - Response are paginated but the result is returned without any pagination metadata.
+      Below is the returned format.
```
MarkKoz:
```suggestion
    Response is paginated but the result is returned without any pagination metadata.
```

pydis_site/apps/api/viewsets/bot/infraction.py
```diff
@@ -38,6 +39,8 @@ class InfractionViewSet(
     - **active** `bool`: whether the infraction is still active
     - **actor__id** `int`: snowflake of the user which applied the infraction
     - **hidden** `bool`: whether the infraction is a shadow infraction
+    - **limit** `int`: number of results return per page. default limit is 100
+    - **offset** `int`: the initial index from which to return the results, default is 0
```
MarkKoz:
```suggestion
    - **limit** `int`: number of results return per page (default 100)
    - **offset** `int`: the initial index from which to return the results (default 0)
```

pydis_site/apps/api/pagination.py
```diff
@@ -0,0 +1,49 @@
...
+                "inserted_at": "2021-01-26T21:13:35.477879Z",
+                "expires_at": null,
+                "active": false,
+                "user": 1,
+                "actor": 2,
+                "type": "warning",
+                "reason": null,
+                "hidden": false
+            }]
+
```
MarkKoz, line 30: Remove blank line at end.
```suggestion
```

pydis_site/apps/api/pagination.py
```diff
@@ -0,0 +1,49 @@
...
+                        "expires_at": null,
+                        "active": false,
+                        "user": 1,
+                        "actor": 2,
+                        "type": "warning",
+                        "reason": null,
+                        "hidden": false
+                    }]
+            }
+    ## Required response
```
MarkKoz: Add blank line between responses
```suggestion

    ## Required response
```

pydis_site/apps/api/pagination.py
```diff
@@ -0,0 +1,49 @@
...
+                "id": 6,
+                "inserted_at": "2021-01-26T21:13:35.477879Z",
+                "expires_at": null,
+                "active": false,
+                "user": 1,
+                "actor": 2,
+                "type": "warning",
+                "reason": null,
+                "hidden": false
+            }]
```
MarkKoz: For consistency, ideally these would be formatted like the docs for viewsets. They are formatted as if they were copied from a Python REPL. They are also not indented consistently.

pydis_site/apps/api/pagination.py
```diff
@@ -0,0 +1,49 @@
+import typing
+
+from rest_framework.pagination import LimitOffsetPagination
+from rest_framework.response import Response
+
+
+class LimitOffsetPaginationExtended(LimitOffsetPagination):
```
MarkKoz, line 7: Not the most descriptive name, but I can't think of anything much better, so we can roll with it.

MarkKoz (approved):
This works fine and I will approve it. However, I've left some more minor suggestions to clean up the documentation and make it more consistent.

Akarys42 (approved):
Thanks a lot for the PR, this works very nicely!

Akarys42 (commented):
Refresh

Akarys42 (approved):

Qwerty-133 (approved):
Approving in place of @Akarys42

PR #446, opened 2021-02-20
Allow multiple nominations per users

### LINKED ISSUE #445: Allow many nominations per user
Currently, the viewset [`/bot/nominations`](https://github.com/python-discord/site/blob/master/pydis_site/apps/api/viewsets/bot/nomination.py) has a validator making sure that a user can't have two active nominations at the same time.

https://github.com/python-discord/site/blob/bceaef8a1921a4013fb525c3b68f41810bec5c14/pydis_site/apps/api/viewsets/bot/nomination.py#L165-L167

With https://github.com/python-discord/bot/issues/1425 we now want to allow this. Those lines should be removed, and [this](https://github.com/python-discord/site/blob/master/pydis_site/apps/api/tests/test_nominations.py#L40) test has to be updated.

You can use [this patch](https://paste.pythondiscord.com/cotezesuye.patch?noredirect) to get started.

ks129:
Removes check that doesn't allow users have more than 1 active nomination. Updated docs and tests to match with this change. Closes #445

### DIFF
--- a/pydis_site/apps/api/tests/test_nominations.py
+++ b/pydis_site/apps/api/tests/test_nominations.py
@@ -37,7 +37,7 @@ class CreationTests(APISubdomainTestCase):
         self.assertEqual(nomination.reason, data['reason'])
         self.assertEqual(nomination.active, True)
 
-    def test_returns_400_on_second_active_nomination(self):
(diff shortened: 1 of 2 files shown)


ks129:
This will be resolved during the bigger refactor of nomination storage.

PR #447, opened 2021-02-22
Rework nominations

### LINKED ISSUE #445: Allow many nominations per user

ks129:
**Note! This HAVE to be merged EXACTLY the same time as bot PR**

# Changes
Instead of one nominations table, there are now 2 tables: Nominations and nomination entries. The nominations table got one new field `reviewed` (boolean). Here are schemas of these tables:
```
Table: nomination

active (boolean)
user (User)
inserted_at (datetime)
end_reason (string)
ended_at (datetime)
reviewed (boolean)
```
```
Table: nomination entries

nomination (Nomination)
actor (User)
reason (string)
inserted_at (datetime)
```

Now is also allowed multiple nominations per-user and they are created as nomination entries. One user can create only 1 entry for each nomination.

Also updated tests to match with all these changes.

# What to test?
- Is the migration script properly moving data from old schema to new.
- Is GET routes working as expected?
- Is the POST route creating a new nomination entry if nomination already exists?
- Is the PATCH route `reviewed` field setting works?
- Is the PATCH patching behaviors correct?
- Is tests covering everything (there is one place where I had to ignore it)?

Closes #445


pydis_site/apps/api/models/bot/nomination.py
```diff
@@ -52,3 +45,33 @@ class Meta:
...
+    actor = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        help_text="The staff member that nominated this user.",
+        related_name='nomination_set'
+    )
+    reason = models.TextField(
+        help_text="Why the actor nominated this user.",
+        null=True,
+        blank=True
```
SebastiaanZ: Do we want two different options to represent an empty reason, `""` and `null`? Could you explain why? The Django docs recommend you to not use `null=True` for character-based fields, as they already have a "logical" null value being an empty string.

pydis_site/apps/api/models/bot/nomination.py
```diff
@@ -52,3 +45,33 @@ class Meta:
...
+    )
+    inserted_at = models.DateTimeField(
+        auto_now_add=True,
+        help_text="The creation date of this nomination entry."
+    )
+
+    class Meta:
+        """Meta options for NominationEntry model."""
+
+        verbose_name_plural = "nomination entries"
```
SebastiaanZ, line 78: It could be useful to specify a default [`ordering`](https://docs.djangoproject.com/en/3.1/ref/models/options/#ordering) here that uses the `inserted_at` date, either ascending or descending. This makes it so that nominations can be displayed in chronological order (or reverse chronological order) without having to specify anything anywhere else.

pydis_site/apps/api/serializers.py
```diff
@@ -338,6 +339,21 @@ def create(self, validated_data: dict) -> User:
             raise ValidationError({"id": ["User with ID already present."]})
 
 
+class NominationEntrySerializer(ModelSerializer):
+    """A class providing (de-)serialization of `NominationEntry` instances."""
+
+    nomination = PrimaryKeyRelatedField(
+        queryset=Nomination.objects.all(),
+        write_only=True
+    )
```
SebastiaanZ, line 351: Could you explain what this does? I'd expected to see a forward `PrimaryKeyRelatedField` serializer relationship in which a `Nomination` has a list with all the `NominationEntries` that belong to that specific nomination, like the `Album` model in the example in the docs having a `many=True` `PrimaryKeyRelatedField` relationship to the individual tracks on the album.

Don't we have something very similar? A single Nomination with several NominationEntries like an Album has many individual Tracks?

Ideally, I'd like to see reasoning for the choice you've made in your commit message, but that may now be hard to add after the fact, as it would rewrite history.

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -148,10 +160,44 @@ class NominationViewSet(CreateModelMixin, RetrieveModelMixin, ListModelMixin, Ge
...
+        """
+        queryset = self.filter_queryset(self.get_queryset())
+        data = NominationSerializer(queryset, many=True).data
+
+        for i, nomination in enumerate(data):
+            entries = NominationEntrySerializer(
+                NominationEntry.objects.filter(nomination_id=nomination["id"]),
+                many=True
+            ).data
+            data[i]["entries"] = entries
```
SebastiaanZ: Wouldn't this be solved by adding a `serializers.PrimaryKeyRelatedField` to the `NominationSerializer`? That saves us from doing a manual `for`-loop here. In fact, it probably means we wouldn't have to overwrite the list method at all, as we don't have need for custom functionality that way.

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -148,10 +160,44 @@ class NominationViewSet(CreateModelMixin, RetrieveModelMixin, ListModelMixin, Ge
...
+        """
+        nomination = self.get_object()
+
+        data = NominationSerializer(nomination).data
+        data["entries"] = NominationEntrySerializer(
+            NominationEntry.objects.filter(nomination_id=nomination.id),
+            many=True
+        ).data
+
+        return Response(data)
```
SebastiaanZ: My guess is that we don't need this if we make sure to serialize a `Nomination` with its related `NominationEntry` information in the serializer part of the framework. Saves us from overwriting a method.

SebastiaanZ (changes requested):
I ran out of time for now, but I'm a few commits in. I've just gone over the code as of now and I wanted to give my feedback early, rather than late, as I'm not sure when I'll have time again. 

Things I haven't reviewed yet:
- Grammar
- Functional test

As a general comment: None of the commit messages I've read have a meaningful commit message body and some of the titles are a bit too long. Ideally, a commit message should explain the what and why of the change you're committing to the history instead of having just a single summary line.

Some of these commits are fairly sizeable and you've made a number of choices within those commits; try to add the reasoning behind those choices into the commit message body so that reasoning becomes part of the git history of the repository.

These are few links @lemonsaurus typically shares when talking about commit messages:

https://chris.beams.io/posts/git-commit/
https://thoughtbot.com/blog/5-useful-tips-for-a-better-commit-message
http://ablogaboutcode.com/2011/03/23/proper-git-commit-messages-and-an-elegant-git-history
https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

A more extreme example: https://dhwthompson.com/2019/my-favourite-git-commit

This page is also really nice: https://mtlynch.io/code-review-love/

pydis_site/apps/api/serializers.py
ks129: I don't want to include nomination ID in the read result, but if I don't define queryset, I get `AssertionError: Relational field must provide a queryset argument, override get_queryset, or set read_only=True.`.

HassanAbouelela:
I've looked over the new schema, and from a functional standpoint, it seems good.

Here are the old vs new schemas if anyone wants to compare (you can play around with it [here](https://dbdiagram.io/d/603ca005fcdcb6230b21f070)):
| Old | image |
| ---- | ----- |
| New  | image |

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -75,8 +86,9 @@ class NominationViewSet(CreateModelMixin, RetrieveModelMixin, ListModelMixin, Ge
     Create a new, active nomination returns the created nominations.
     The `user`, `reason` and `actor` fields are required and the `user`
     and `actor` need to know by the site. Providing other valid fields
-    is not allowed and invalid fields are ignored. A `user` is only
-    allowed one active nomination at a time.
+    is not allowed and invalid fields are ignored. If `user` already have
```
lemonsaurus:
```suggestion
    is not allowed and invalid fields are ignored. If `user` already has an
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -75,8 +86,9 @@ class NominationViewSet(CreateModelMixin, RetrieveModelMixin, ListModelMixin, Ge
...
+    active nomination, new nomination entry will be created assigned to
```
lemonsaurus:
```suggestion
    active nomination, a new nomination entry will be created and assigned as the
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -221,6 +281,34 @@ def partial_update(self, request: HttpRequest, *args, **kwargs) -> Response:
...
+                )
+
+            entry_filter = NominationEntry.objects.filter(
+                nomination_id=instance.id,
+                actor__id=request.data['actor']
+            )
+
+            if not entry_filter.exists():
+                raise ValidationError(
+                    {'actor': ["Actor don't exist or have not nominated user."]}
```
jb3: Single quotes.
```suggestion
                    {'actor': ['Actor don't exist or have not nominated user.']}
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -102,16 +113,18 @@ class NominationViewSet(CreateModelMixin, RetrieveModelMixin, ListModelMixin, Ge
...
     2. Ending an `active` nomination;
     3. Updating the `end_reason` or `reason` field of an `inactive` nomination.
+    4. Updating `reviewed` field of `active` nomination.
 
     While the response format and status codes are the same for all three operations (see
     below), the request bodies vary depending on the operation. For all operations it holds
     that providing other valid fields is not allowed and invalid fields are ignored.
 
-    ### 1. Updating the `reason` of `active` nomination
+    ### 1. Updating the `reason` of `active` nomination. Actor field is required.
```
lemonsaurus:
```suggestion
    ### 1. Updating the `reason` of `active` nomination. The `actor` field is required.
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -163,19 +187,50 @@ def create(self, request: HttpRequest, *args, **kwargs) -> Response:
...
-        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+        entry_serializer.is_valid(raise_exception=True)
+
+        # Don't allow user creating many nomination entries for one nomination
+        if NominationEntry.objects.filter(
+                nomination_id=nomination_filter[0].id,
+                actor__id=entry_serializer.validated_data["actor"].id
+        ).exists():
+            raise ValidationError(
+                {'actor': ['This actor have already created nomination entry for this nomination.']}
```
jb3: Grammar
```suggestion
                {'actor': ['This actor has already endorsed this nomination.']}
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -163,19 +187,50 @@ def create(self, request: HttpRequest, *args, **kwargs) -> Response:
...
+                )
+            )
+            serializer.is_valid(raise_exception=True)
+            nomination = Nomination.objects.create(**serializer.validated_data)
 
-        serializer = self.get_serializer(
-            data=ChainMap(
-                request.data,
-                {"active": True}
+            # Serializer truncate unnecessary data away
```
lemonsaurus:
```suggestion
            # The serializer will truncate and get rid of excessive data
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -213,6 +268,11 @@ def partial_update(self, request: HttpRequest, *args, **kwargs) -> Response:
                     {'end_reason': ['This field is required when ending a nomination.']}
                 )
 
+            if 'reviewed' in request.data:
+                raise ValidationError(
+                    {'reviewed': ['This field cannot be set same time than ending nomination.']}
```
jb3: I don't understand what this means, but I think this is an improvement. (It may be the opposite).
```suggestion
                    {'reviewed': ['This field cannot be set while you are ending a nomination.']}
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -221,6 +281,34 @@ def partial_update(self, request: HttpRequest, *args, **kwargs) -> Response:
...
+            # 4. We're setting nomination reviewed
+            if not instance.active:
+                raise ValidationError(
+                    {'reviewed': ['This field cannot be set if nomination is inactive.']}
+                )
+
+        if 'reason' in request.data:
+            if 'actor' not in request.data:
+                raise ValidationError(
+                    {'actor': ['This field is required when editing reason.']}
```
jb3:
```suggestion
                    {'actor': ['This field is required when editing the reason.']}
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -221,6 +281,34 @@ def partial_update(self, request: HttpRequest, *args, **kwargs) -> Response:
                 {'active': ['This field can only be used to end a nomination']}
             )
 
+        # This is actually covered, but for some reason coverage don't think so.
+        elif 'reviewed' in request.data:  # pragma: no cover
+            # 4. We're setting nomination reviewed
+            if not instance.active:
+                raise ValidationError(
+                    {'reviewed': ['This field cannot be set if nomination is inactive.']}
```
jb3:
```suggestion
                    {'reviewed': ['This field cannot be set if the nomination is inactive.']}
```

lemonsaurus (approved):
I'm mostly okay with this PR. There are grammar issues, but it looks like it does what it needs to do.

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -221,6 +281,34 @@ def partial_update(self, request: HttpRequest, *args, **kwargs) -> Response:
                 {'active': ['This field can only be used to end a nomination']}
             )
 
+        # This is actually covered, but for some reason coverage don't think so.
+        elif 'reviewed' in request.data:  # pragma: no cover
+            # 4. We're setting nomination reviewed
```
jb3:
```suggestion
            # 4. We are altering the reviewed state of the nomination.
```

jb3 (approved):
Mainly grammar + code style fixes, but otherwise looks good.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: If this is going to be changed, it's needs to be changed in the test as well

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: Make sure to change the test as well.

pydis_site/apps/api/viewsets/bot/nomination.py
SebastiaanZ: make sure to adjust the test as well

SebastiaanZ (approved):

pydis_site/apps/api/models/bot/nomination.py
```diff
@@ -42,6 +31,10 @@ class Nomination(ModelReprMixin, models.Model):
         help_text="When the nomination was ended.",
         null=True
     )
+    reviewed = models.BooleanField(
+        default=False,
+        help_text="Whether voting message have been made."
```
mbaruh: Another small correction
```suggestion
        help_text="Whether a review was made."
```

mbaruh (changes requested):

pydis_site/apps/api/models/bot/nomination.py
```diff
@@ -52,3 +45,38 @@ class Meta:
         """Set the ordering of nominations to most recent first."""
 
         ordering = ("-inserted_at",)
+
+
+class NominationEntry(ModelReprMixin, models.Model):
+    """A nomination entry created by single staff."""
```
mbaruh:
```suggestion
    """A nomination entry created by a single staff member."""
```

pydis_site/apps/api/models/bot/nomination.py
```diff
@@ -52,3 +45,38 @@ class Meta:
...
+
+    nomination = models.ForeignKey(
+        Nomination,
+        on_delete=models.CASCADE,
+        help_text="Nomination to what this entry belongs.",
```
mbaruh:
```suggestion
        help_text="The nomination this entry belongs to.",
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -75,8 +86,9 @@ class NominationViewSet(CreateModelMixin, RetrieveModelMixin, ListModelMixin, Ge
     Create a new, active nomination returns the created nominations.
     The `user`, `reason` and `actor` fields are required and the `user`
     and `actor` need to know by the site. Providing other valid fields
-    is not allowed and invalid fields are ignored. A `user` is only
-    allowed one active nomination at a time.
+    is not allowed and invalid fields are ignored. If `user` already has an
+    active nomination, a new nomination entry will be created and assigned as the
```
mbaruh:
```suggestion
    active nomination, a new nomination entry will be created and assigned to the
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -163,19 +187,50 @@ def create(self, request: HttpRequest, *args, **kwargs) -> Response:
...
+        entry_serializer = NominationEntrySerializer(
+            data=ChainMap(request.data, {"nomination": nomination_filter[0].id})
         )
-        serializer.is_valid(raise_exception=True)
-        self.perform_create(serializer)
-        headers = self.get_success_headers(serializer.data)
-        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+        entry_serializer.is_valid(raise_exception=True)
+
+        # Don't allow user creating many nomination entries for one nomination
```
mbaruh:
```suggestion
        # Don't allow a user to create many nomination entries in a single nomination
```

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -221,6 +281,34 @@ def partial_update(self, request: HttpRequest, *args, **kwargs) -> Response:
...
+                )
+
+            entry_filter = NominationEntry.objects.filter(
+                nomination_id=instance.id,
+                actor__id=request.data['actor']
+            )
+
+            if not entry_filter.exists():
+                raise ValidationError(
+                    {'actor': ["Actor don't exist or have not nominated user."]}
```
mbaruh:
```suggestion
                    {'actor': ["The actor doesn't exist or has not nominated the user."]}
```

mbaruh (changes requested):
A few more

mbaruh (approved):

PR #464, opened 2021-03-25
Use real names in the timeline

lemonsaurus:
Using nicknames like Ves Zappa and Lemon doesn't always feel appropriate, so I've replaced some of those references with real names.

This PR also updates references to SeasonalBot to Sir Lancebot.

### DIFF
--- a/pydis_site/templates/home/timeline.html
+++ b/pydis_site/templates/home/timeline.html
@@ -90,7 +90,7 @@
           <h2>First code jam with the theme “snakes”</h2>
           <p class="color-contrast-medium">Our very first Code Jam attracts a handful of users who work in random
             teams of 2. We ask our participants to write a snake-themed Discord bot. Most of the code written
-            for this jam still lives on in SeasonalBot, and you can play with it by using the
+            for this jam still lives on in Sir Lancebot, and you can play with it by using the
             <code>.snakes</code> command. For more information on this event, see <a
                 href="https://pythondiscord.com/pages/code-jams/code-jam-1-snakes-bot/">the event page</a></p>
 
@@ -178,7 +178,7 @@
         </div>
 
         <div class="cd-timeline__content text-component">
-          <h2>First Hacktoberfest PyDis event; @SeasonalBot is created</h2>
+          <h2>First Hacktoberfest PyDis event; @Sir Lancebot is created</h2>
           <p class="color-contrast-medium">We create a second bot for our community and fill it up with simple,
             fun and relatively easy issues. The idea is to create an approachable arena for our members to cut
             their open-source teeth on, and to provide lots of help and hand-holding for those who get stuck.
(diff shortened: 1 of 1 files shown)


pydis_site/templates/home/timeline.html
```diff
@@ -623,8 +622,8 @@ <h2>Lemon on Talk Python To Me</h2>
         </div>
 
         <div class="cd-timeline__content text-component">
-          <h2>Lemon on Teaching Python</h2>
-          <p class="color-contrast-medium">Lemon goes on the Teaching Python podcast to discuss how the pandemic has
+          <h2>We're on the Teaching Python podcast!</h2>
+          <p class="color-contrast-medium">Lemon joins Sean and Kelly on the Teaching Python podcast to discuss how the pandemic has
```
Akarys42: Still lemon on that line?

lemonsaurus: I did this on purpose but I don't know, I guess I'm open to suggestions. Leon looks so weird to me. Nobody calls me that. But yeah maybe I should just let go and change this too.

lemonsaurus: Fixed, and squashed fixes into initial commit.

Akarys42: Sweet! I think Leon does make sense here, since lemon doesn’t appear anywhere on the timeline. That said, if you prefer lemon I really don’t mind haha

Akarys42 (approved):

pydis_site/templates/home/timeline.html
```diff
@@ -19,8 +19,8 @@
 
         <div class="cd-timeline__content text-component">
           <h2>Python Discord is created</h2>
-          <p class="color-contrast-medium"><strong>joe</strong> becomes one of the owners around 3 days after it
-            is created, and <strong>lemon</strong> joins the owner team later in the year, when the community
+          <p class="color-contrast-medium"><strong>Joe Banks</strong> becomes one of the owners around 3 days after it
+            is created, and <strong>Leon Sandøy</strong> joins the owner team later in the year, when the community
```
gustavwilliam:
```suggestion
            is created, and <strong>Leon Sandøy</strong> (lemon) joins the owner team later in the year, when the community
```
What do you think about putting the nicknames inside parentheses, so people will be recognized on the server? This would go for all people mentioned.

gustavwilliam (approved):
Looks great! I have one minor comment, but it's fine either way.

pydis_site/templates/home/timeline.html
lemonsaurus: I was actually doing that earlier, I just removed it. I don't know if it's necessary for joe and sebastiaan since their names are their nicknames.

gustavwilliam: Yeah, that's true. I think it would be nice nice to have for you, though.

lemonsaurus: Okay, added! 👍🏼

PR #468, opened 2021-03-29
Dewikification: Content app improvements.

kosayoda:
### Simplifications
- Metadata for categories now use `title` as the key to match page metadata, previously it was `name`.
- Rather than having two separate keys for the Font Awesome icon class and icon, it is now under `icon`.

### Structure enhancements
This PR adds a way to have a path be both a category and a page. This is useful for places like `/contributing`, where https://pythondiscord.com/pages/contributing/ is a rendered page while still being a category for https://pythondiscord.com/pages/contributing/sir-lancebot/ and the like.

### Markdown additions
Certain markdown-to-html additions were made to enable feature parity with the current `wiki`-ed site guides, but without the article writer needing to write HTML.

Specifically, these features are added:
- Image captions
image
- Table of contents
image


kosayoda:
tfw when you do everything except update tests

pydis_site/apps/content/resources/guides/pydis-guides/how-to-contribute-a-page.md
```diff
@@ -22,17 +22,38 @@ As website changes require staff approval, discussing the page content beforehan
...
 ```yml
-name: Category name
+title: Category name
 description: Category description
+icon: fas fa-folder # Optional
 ```
 
 All the markdown files in this folder will then be under this category.
 
+#### Having the Category also be a Page
```
MarkKoz:
```suggestion
#### Having the Category Also Be a Page
```

pydis_site/apps/content/resources/guides/pydis-guides/how-to-contribute-a-page.md
```diff
@@ -22,17 +22,38 @@ As website changes require staff approval, discussing the page content beforehan
...
+In order to make categories a page, place a page inside the category folder **with the same name as the category folder**.
+
+```plaintext
+guides
+├── contributing
+│   ├── _info.yml
+│   ├── contributing.md
+│   └── bot.md
+└── _info.yml
+```
```
MarkKoz: On one hand, it may be more convenient to keep all files related to a category within the same directory. On the other hand, it's less intuitive when considering how normal categories and pages are defined. Have you thought about moving the page out of the directory and instead detecting if the directory contains both a file and subdirectory with the same name (excluding the file extension)?

pydis_site/apps/content/resources/guides/pydis-guides/how-to-contribute-a-page.md
```diff
@@ -59,9 +79,9 @@ Pages, which include guides, articles, and other static content,...
 - **description:** Short, 1-2 line description of the page's content.
 
 ### Optional Fields
-- **icon_class:** Favicon class for the category entry for the page. Default: `fab`
-- **icon:** Favicon for the category entry for the page. Default: `fa-python` <i class="fab fa-python is-black" aria-hidden="true"></i>
+- **icon:** Icon for the category entry for the page. Default: `fab fa-python` <i class="fab fa-python is-black" aria-hidden="true"></i>
 - **relevant_links:** A YAML dictionary containing `text:link` pairs. See the example above.
+- **toc:** A number representing the smallest heading tag to show in the table of contents.
```
MarkKoz, line 85: How to define a table of contents is currently not documented.

MarkKoz (changes requested):

pydis_site/apps/content/resources/guides/pydis-guides/how-to-contribute-a-page.md
```diff
@@ -22,17 +22,38 @@ As website changes require staff approval, discussing the page content beforehan
...
+├── contributing
+│   ├── _info.yml
+│   ├── contributing.md
+│   └── bot.md
+└── _info.yml
+```
+
+In the above example, `www.pythondiscord.com/guides/` will list `Contributing` as a category entry with information from `contributing/_info.yml`.
+
+However, `www.pythondiscord.com/guides/contributing` will render `contributing.md` rather than show the category contents, so *it is the article's responsibility to link to any subpages under the article*.
```
MarkKoz: Have you considered any way to show the direct children, or even all children? Perhaps a dropdown menu somewhere, like near the breadcrumbs.

pydis_site/apps/content/resources/guides/pydis-guides/how-to-contribute-a-page.md
kosayoda: I've thought of that. I personally find putting category pages in the category itself more intuitive.
If `contributing.md` is placed in the `guides` folder, what would be the description shown in the entry for `Contributing` in `..com/guides/`? The current implementation shows the information from `contributing/_info.yml`, while the rendered page has metadata from `contributing.md`.
Would placing `contributing.md` in the `guides` folder then make it confusing as to what information gets displayed where?

MarkKoz: I see it as less intuitive because it breaks the established convention of a page being a file that is a direct child of the directory. Your point about info coming from different pages follows the same behaviour a normal page and category would follow. Really, the only deviation is that clicking the link in the parent category listing will lead to the page rather than to another category listing. It makes sense to me since I understand that when the parent lists it, it treats it as a normal category in every sense except for what is rendered when the link is clicked.

I don't feel strongly about this, but I think it's worth discussing at least.

MarkKoz: Also note that with the current approach, it's impossible for it to contain a child page with the same name as the category. Granted, I don't foresee this being something we'd need to do.

kosayoda: I see your point. The change won't be hard to make, the main addition is just to not list the category page in the listing if there is also a category with the same name in the directory.

pydis_site/apps/content/resources/guides/pydis-guides/how-to-contribute-a-page.md
kosayoda: I have. Since I'm just trying to get feature parity with the current site, I'll open a separate issue for this so the app can move along.

kosayoda:
Just realized there's a logic error in the category page path, will fix.

pydis_site/apps/content/resources/guides/pydis-guides/how-to-contribute-a-page.md
kosayoda: Plot twist: I did it anyways since it's not a big change. The dropdown only shows up on category-page-hybrids and look and function similar to the one shown on the current site when you're logged in.

The differences are that it only shows direct children, since any sub-categories show their children on their page anyway, and that the dropdown is click-toggled and not hover-toggled to be mobile-friendly.

pydis_site/templates/content/dropdown.html
```diff
@@ -0,0 +1,36 @@
...
+            <span>Sub-Articles</span>
+            <span class="icon is-small">
+                <i class="fas fa-angle-down" aria-hidden="true"></i>
+            </span>
+        </a>
+    </div>
+    <div class="dropdown-menu" id="subarticle-menu" role="menu">
+        <div class="dropdown-content">
+            {% for page in subarticles|dictsort:0 %}
+                <a href="{{page}}" class="dropdown-item">{{ page.title|replace_hyphens:" " }}</a>
```
MarkKoz: While this is a simple and fast approach, it is too naïve and capitalises all words. Furthermore, file names may not necessarily be anything close to the titles in the metadata (unless we start saying that they should be).

pydis_site/apps/content/views/page_category.py
```diff
@@ -33,29 +39,48 @@ def get_context_data(self, **kwargs) -> t.Dict[str, t.Any]:
...
+            context.update(self._get_category_context(self.category_path))
+            context["path"] = f"{self.location}/"  # Add trailing slash to simplify template
         else:
             raise Http404
 
+        # Add subarticle information for dropdown menu if the page is also a category
+        if self.page_path.is_file() and self.category_path.is_dir():
+            context["subarticles"] = [
+                path.stem for path in self.category_path.iterdir()
+                if path.suffix != ".yml"
```
MarkKoz: A strictly correct check would be `path.is_dir() or path.suffix == ".md"`.

MarkKoz (changes requested):

pydis_site/templates/content/dropdown.html
kosayoda: Hmm, I didn't think of that. I like the filename freedom, as we want the site URL to be as concise as possible, while having a more descriptive page title.

MarkKoz (approved):

PR #477, opened 2021-04-13
Added Favicon

NeilShah2026:
Added the Py-Dis Logo As A Favicon For The 404 Page


HassanAbouelela (changes requested):
Hmm, let's stop using Imgur on our site. Can you use the asset from the github repo, or add it to the static files?

NeilShah2026:
Ok

jb3 (changes requested):
Can we include the support for `/favicon.ico` in this PR as well? Or if not, open an issue to do so.

That means that devices would fallback on pages where we don't explicitly specify a favicon.

ChrisLovering:
It seems the fork this PR is based from has been deleted, so changes cannot be made. Therefore, I am closing this PR.

If someone still wants to implement these changes, please open an issue for it to be discussed first.

PR #481, opened 2021-04-21
Change index.html to say "100 dedicated expert helpers" instead of 90

whispersofthedawn:
Currently, the index.html file says that we have 90 dedicated helpers under the "Who are we?" section. This pull request changes that to 100.

This was discussed in Discord here: https://canary.discord.com/channels/267624335836053506/635950537262759947/834217942430253067

### DIFF
--- a/pydis_site/templates/home/index.html
+++ b/pydis_site/templates/home/index.html
@@ -82,7 +82,7 @@
           </p>
           <p>
             You can find help with most Python-related problems in one of our help channels.
-            Our staff of over 90 dedicated expert Helpers are available around the clock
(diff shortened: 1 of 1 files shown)


Akarys42:
I feel like we should make it dynamic by fetching the amount of users with the Helpers role in the database. :eyes:

jb3:
> I feel like we should make it dynamic by fetching the amount of users with the Helpers role in the database. 👀

Bear in mind that after dewikification & the API move there will not be a database to fetch this from. Let's just keep it as a constant in the HTML right now, we could add an API route at some stage.

jb3 (approved):

Akarys42 (approved):

PR #482, opened 2021-04-27
New filter schema

### LINKED ISSUE #479: New Filter schemas
# Abstract 

While https://github.com/python-discord/bot/issues/1530 describe the new changes we want for the filter cog group in our bot functionality-wise, this issue will focus on the new database schema proposed to implement those changes. I'd highly recommend you to check out 1530 before reading this issue.

# New Schema Overview 

Here is a visual representation of the proposed schema : 
image
> Note: Due to technical limitations, the list notation for FK had to be placed in the column name.

The schema articulate around the `filter_list` table, representing the persistent state of a `FilterList` class, identified by its name and type. It has a list of `filter`s representing each a `Filter` instance along with three tables, `filter_action`, `filter_list_settings` and `channels_range` to define its default values. Each `filter` is allowed to override its default value by setting a field in its `filter_override` to not NULL.

# Schema Specifications
## `filter_list`
This is the core of the schema, each one representing a particular list in its allow or deny version.

- PK `name`: The name used in the bot-side source code.
- PK `list_type`: One of "deny" or "allow". Whenever this is an allow or deny list.
- FK array `filters`: References to `filter` composing this list
- FK `default_settings`: Reference to the `filter_list_settings` making the default parameter of this list.

## `filter_settings`

This table represents the state of a `Setting` object.

- array `ping_type`: One of "everyone", "here", "moderators", "onduty", "none" or UID. The role(s) or user(s) to ping when the filter triggers.
- `filter_dm`: Should the filter run on DM messages?
- array `dm_ping_type`: Same as `ping_type`. The role to ping when the filter triggers on a DM.
- `delete_messages`: Should the bot automatically delete messages that triggered the message?
- array `bypass_roles`: List of roles or user IDs that are ignored by this filter.
- `enabled`: Is this filter enabled?
- FK `default_action` and `channels_range`: References extending this setting.

## `filter_action`

The moderation action to automatically take when the filter is triggered.

- `user_dm`: DM to send to the user when they trigger the filter. Null if disabled.
- `infraction_type`: One of "warn", "mute", "ban". The infraction to apply to the user if they trigger the filter. Null if disabled.
- `infraction_duration`: The duration of the infraction to apply, if any. Null if permanent.

## `channel_ranges`

The channels where the filter should apply. It contains arrays of allowed and disallowed channels and categories IDs and a default boolean. The resolution is done in order:
- disallowed channels
- disallowed categories
- allowed categories
- allowed channels
- default value

## `filter`

One entry of an allow or denylist. 

- `content`: The string actually making this filter.
- `description`: Information added to the filter to explain why it has been added.
- `additional`: An additional field that is left to the `FilterList` implementation to name and use. Null is not used.
- FK `override`: Reference to a `filter_override` used to override the settings provided by the list itself. 

## `Filter_override`

This contains the override of a specific filter on the filter list default. The field is set to null if no override is performed.

- `ping_type`, `filter_dm`, `dm_ping_type`, `delete_messages`, `bypass_roles`: Override the corresponding fields.
- FK `filter_action`: Reference to a `filter_action` overriding `default_action` all at once.
- FK `filter_range`: Reference to a `channels_range` overriding `default_range` all at once.

# Migration

Migration should be written to take the current content of our filter list and migrate them to the new schema.

Akarys42:
Implements the new filter related schema described in #479. The definition of each model can be found on this issue.

It adds a bunch of new routes under `/bot/filters` for each model, for example `/bot/filter/filter_lists` will provide GET, POST, PUT and DELETE for the `FilterList` model.

It also adds a migration (number 70) that will convert our current filters to the new schema and set some defaults that should mimic our current setup. Note that it will use a default value if the description isn't set, we will have to set one.

Tests have also been updated. I tried to make them modular in case we ever want to modify the design and to avoid having 800 lines of copy pasted code.

Most of this PR is just REST framework boilerplate, enjoy!

Closes #479


pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+from django.core.exceptions import ValidationError
+from django.db import models
+from django.db.models import UniqueConstraint
+
+
+class FilterListType(models.IntegerChoices):
+    """Choice between allow or deny for a list type."""
+
+    ALLOW: 1
+    DENY: 0
```
mbaruh: What is the motivation to make it integer choices, rather than text choices as the infraction types below?

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+
+    NOTE = "Note"
+    WARN = "Warn"
+    MUTE = "Mute"
+    KICK = "Kick"
+    BAN = "Ban"
+
+
+# Valid special values in ping related fields
+VALID_PINGS = ("everyone", "here", "moderators", "onduty", "admins")
```
mbaruh: Is it necessary to define these? we could just have a list of role ID's.

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+        raise ValidationError(f"{value!r} isn't a valid ping type.")
+
+
+class FilterList(models.Model):
+    """Represent a list in its allow or deny form."""
+
+    name = models.CharField(max_length=50, help_text="The unique name of this list.")
+    list_type = models.IntegerField(
+        choices=FilterListType.choices,
+        help_text="Whenever this list is an allowlist or denylist"
```
mbaruh:
```suggestion
        help_text="Whether this list is an allowlist or a denylist"
```

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+
+class FilterSettings(models.Model):
+    """Persistent settings of a filter list."""
+
+    ping_type = ArrayField(
+        models.CharField(max_length=20),
+        validators=(validate_ping_field,),
+        help_text="Who to ping when this filter triggers."
+    )
+    filter_dm = models.BooleanField(help_text="Whenever DMs should be filtered.")
```
mbaruh:
```suggestion
    filter_dm = models.BooleanField(help_text="Whether DMs should be filtered.")
```

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+    dm_ping_type = ArrayField(
+        models.CharField(max_length=20),
+        validators=(validate_ping_field,),
+        help_text="Who to ping when this filter triggers on a DM."
+    )
+    delete_messages = models.BooleanField(
+        help_text="Whenever this filter should delete messages triggering it."
```
mbaruh:
```suggestion
        help_text="Whether this filter should delete messages triggering it."
```

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+    bypass_roles = ArrayField(
+        models.BigIntegerField(),
+        help_text="Roles and users who can bypass this filter."
```
mbaruh: How do you tell whether an ID belongs to a role or to a user? should be limited to roles IMO.

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+    enabled = models.BooleanField(
+        help_text="Whenever ths filter is currently enabled."
```
mbaruh:
```suggestion
        help_text="Whether this filter is currently enabled."
```

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+    default_action = models.ForeignKey(
+        "FilterAction",
+        models.CASCADE,
+        help_text="The default action to perform."
```
mbaruh: Not really clear what that means
```suggestion
        help_text="What action to perform on the triggering user."
```

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+    default_range = models.ForeignKey(
+        "ChannelRange",
+        models.CASCADE,
+        help_text="Where does this filter apply."
```
mbaruh:
```suggestion
        help_text="The channels and categories in which this filter applies."
```

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+        models.SET_NULL,
+        null=True,
+        help_text="Override the default settings."
+    )
+
+    def __str__(self) -> str:
+        return f"Filter {self.content!r}"
+
+
+class FilterOverride(models.Model):
```
mbaruh: Can we call it SettingsOverride? since we're overriding the settings.

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+    dm_ping_type = ArrayField(
+        models.CharField(max_length=20),
+        validators=(validate_ping_field,),
+        null=True
+    )
+    delete_messages = models.BooleanField(null=True)
+    bypass_roles = ArrayField(models.IntegerField(), null=True)
+    enabled = models.BooleanField(null=True)
+    filter_action = models.ForeignKey("FilterAction", models.CASCADE, null=True)
+    filter_range = models.ForeignKey("ChannelRange", models.CASCADE, null=True)
```
mbaruh: I think these names should also be the names in FilterSettings.

mbaruh (changes requested):
Thank you, that was record speed.

I think it's a good start so that we have something to test with. Aside from the minor comments below, there are two things I think are important:
* An endpoint to recursively fetch all data in the schema, so that we don't have to join tables manually on the bot side, and just make one API request on startup.
* Very detailed docs, to the point of saying where exactly you need to go to remove a kind of setting.

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+class FilterList(models.Model):
+    """Represent a list in its allow or deny form."""
+
+    name = models.CharField(max_length=50, help_text="The unique name of this list.")
+    list_type = models.IntegerField(
+        choices=FilterListType.choices,
+        help_text="Whenever this list is an allowlist or denylist"
+    )
+
+    filters = models.ManyToManyField("Filter", help_text="The content of this list.", default=[])
```
MarkKoz: This should not be a many-to-many relationship. It doesn't make sense for a filter to belong to multiple lists. It should be a one-to-many relationship, which is simply defined with a foreign key. Django supports [following relationship backwards](https://docs.djangoproject.com/en/3.2/topics/db/queries/#following-relationships-backward), if needed.

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+
+    class Meta:
+        """Constrain name and list_type unique."""
+
+        constraints = (
+            UniqueConstraint(fields=("name", "list_type"), name="unique_name_type"),
+        )
+
+    def __str__(self) -> str:
+        return f"Filter {'allow' if self.list_type == 1 else 'deny'}list {self.name!r}"
```
MarkKoz: Don't hard-code this. It's like a normal Python enum, so you can do `FilterListType(self.list_type)` to get the name from the value. If it had labels defined, then, of course, you could use `FilterListType(self.list_type).label`.

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+        "ChannelRange",
+        models.CASCADE,
+        help_text="Where does this filter apply."
+    )
+
+
+class FilterAction(models.Model):
+    """The action to take when a filter is triggered."""
+
+    user_dm = models.CharField(
```
MarkKoz: `dm_content` is a better name.

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+        help_text="The infraction to apply to this user."
+    )
+    infraction_reason = models.CharField(
+        max_length=1000,
+        help_text="The reason to give for the infraction."
+    )
+    infraction_duration = models.DurationField(
+        null=True,
+        help_text="The duration of the infraction. Null if permanent."
+    )
```
MarkKoz: There should be some validation (*ideally at the database level*) for these. It would make the API more robust. It is likely a mistake if a duration or reason is set but a type isn't. Furthermore, some infraction types don't support permanent durations. In fact, I'm not sure if this is normalised given that there's (sort of?) a relationship between the type and duration.

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+        models.CASCADE,
+        help_text="Where does this filter apply."
+    )
+
+
+class FilterAction(models.Model):
+    """The action to take when a filter is triggered."""
+
+    user_dm = models.CharField(
+        max_length=1000,
```
MarkKoz, line 20: How did you decide this number for the dm and the reason?

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+    allowed_categories = ArrayField(models.IntegerField())
+    default = models.BooleanField()
+
+
+class Filter(models.Model):
+    """One specific trigger of a list."""
+
+    content = models.CharField(max_length=100, help_text="The definition of this filter.")
+    description = models.CharField(max_length=200, help_text="Why this filter has been added.")
+    additional_field = models.BooleanField(null=True, help_text="Implementation specific field.")
```
MarkKoz: This is really limited, both in count and in type. It's conceivable for filters to use a value other than a boolean for setting and to use more than one filter-specific setting.

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+    override = models.ForeignKey(
+        "FilterOverride",
+        models.SET_NULL,
+        null=True,
+        help_text="Override the default settings."
+    )
```
MarkKoz: "Override" is ambiguous. What is it overriding? Only the help text would be able to explain that. Please name this "settings".

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+    - allowed categories
+    - allowed channels
+    - default
+    """
+
+    disallowed_channels = ArrayField(models.IntegerField())
+    disallowed_categories = ArrayField(models.IntegerField())
+    allowed_channels = ArrayField(models.IntegerField())
+    allowed_categories = ArrayField(models.IntegerField())
+    default = models.BooleanField()
```
MarkKoz: What does this mean?

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+    - disallowed categories
+    - allowed categories
+    - allowed channels
+    - default
+    """
+
+    disallowed_channels = ArrayField(models.IntegerField())
+    disallowed_categories = ArrayField(models.IntegerField())
+    allowed_channels = ArrayField(models.IntegerField())
+    allowed_categories = ArrayField(models.IntegerField())
```
MarkKoz: Something to keep in mind:

Arrays may be an obvious example of a non-atomic type. However, 1NF is ambiguous when it comes to atomicity. Arrays _can_ be treated as atomic types, but I believe this means one must not intend to establish any relationships with the individual elements in the array. My further understanding is that as a consequence, one should not intend to perform queries like "which filters does channel x have?"

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,187 @@
...
+    """Validate that the values are either a special value or a UID."""
+    for value in value_list:
+        # Check if it is a special value
+        if value in VALID_PINGS:
+            continue
+        # Check if it is a UID
+        if value.isnumeric():
+            continue
+
+        raise ValidationError(f"{value!r} isn't a valid ping type.")
```
MarkKoz: I propose to split the columns into roles and users so that each can contain homogenous types. Allowing arbitrary IDs for roles offers more flexibility and gets rid of some of the unintuitiveness and complexity of mixing roles and users in a single column.

The alternative to 2 arrays would be a single array with two columns - one for the ID and another for the type.

pydis_site/apps/api/models/bot/filters.py
MarkKoz: There should be validation (again, ideally at the database level) that

1. Each pair of arrays is mutually exclusive.
2. A duplicate value is not inserted into an array.

Starting to make be doubt if arrays are a suitable data type given such constraints. The four arrays could instead be 2 tables that have 2 columns - one for the ID and another for whether its allowed/disallowed. A primary key on the ID would address both of the validation points above.

MarkKoz (changes requested):
I'm only reviewing the schema for now.

A mistake you've made is conflating one-to-many relationships with one-to-one relationships. Recall that a foreign key describes the former, not the latter. However, the relationships between settings and filters is one-to-one, as well as the relationship between settings and actions/ranges.

One-to-one relationships are kind of strange to me in the world of database design, since they _can_ be merged into a single table. There are some technical reasons for separating them (performance, different security policies, etc.), but I think the primary reason it's done here is for semantics. The argument is that it betters the understanding of the schema, and aides in the organisation and maintenance of it. 

The distinction between some of the models in the one-to-one-relationships seem a bit weak. Furthermore, it's not consistent e.g. the infraction fields arguably represent an infraction, so they should be in a separate table for semantic purposes too.

However, at least for the filters to settings relationship, I feel the decision to split them can still hold up. There would be a stronger case if the override was the exact same schema as the defaults (in which case the tables could be merged too), but unfortunately that seems impossible due to one needing to have nullable fields.

The idiomatic way to implement one-to-one relationships in Django is through [inheritance](https://docs.djangoproject.com/en/3.2/topics/db/models/#model-inheritance). The question is whether each model should have its own table or if the fields of the abstract parents should be instead in the child tables.

The latter sounds better to me, since I believe it'd avoid cross-table joins. Since the schema is managed through Django, it will still give the illusion of being separate where it matters. I wonder if it's possible to let the child dictate whether the inherited fields should be nullable; this would avoid defining two separate settings classes.

Note that, if needed, serialisation can be customised to organise things in a different manner than the schema represents it.

MarkKoz:
Should look into which indexes will be good to have, but those can be added at any time down the road if it's now still too early to see how the data will get queried by the client.

pydis_site/apps/api/models/bot/filters.py
mbaruh: There isn't actually anything preventing a mute from being permanent at the database level, we simply chose to not expose such a command on the bot side, but it's still possible to edit a mute to become permanent.
Likewise, notes and warnings seem to be marked as permanent, at least on the bot side.

pydis_site/apps/api/models/bot/filters.py
mbaruh: For infraction reasons I believe there are issues with pagination on the bot side when they're too long. I know we limited nomination reasons to 1000 characters, and recall we did something similar with infractions.

For DM's however the limit is probably not 1000.

pydis_site/apps/api/models/bot/filters.py
mbaruh: Is it acceptable to have a field to tell whether it's a channel or a category? I'd rather these weren't split into multiple tables, as then it complicates grouping them on the bot side.

pydis_site/apps/api/models/bot/filters.py
mbaruh: I prefer the latter.

mbaruh:
@MarkKoz the guiding principle in deciding where we should split settings, is that columns in the settings table represent entries that can act independently. Splitting the infraction fields and the user DM field will mean, as you mentioned in the bot issue, that we might try to DM a user who has been already banned.

Regarding the overrides, it is possible to handle null values on the bot side and just say it means we should skip this attribute, and then make sure we're not writing null values to the defaults. However I'm not sure if this is a good idea.

pydis_site/apps/api/models/bot/filters.py
MarkKoz: Sorrry, I made a mistake. To be clear, I meant a single table with 2 columns, not a single array.

pydis_site/apps/api/models/bot/filters.py
MarkKoz: Yes, I don't see why it wouldn't be acceptable. I don't believe it's possible for IDs to overlap either, if that was a concern.

As for your other point, keep in mind DRF lets you serialise the data in any way you want. It is not limited to being a 1:1 representation of the DB schema. Granted, it's easier when they do align, so the added work and any performance impacts should be weighed in along with  the benefit I mentioned in the original comment.

pydis_site/apps/api/models/bot/filters.py
MarkKoz: Can this import the infraction model and use its reason limit here instead of having two separate constants that need to be maintained in tandem?

mbaruh: This seems like a good approach, yeah.

pydis_site/apps/api/models/bot/filters.py
MarkKoz: I should have said it the other way around. Some infractions aren't meant to expire, like notes and warnings. That's a semantic limitation, not a technical one. However, there is validation imposed that prevents it.

We could allow it here and defer the error to when it actually tries to apply an infraction, but that is a bad idea.

MarkKoz:
> @MarkKoz the guiding principle in deciding where we should split settings, is that columns in the settings table represent entries that can act independently. Splitting the infraction fields and the user DM field will mean, as you mentioned in the bot issue, that we might try to DM a user who has been already banned.

As I mentioned, the JSON schema doesn't have to match the DB schema. Making a split for the reason you mention feels like encoding implementation-specific semantics/detsils into the DB schema. Again, one to one relationships are strange to see and I'm not sure about best practices here, so I don't feel too strongly either way. I do think it's worth discussing and looking into.

> 
> Regarding the overrides, it is possible to handle null values on the bot side and just say it means we should skip this attribute, and then make sure we're not writing null values to the defaults. However I'm not sure if this is a good idea. 

It isn't. It should be enforced in the DB level. As mentioned, we should look into some metaprogramming trickery to let the subclass make the inherited attributes all nullable, assuming the abstract class is the agreed-upon approach.

mbaruh:
Putting everything under settings, and then grouping fields in the JSON schema is fine by me.

pydis_site/apps/api/models/bot/filters.py
Akarys42: There is no role ID for `here` though, how do we want to handle this?

pydis_site/apps/api/models/bot/filters.py
Akarys42: It was just a matter of having a little bit more lightweight field, as this is roughly just a boolean field, there are only two states. I don’t mind changing it though.

pydis_site/apps/api/models/bot/filters.py
Akarys42: We usually do that by trying to resolve it as a role and then a user. We cannot run into conflicts this way.

Akarys42:
I don't think I'll be able to finish this in time as I have a bunch of other stuff to do, sorry.

If you wish to continue this PR feel free to ping me (Akarys#2780) in Discord, and I can give you a very detailed runbook of what is left to do.

pydis_site/apps/api/models/bot/filters.py
kosayoda: Given the situation where a channel/category could be whitelisted for a filter and blacklisted for another, how would a primary key on the ID work to address the validation issues? From what I can see of the current implementation, a `ChannelRange` entry is just the domain in which a filter applies.

MarkKoz: It's been a long time since I've looked into this. I think it'd have to be a composite PK on the channel ID and the filter ID.

pydis_site/apps/api/models/bot/filters.py
mbaruh: I think "default" should be removed.

PR #485, opened 2021-05-04
Dewikification - Implement redirections app

ks129:
This PR adds a new `redirect` app whose responsibility is ensuring that after Dewikification links still work. I had to create a custom redirect view to add functionality for our use case.

There is possible to define 2 types of redirects: exact redirects and prefix redirects.

There are survival tests for every redirect route that ensures:
1. There is only 1 redirect.
2. Redirects to the given URL.
3. Final destination response code is 200.

Add resources and events app redirects, ~~content apps redirects still have to be added (except contributing prefix redirect)~~.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -11,3 +11,7 @@ This directory contains useful documentation for working with and using our site
   * [Development with Docker](setup.md#development-with-docker)
 
   * [Development with `pip`](setup.md#development-with-pip)
+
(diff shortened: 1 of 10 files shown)


lemonsaurus (commented):
Looks like an overall reasonable and well-coded approach. Let's just make sure we've got everything we talked about in #dev-core and I think this is good to merge.

kosayoda (approved):
Would be nice to have one more set of eyes and testing over this.

PR #504, opened 2021-05-17
Fix Build & Deploy workflow shield

doublevcodes:
## Workflow shields
The `Build and Deploy` workflow seems to have been separated into two different workflows. IDK when.

This PR just separates the shield into 2 separate shields for each workflow, because the old README seemed to mess up because of the non-existent shield

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,25 +1,29 @@
 # Python Discord: Site
-[![Discord][10]][11]
+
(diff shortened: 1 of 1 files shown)


Akarys42 (approved):
Lgtm, thanks a lot!

ToxicKidz (approved):

PR #508, opened 2021-05-19
offtopicnames active attribute and support for PUT and PATCH request

### LINKED ISSUE #490: Re-roll off topic names backend changes.
Bot side issue: https://github.com/python-discord/bot/issues/1559

### Context
The re-roll command on the bot side re-names an off-topic channel and marks the previous off-topic channel name to be devoid of being used again.
We will need a new field in the OffTopicChannelName model ([here](https://github.com/python-discord/site/blob/main/pydis_site/apps/api/models/bot/off_topic_channel_name.py)) that describes if the name is to be considered for the channel names.

We are still keeping the names instead of deleting them forever as it will be used to filter new names (to avoid adding similar names).

### Implementation
- New field in the OffTopicChannelName model.
   - Name: `active`
   - Type: `bool`
   - Default: `True`
- Modify current GET filter to select names with the field value `active` equal to `True`.

RohanJnr:
closes #490 
linked to https://github.com/python-discord/bot/pull/1568
## Changes
- New attribute `active` for OffTopicName.
- Can query based on the `active` attribute of OffTopicName.
- Patch and Put support for OffTopicName.
- DETAIL view support for OffTopicName.
- Tests modified as required.


pydis_site/apps/api/serializers.py
```diff
@@ -200,16 +202,10 @@ def to_representation(self, instance: Infraction) -> dict:
...
-    class Meta:
-        """Metadata defined for the Django REST Framework."""
-
-        model = OffTopicChannelName
-        fields = ('name',)
-
-    def to_representation(self, obj: OffTopicChannelName) -> str:
+    def to_representation(self, objects: List[OffTopicChannelName]) -> List[str]:
         """
         Return the representation of this `OffTopicChannelName`.
```
jchristgit: This docstring needs to be updated.

jchristgit (approved):
This looks great! A small comment. Nice use of Walrus.

jchristgit:
@RohanJnr thanks for the docstring update.

There's currently an error in CI due to conflicting migrations:
```py
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0070_auto_20210519_0545, 0071_increase_message_content_4000 in api).
20
To fix them run 'python manage.py makemigrations --merge'
```
Could you look into fixing this?

It also seems the branch needs to be updated.

RohanJnr:
Looks like I forgot to push the new commits, haha. Will do when I am at my pc tmw 👍

pydis_site/apps/api/serializers.py
```diff
@@ -1,4 +1,6 @@
 """Converters from Django models to data interchange formats and back."""
+from typing import List
```
jchristgit: We are on Python 3.9 here, so we could just use `list` directly, right?

pydis_site/apps/api/serializers.py
```diff
@@ -200,25 +202,30 @@ def to_representation(self, instance: Infraction) -> dict:
...
-        """Metadata defined for the Django REST Framework."""
-
-        model = OffTopicChannelName
-        fields = ('name',)
-
-    def to_representation(self, obj: OffTopicChannelName) -> str:
+    def to_representation(self, objects: List[OffTopicChannelName]) -> List[str]:
         """
-        Return the representation of this `OffTopicChannelName`.
+        Return a list representing a list of `OffTopicChannelName`.
```
jchristgit:
```suggestion
        Return a list with all `OffTopicChannelName`s in the database.
```

pydis_site/apps/api/serializers.py
```diff
@@ -200,25 +202,30 @@ def to_representation(self, instance: Infraction) -> dict:
...
 
         This only returns the name of the off topic channel name. As the model
         only has a single attribute, it is unnecessary to create a nested dictionary.
         Additionally, this allows off topic channel name routes to simply return an
         array of names instead of objects, saving on bandwidth.
```
jchristgit, line 222: This paragraph will need to be updated.

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -69,11 +68,11 @@ def get_object(self) -> OffTopicChannelName:
         name = self.kwargs[self.lookup_field]
         return get_object_or_404(queryset, name=name)
 
-    def get_queryset(self) -> QuerySet:
+    def get_queryset(self, **kwargs) -> QuerySet:
```
jchristgit: I'm kind of 50 / 50 on overriding this to add our own arguments, because this
is also called by REST Framework from what I know. That said, we also have
other places where this method is called, and `.filter()` is called again on it
right after. Would it make more sense to stick with the definition as-is
(without the changes from `all`) and update the `get_queryset(**params)` call
from below to perform the `filter` itself?

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
```diff
@@ -69,11 +68,11 @@ def get_object(self) -> OffTopicChannelName:
...
         """Returns a queryset that covers the entire OffTopicChannelName table."""
-        return OffTopicChannelName.objects.all()
+        return OffTopicChannelName.objects.filter(**kwargs)
 
-    def create(self, request: HttpRequest) -> Response:
+    def create(self, request: Request, *args, **kwargs) -> Response:
```
jchristgit, line 75: The `*args` and `**kwargs` you add here, where do they come from? This is called from REST framework.

jchristgit: (same question for `list`)

jchristgit (changes requested):
I like this. I have a few questions, but apart from that, I like this.

RohanJnr:
Thanks for the review again @jchristgit  :D, I will work on them asap.

RohanJnr:
Looks like the otn endpoint is being migrated to https://github.com/python-discord/api, therefore I believe we should be moving this PR to `api` repo as well once https://github.com/python-discord/api/pull/26 is merged.
@jchristgit

RohanJnr:
Looks like the otn endpoint is being migrated to https://github.com/python-discord/api, therefore I believe we should be moving this PR to `api` repo as well once https://github.com/python-discord/api/pull/26 is merged.
@jchristgit

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
RohanJnr: Right, I think its better to stick to the original definition, and we can perform a filter() instead of passing in params to queryset

pydis_site/apps/api/viewsets/bot/off_topic_channel_name.py
RohanJnr: They come from the function definition within the framework.

jchristgit: Sane and good.

jchristgit (approved):
Solid work. Thank you!

One minor suggestion for the future: Instead of using the autogenerated names by Django for migrations (like `auto_20210519_0545`), it's nice to have a short name describing the migration.

D0rs4n:
Hey! Would you mind opening an issue (and a PR if you'd like to) in the [API](https://github.com/python-discord/api) repository regarding the database model changes? @RohanJnr

Bluenix2:
> @RohanJnr thanks for the docstring update.
> 
> There's currently an error in CI due to conflicting migrations:
> ```py
> CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0070_auto_20210519_0545, 0071_increase_message_content_4000 in api).
> 20
> To fix them run 'python manage.py makemigrations --merge'
> ```
> Could you look into fixing this?
> 
> It also seems the branch needs to be updated.

The fix for this was the 0072_merge migration I see correct?

Do want to give `0070_auto_20210519_0545` a more meaningful name?

jchristgit:
> The fix for this was the 0072_merge migration I see correct?

Yes

> Do want to give 0070_auto_20210519_0545 a more meaningful name?

I think due to the dependencies of the other migrations on this migration this will get a bit hairy, unfortunately :(

Bluenix2:
The behaviour of the PATCH/PUT method is a little unexpected. You can send a request to `/api/bot/off-topic-channel-names/existing-name` with the body `{"name": "another-one"}`... which causes *another* off-topic name to be created. Should this be changed or should I approve? Considering that is my only complaint..

That said when using PATCH correctly it worked nicely and made the names no longer show up when specifying `?active=true` (as expected from the changes by this PR).

RohanJnr:
Practically, The `PATCH` method allows modification of all fields on the model. We could leave it as it is as we will only be patching the `active` field. If we really want to restrict the PATCH to 1 field only, i.e, the `active` field, we can add a check to make sure the request body only contains the `active` field. But, I am not sure if we should have this check or leave it as it is?

Bluenix2:
> Practically, The `PATCH` method allows modification of all fields on the model. We could leave it as it is as we will only be patching the `active` field. If we really want to restrict the PATCH to 1 field only, i.e, the `active` field, we can add a check to make sure the request body only contains the `active` field. But, I am not sure if we should have this check or leave it as it is?

What is unexpected is the fact that you can create a new off-topic-name by PUTting (this is primarily what I tried, don't remember if it works the same with PATCH) to an existing off-topic-name. I don't really understand what even caused the support for these new methods as it is done implicitly by Django so I am not sure how much it would take to change this behaviour.

RohanJnr:
My comment above applies to PUT also. The reason for the creation of a new
object is django behaviour towards primary keys. This is from the docs:


`The primary key field is read-only. If you change the value of the primary
key on an existing object and then save it, a new object will be created
alongside the old one.`



https://docs.djangoproject.com/en/4.0/ref/models/fields/#primary-key

Bluenix2 (approved):
I think what we discussed is fine then if it seems like it would be a lot of code to change it - it's not a big deal. Thanks for this, let's get it merged!

jchristgit:
Ready for review.

jchristgit:
Sorry, wrong project...

PR #510, opened 2021-05-22
Add new infraction filters for the infraction rescheduler

bast0006:
Add `permanent`, `types`, `expires_before`, and `expires_before` as filter options for the `bot/infractions` endpoint. This let's us filter for only in-flight infractions to expire, and also lets us put limits on how far in the future we reschedule infractions if we desire to.

### DIFF
--- a/pydis_site/apps/api/tests/test_infractions.py
+++ b/pydis_site/apps/api/tests/test_infractions.py
@@ -163,6 +163,20 @@ class InfractionTests(APISubdomainTestCase):
         self.assertEqual(len(infractions), 1)
         self.assertEqual(infractions[0]['id'], self.superstar_expires_soon.id)
 
+    def test_filter_after_invalid(self):
+        url = reverse('bot:infraction-list', host='api')
+        response = self.client.get(f'{url}?expires_after=gibberish')
+
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(list(response.json())[0], "expires_after")
+
+    def test_filter_before_invalid(self):
+        url = reverse('bot:infraction-list', host='api')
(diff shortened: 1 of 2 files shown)


bast0006:
Ok. I'll add tests for those validation failures.

kosayoda (commented):
From an initial glance, should we have to worry about both the `type` and `types` parameters being used together?

bast0006:
> From an initial glance, should we have to worry about both the `type` and `types` parameters being used together?

I'm not sure, but I don't think so. At the moment, I'm 99% sure doing so behaves mostly as someone would expect. If you specify both `type` and `types`, and `type` is in `types`, then it will return only `type` results. If you specify a `type` that is not in your specified `types` you will get no results. 

We could add an error condition for mixing of parameters like that, but if we do, we might want to do so for the `expires_before` and `expires_after` to ensure that they don't overlay badly?

kosayoda:
I would personally prefer strict parameter validation, although I'm not sure about others. It would go something like:

- `type` and `types` should be mutually exclusive (or maybe just have one parameter for filtering by a type and multiple types?)
- `permanent` and `expires_*` should be mutually exclusive
- `expires_after` should be before `expires_before`

bast0006:
Since `types` and `type` do a different query (IN vs =) I figured I'd leave them and make them exclusive to each other. 

Otherwise, `permanent` and `expires_` are now exclusive and `expires_before` must now be after `expires_after`.

kosayoda (approved):
Looks/tests good to me! Any issues I've missed should surface when we implement the bot side of things.

ChrisLovering (approved):
Tested and all looks good to me!

Backwards compatible too, so can be merged before the bot PR

PR #518, opened 2021-06-04
Update `IntegrityError` pattern for PostgreSQL 13.

jchristgit:
A more specific error is raised on higher PostgreSQL versions, for example, as
with the PostgreSQL version running on my local machine. This commit fixes the
error message pattern matching behaviour by adding an optional match to the
detailed error message.


MarkKoz (approved):
Ideally we would not rely on comparing message strings, but there doesn't seem to be any other source of information that is more stable. At best, we might be able to get an error code, but still wouldn't know the column name.

https://github.com/psycopg/psycopg2/blob/master/lib/errorcodes.py#L236

kosayoda (approved):

PR #521, opened 2021-06-04
Update configuration docs

jchristgit:
When I started the Django app in January 2019, I documented a setup guide and environment variables for local configuration of the app here. These were lacking the new `$METRICITY_DB_URL` environment variable. The `setup.md` file was also changed to simply be a symlink to the current contributing guide for the site.


kosayoda (commented):
Do we even need a `docs/` directory at this point, since all of the configuration here is essentially duplicated in the contributing guide?

jchristgit:
Yes, but also no. The `setup.md` now links to the contributing guide. However, the configuration information is not part of the contributing guide. That said, I'm not sure how important the configuration information is for someone starting to contribute, as the setup guide documents how to run the app locally & `configuration.md` mostly documents additional options.

kosayoda:
Is `configuration.md` not essentially [this section right here](https://pythondiscord.com/pages/guides/pydis-guides/contributing/site/#2-environment-variables)?

The only variables I see that are missing in the contributing guide are `LOG_LEVEL` and `ALLOWED_HOSTS`, both of which I doubt is much used for testing (although I have 0 qualms with adding them to the site guide).

jchristgit:
> Is configuration.md not essentially this section right here?

Partially, yes. `config.md` was made to explain what the environment variables actually do, whilst the setup guide mostly recommends what to set it to. I'd be open to moving the explanations of `config.md` to the contributing guide.

kosayoda:
I'll discuss this with the core devs then get back to you, thanks for the other minor enhancements by the way!

kosayoda:
@jchristgit The consensus is in favour of removing the `docs/` directory and moving the knowledge to the site contributing guide. Would you be willing to create a PR for that?

jchristgit:
> @jchristgit The consensus is in favour of removing the docs/ directory and moving the knowledge to the site contributing guide. Would you be willing to create a PR for that?

Yes, I'd be happy to do so. I'll try to do that today, if that doesn't work, it'll chime in here next week

PR #526, opened 2021-06-07
Move subdomains to query paths.

### LINKED ISSUE #391: Get rid of subdomains in our site
There is no reason we should use subdomains in our app. Subdomains are a complexity that makes authentication management much more difficult, because OAuth across subdomains does not work well.

## Specification
- Get rid of all subdomain handling. instead of `staff.pythondiscord.com`, we will simply use `pythondiscord.com/staff/`. 
- Get rid of the subdomain dependencies.
- Documentation that refers to subdomains (like the Contributing guide for the bot) may need to be updated. Get someone with access to editing it to help you.
- We will need to update the bot, which specifically uses the subdomains in its configuration. This should just be a matter of changing a few lines in the `default-config.yml` file. Please make sure this change is made, that both pull requests are open at the same time and that they get merged at the same time. We cannot risk downtime of the entire API because of this pull request. Have a core dev or an Admin assist with deployment of this change.

jchristgit:
In more detail:

- Use Django URL namespaces (e.g. `api:bot:infractions`) instead of
  `django_hosts` host argument.
- Update the hosts file setup documentation to remove subdomain entries.
- Update the hosts file setup documentation to mention that the entry of
  `pythondiscord.local` is not required and mainly for convenience.
- Rename the `APISubdomainTestCase` to the more fitting
  `AuthenticatedAPITestCase`, as authentication is all that is left that the
  class is doing.
- Drop dependency to `django_hosts`.

Closes #391.

Note to reviewers: Please do not merge the code, I am only seeking your approval. I will coordinate merge and deployment with @jb3 as needed.


lemonsaurus (approved):
A simple and straight forward change, looks like the right idea. I don't comb through every single line of code but if tests are passing and manual testing looks good, this seems fine to me.

jb3 (approved):

jb3 (changes requested):
We might need further modification to support internal routing through NGINX, will coordinate on Discord.

jchristgit:
Thanks, Joe.

jb3 (approved):

PR #532, opened 2021-06-15
Approve more frameworks for CJ8

janine9vn:
- Python Prompt Toolkit is added as an approved framework.
- Unicurses and nurses are added as additional frameworks you can use with curses, although it's a use at your own risk and no promises are made for functionality or stability.


jacobmonck (approved):
Tested it locally, looks good to me.

jb3 (approved):

fisher60 (approved):
All looks good to me, maybe consider changing `Unicurses` to `UniCurses` to match the pypi project, but this is such a tiny detail.

PR #533, opened 2021-06-15
Add cache purge step to CI deploy stage

jb3:
I want to serve site from the edge, since the content there is primarily static. This PR ensures that cached content is cleared upon deploy so that we can benefit from high edge cache TTLs while still having new content deploy immediately.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -44,7 +44,7 @@ jobs:
               site/deployment.yaml
           images: 'ghcr.io/python-discord/site:${{ steps.sha_tag.outputs.tag }}'
           kubectl-version: 'latest'
-          
+
       - name: Purge Cloudflare Edge Cache
         uses: jakejarvis/cloudflare-purge-action@master
         env:
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -44,3 +44,9 @@ jobs:
               site/deployment.yaml
           images: 'ghcr.io/python-discord/site:${{ steps.sha_tag.outputs.tag }}'
           kubectl-version: 'latest'
+          
+      - name: Purge Cloudflare Edge Cache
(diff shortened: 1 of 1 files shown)


ToxicKidz (approved):
Lgtm

jacobmonck (approved):
Joe is the best programmer in the world

MarkKoz:
Is there a smarter way to clear invalidated cache items rather than purging the whole thing? Not sure if that even make sense with the way they store cache, but I don't know how they do that.

whispersofthedawn (approved):

jb3:
> Is there a smarter way to clear invalidated cache items rather than purging the whole thing? Not sure if that even make sense with the way they store cache, but I don't know how they do that.

@MarkKoz If you pay Cloudflare, you can pass a list of wildcards to invalidate. We could pass a list of individual file URLs but it's unnecessary complexity. It's better to just have pages have a slightly higher initial load time (which assets already do – considering we currently serve django pages with no cache other than for static assets).

Below graph shows HTTP reqs to our server from clients in US, HK & UK.

Start of graph is serving from our servers, so responses range from 300ms to 1.2s, so worst case scenario for each Cloudflare PoP that has no cached copy someones request takes up to that duration.

End of this graph shows serving from the Cloudflare PoPs, which makes site response times considerably faster, consistently sub-100ms.

image

MarkKoz:
>  It's better to just have pages have a slightly higher initial load time

Better in terms of what, just reduced complexity?

Anyway a max of 1.2s for a few unlucky people sounds acceptable.

jb3:
> Better in terms of what, just reduced complexity?

Yeah, that's all. The process for conditional invalidation would be along the lines of getting a diff of files changed, mapping that to URLs and then pushing that to Cloudflare, it's possible but is another component to maintain. We'd need to build static assets, reverse all the resource and guide URLs and so on, which is possible but I'm not sure it's worth it.

Akarys42 (approved):

PR #534, opened 2021-06-18
Add support for Discord.py 1.7+

ChrisLovering:
With the addition of new permissions in Discord, the max permission number now exceeds that which can be stored in an int.

This needs to be merged before we can bump the bot up to 1.7 and Discord.py add support for these new permissions. I have removed the max value validator, as this just adds maintenance burden for us each time Discord adds new permissions.

See this table in the Discord API docs to see that the new max number 1 << 36
https://discord.com/developers/docs/topics/permissions#permissions-bitwise-permission-flags


ToxicKidz (dismissed):

Numerlor:
Is the MaxValueValidator necessary? It doesn't feel like it adds much value while adding a maintenance cost. The current max value in discord.py is also 33 bit so I'm not even sure if that or 36 would fit more

vcokltfre:
Bear in mind thread permissions are coming at some point, so with a constraint on bit length another PR like this will need to be made to upgrade to a d.py version that supports those permissions

ChrisLovering:
> Is the MaxValueValidator necessary? It doesn't feel like it adds much value while adding a maintenance cost. The current max value in discord.py is also 33 bit so I'm not even sure if that or 36 would fit more

Hmm yea, if we went with d.py, then we'd need to bump this with a migration each time. If we went with 36, it gives us some time, but it won't be long until Discord adds new permissions.

I tend to agree that we could remove this validator entirely.

vcokltfre (dismissed):
Approving, but bear in mind above comments for the future

Numerlor:
> Bear in mind thread permissions are coming at some point, so with a constraint on bit length another PR like this will need to be made to upgrade to a d.py version that supports those permissions

the 36 bit val should take in account the new thread perms (unless some new ones are added after), but then it won't be a correct check for current d.py. I'd say the validator should just be removed

ChrisLovering:
converted to draft as I am going to remove the max value validator later this evening.

Numerlor (approved):

Xithrius (approved):

kosayoda (approved):

PR #536, opened 2021-06-18
Adjust Message model to support new nitro messages with over 4000 chars

### LINKED ISSUE #535: Message Model is limited to 2k characters
https://github.com/python-discord/site/blob/297337fa6552ec32d95e82dc41c018ef1a5e3cf3/pydis_site/apps/api/models/bot/message.py#L46

Nitro now provides the ability to submit messages 4k characters long. When antispam in the bot attempts deletion of messages this size, it uploads them to bot/deleted_messages, which then fails with 400 Bad Request because an individual message doesn't validate. 

I've marked this as high because we actually lose the content of the deleted messages here due to the failed POST from the bot.

bast0006:
Fixes #535


bast0006:
Ah yes. I forgot to run a coverage run. Still really confused how this didn't raise a coverage error before? I didn't delete any tests so coverage shouldn't have gone down?

Oh. It's saying that there isn't test coverage for the actual test itself?

pydis_site/apps/api/tests/test_models.py
```diff
@@ -34,6 +35,43 @@ def test_shows_attributes(self):
...
+            channel_id=666,
+            content="w"*4000,
+            deletion_context=self.context,
+            embeds=[]
+        )
+
+        try:
+            message.clean_fields()
+        except Exception as e:
+            self.fail(f"Creation of message of length 3950 failed with: {e}")
```
Akarys42: This except block is causing the 100% coverage to fail because it is never run. Can you please obliterate it?

Akarys42 (changes requested):
Thanks! The coverage is due to a branch never being ran, see my message below.

pydis_site/apps/api/tests/test_models.py
bast0006: See #539

ChrisLovering: As discussed in dev-core, we should add a pragma no cover here, so this specific branch is ignored, but we keep coverage on other tests

ChrisLovering (approved):

bast0006:
I'm going to rebuild the migration since now migrations have broken with the latest other branch merge to main

Akarys42 (approved):
:shipit:

PR #539, opened 2021-06-21
Fix .coveragerc testing branch coverage of tests as well as project code

bast0006:


Akarys42 (approved):

mbaruh (dismissed):

Numerlor:
Do we want to exclude tests from coverage? Knowing if some test or its portion didn't run can be important, and both are also measured on the bot repo. For cases where it isn't feasible to do we can ignore the part that isn't covered

mbaruh:
Yeah let's go with the other option, let's just ignore the particular test we don't want to be included.

PR #545, opened 2021-06-27
Docker-compose: upgrade to postgres 13

Akarys42:
We have been using postgres 13 in production for a long time, but we never actually updated the docker-compose environment.

A draft of the announcement can be found [here](https://discord.com/channels/267624335836053506/411200599653351425/858693693342875659).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@
 version: "3.6"
 services:
   postgres:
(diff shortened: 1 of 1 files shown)


ChrisLovering (approved):
Number go up 👍

HassanAbouelela (approved):
Best PR, the only PR 🚀

PR #553, opened 2021-07-11
Improvements & fixes to homepage repository display.

jchristgit:
These changes mainly aim to prevent multiple concurrent requests from
deadlocking the table due to the lazy evaluation imposed by `objects.all()`,
and introduce some quality of life changes whilst doing so.

To prevent unwanted locks, the following is done:
- Instead of evaluating the first item of `objects.all()` only in order to
  determine when the table was most recently updated, we query for the least
  recent updated entry and use that instead. As we use `.first()`, the model
  instance is loaded directly.

The following quality of life changes are introduced:
- Instead of manual "update or create" logic for repositories retrieved from
  the API, use Django's `update_or_create` instead.
- Instead of manual bulk creation of entries on initial data retrieve, Django's
  `bulk_create` method is used instead.
- To allow for local testing of this endpoint without having to manually set up
  GitHub REST authorization, anonymous GitHub access is used when no
  `GITHUB_TOKEN` is configured.

### DIFF
--- a/pydis_site/apps/home/views/home.py
+++ b/pydis_site/apps/home/views/home.py
@@ -19,7 +19,6 @@ class HomeView(View):
 
     github_api = "https://api.github.com/users/python-discord/repos?per_page=100"
     repository_cache_ttl = 3600
-    headers = {"Authorization": f"token {GITHUB_TOKEN}"}
 
     # Which of our GitHub repos should be displayed on the front page, and in which order?
     repos = [
@@ -34,6 +33,16 @@ class HomeView(View):
     def __init__(self):
         """Clean up stale RepositoryMetadata."""
         RepositoryMetadata.objects.exclude(repo_name__in=self.repos).delete()
+
+        # If no token is defined (for example in local development), then
+        # it does not make sense to pass the Authorization header. More
+        # specifically, GitHub will reject any requests from us due to the
+        # invalid header. We can make a limited number of anonymous requests
+        # though, which is useful for testing.
+        if GITHUB_TOKEN:
+            self.headers = {"Authorization": f"token {GITHUB_TOKEN}"}
+        else:
+            self.headers = {}
 
     def _get_api_data(self) -> Dict[str, Dict[str, str]]:
         """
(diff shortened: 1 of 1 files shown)


pydis_site/apps/home/views/home.py
```diff
@@ -74,58 +83,49 @@ def _get_api_data(self) -> Dict[str, Dict[str, str]]:
 
     def _get_repo_data(self) -> List[RepositoryMetadata]:
         """Build a list of RepositoryMetadata objects that we can use to populate the front page."""
-        database_repositories = []
-
-        # First, let's see if we have any metadata cached.
-        cached_data = RepositoryMetadata.objects.all()
+        # First off, load the timestamp of the least recently updated entry.
+        oldest_entry = RepositoryMetadata.objects.order_by("last_updated").first()
```
bast0006: Consider using .values_list("last_updated", flat=True) here to not fetch the entire object and only get the field we need?

bast0006 (commented):
Otherwise, looks good to me. We are reconsidering using the database to store this information, but for now this will hopefully fix the deadlock issue.

pydis_site/apps/home/views/home.py
jchristgit: I thought about this initially but decided against it due to flake8 then wanting to wrap the line. I ended up with something like:

```py
last_update = (
    RepositoryMetadata.objects.order_by("last_updated")
    .values_list("last_updated", flat=True).first()
)
```
or
```py
last_update = (
    RepositoryMetadata
    .objects
    .order_by("last_updated")
    .values_list("last_updated", flat=True)
    .first()
)
```

and I think the current approach is much more readable. I can still change it though.

bast0006: Well, given the flexibility in ordering we have due to how django QuerySet's work, I think the following would probably be cleanest:

```py
last_update = (
    RepositoryMetadata.objects.values_list("last_updated", flat=True)
    .order_by("last_updated").first()
)
```

Basically
- what we're getting
- how we're filtering it

jchristgit: Pushed

ChrisLovering (approved):
This change looks good to me 👍

bast0006 (approved):
LGTM

PR #563, opened 2021-08-11
Implement front-end spec for Smarter Resources

doublevcodes:
This PR will do the following things:
 - [X] Make the following search options be ticked by default
     - General
     - Tutorial
     - Free
     - Beginner
 - [x] Split the `Topic` column into two
 - [X] Center the `Search` and `Clear` buttons
 - [ ] Link to "Edit resource on GitHub"
 - [ ] Show an error message if no options are selected


swfarnsworth:
Great work so far! I forgot to mention: I'd also like for there to be a specific "error message" if the user doesn't check any options for a given parameter (as this will cause an intersection with the empty set and yield no results for the whole query). If this must be handled by the backend, let me know in what way and I'll adjust it accordingly.

pydis_site/apps/resources/views/resources.py
```diff
@@ -23,12 +23,15 @@ def resource_view(request: HttpRequest) -> HttpResponse:
...
+    topics = sorted(RESOURCE_META_TAGS.get("topics"))
+
     return render(
         request,
         template_name="resources/resources.html",
         context={
             "checkboxOptions": checkbox_options,
-            "topics": sorted(RESOURCE_META_TAGS.get("topics")),
+            "topics_1": topics[:len(topics) // 2],
+            "topics_2": topics[len(topics) // 2:],
```
swfarnsworth, line 34: We need to double check that this doesn't affect the back end logic in any way.

doublevcodes: It doesn't

pydis_site/apps/resources/templatetags/get_category_icon.py
```diff
@@ -0,0 +1,38 @@
...
+        "Tool": "fa-tools",
+        "Tutorial": "fa-clipboard-list",
+        "Video": "fa-video",
+        "Free": "fa-first-aid",
+        "Paid": "fa-sack",
+        "Subscription": "fa-credit-card",
+        "Beginner": "fa-play-circle",
+        "Intermediate": "fa-align-center"
+    }
+    icon_name = icons[name]
```
swfarnsworth, line 37: This dictionary should be a global so that it's not being re-created each time. Try naming it `_ICONS`.

Bluenix2: Pretty sure Python handles this and uses LOAD CONST, that said using an all-caps variable at the top of the file is cleaner.

swfarnsworth: Also I wish we had function composition so that this function could just be `get_category_icon = register.filter @ 'fa {icon_name}'.format @ icons.get`, but no :disappointed:

swfarnsworth:
Superseded by #570, which includes all of these commits. Thanks again!

PR #567, opened 2021-08-20
User events management

RohanJnr:
WIP
for https://github.com/issues/assigned


jchristgit:
Hi @RohanJnr, just checking in to see whether you need any help with this?

jb3:
Unfortunately, I don't think we're going to be needing this system right now, since we don't host user events anymore and can handle staff event management without it. With the introduction of the new Discord Events system we can handle interest registration and notifications for events both privately and privately. Thank you for the contributions nonetheless though, it is greatly appreciated!

PR #568, opened 2021-08-25
Run collectstatic in the Docker build process and disable in production

jb3:
Don't run collectstatic on container start, run it in the build process!

This bakes all our static files into the Docker image meaning that we don't have to perform writes to site-packages on startup allowing us to remove some volumes from python-discord/kubernetes#92 and have a slightly faster application start up time.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -142,7 +142,7 @@ class SiteManager:
         if self.debug:
             # In Production, collectstatic is ran in the Docker image
             print("Collecting static files.")
-            call_command("collectstatic", interactive=False, clear=True, verbosity=self.verbosity)
+            call_command("collectstatic", interactive=False, clear=True, verbosity=self.verbosity - 1)
 
             self.set_dev_site_name()
             self.create_superuser()
--- a/manage.py
+++ b/manage.py
@@ -138,10 +138,12 @@ class SiteManager:
 
         print("Applying migrations.")
         call_command("migrate", verbosity=self.verbosity)
-        print("Collecting static files.")
-        call_command("collectstatic", interactive=False, clear=True, verbosity=self.verbosity)
 
         if self.debug:
+            # In Production, collectstatic is ran in the Docker image
+            print("Collecting static files.")
+            call_command("collectstatic", interactive=False, clear=True, verbosity=self.verbosity)
+
             self.set_dev_site_name()
             self.create_superuser()
 
--- a/manage.py
+++ b/manage.py
@@ -142,7 +142,12 @@ class SiteManager:
         if self.debug:
             # In Production, collectstatic is ran in the Docker image
             print("Collecting static files.")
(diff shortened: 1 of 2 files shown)


MarkKoz:
Docker image fails to build locally

```
You have requested to collect static files at the destination
location as specified in your settings:

    /app/staticfiles

This will overwrite existing files!
Are you sure you want to do this?

Type 'yes' to continue, or 'no' to cancel: Traceback (most recent call last):
  File "/app/manage.py", line 198, in <module>
    main()
  File "/app/manage.py", line 194, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 328, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 369, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.9/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 184, in handle
    if input(''.join(message)) != 'yes':
EOFError: EOF when reading a line
ERROR: Service 'web' failed to build: The command '/bin/sh -c SECRET_KEY=dummy_value     DATABASE_URL=postgres://localhost     METRICITY_DB_URL=postgres://localhost     python manage.py collectstatic' returned a non-zero code: 1
```

jb3:
Interesting, I think interactive mode needs disabling and clear enabling, as we do in the manage.py file. I've pushed a change for that, can you give it a try now?

MarkKoz (approved):
Yup that fixed it.

manage.py
```diff
@@ -138,10 +138,12 @@ def prepare_server(self) -> None:
 
         print("Applying migrations.")
         call_command("migrate", verbosity=self.verbosity)
-        print("Collecting static files.")
-        call_command("collectstatic", interactive=False, clear=True, verbosity=self.verbosity)
 
         if self.debug:
+            # In Production, collectstatic is ran in the Docker image
+            print("Collecting static files.")
+            call_command("collectstatic", interactive=False, clear=True, verbosity=self.verbosity)
```
Akarys42: Right now, due to the verbosity of this command, reloading the server generates about 200 log lines, saying the existing assets have been deleted. I think it is causing more harm than good, so I propose to reduce the verbosity by one.

```suggestion
            call_command("collectstatic", interactive=False, clear=True, verbosity=self.verbosity - 1)
```

With this suggestion applied, it only generates one log line, `Collecting static files.`.

Akarys42 (approved):
Looks like a decent fix to me. I have one pretty much kaizen suggestion before merging.

PR #572, opened 2021-08-31
Create a signal to unassign roles from user when the role is deleted

### LINKED ISSUE #369: User's role field needs to have a cascade delete
#357 replaced the FK with an `ArrayField`. This introduced a problem of the roles table falling out of sync with the users table. I encountered an issue when trying to sync a user with the bot. One of their roles had been deleted from the roles table, but it was still present in their roles field. The validator ensures all roles exist in the table, so the request was failing.

Apparently [Django Postgres Extensions](https://django-postgres-extensions.readthedocs.io/en/latest/arrays.html) have a solution for this.

D0rs4n:
Add a signal to the api app that automatically unassigns
deleted roles from users that have them
Closes #369

### DIFF
--- a/pydis_site/apps/api/signals.py
+++ b/pydis_site/apps/api/signals.py
@@ -1,0 +1,12 @@
+from django.db.models.signals import pre_delete
+from django.dispatch import receiver
+
+from pydis_site.apps.api.models.bot import Role, User
+
+
+@receiver(signal=pre_delete, sender=Role)
+def delete_role_from_user(sender: Role, instance: Role, **kwargs) -> None:
+    """Unassigns the Role (instance) that is being deleted from every user that has it."""
+    for user in User.objects.filter(roles__contains=[instance.id]):
+        del user.roles[user.roles.index(instance.id)]
+        user.save()
--- a/pydis_site/apps/api/signals.py
+++ b/pydis_site/apps/api/signals.py
@@ -1,10 +1,10 @@
-from django.db.models.signals import pre_delete
+from django.db.models.signals import post_delete
 from django.dispatch import receiver
 
 from pydis_site.apps.api.models.bot import Role, User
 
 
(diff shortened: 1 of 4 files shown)


pydis_site/apps/api/signals.py
```diff
@@ -0,0 +1,12 @@
...
+
+from pydis_site.apps.api.models.bot import Role, User
+
+
+@receiver(signal=pre_delete, sender=Role)
+def delete_role_from_user(sender: Role, instance: Role, **kwargs) -> None:
+    """Unassigns the Role (instance) that is being deleted from every user that has it."""
+    for user in User.objects.filter(roles__contains=[instance.id]):
+        del user.roles[user.roles.index(instance.id)]
+        user.save()
```
Bluenix2, line 12: Is there not a way to do this in batch? We should let SQL do this

D0rs4n: Oh, I see your concern. That wasn't a wise choice indeed. Although, I think there's a way to solve this using Django ORM.

D0rs4n: It seems like, the Django ORM, F objects in particular does not support ArrayFields. Hence that's the only way to solve it via the Django ORM, for now.
However, I've never really worked with arrayfields using raw SQL.

jchristgit: This isn't possible with plain PostgreSQL, except if we were to use triggers, and I would currently stay away from that personally.

The original approach here was the correct way: foreign key constraints. But since this got removed due to performance issues (waiting on reply from a Django expert friend on this topic), this is a fine approach.

jchristgit (approved):
This looks good, thank you!

Could you rebase off master, please?

jchristgit:
Hi @Akarys42, could you give this a second review?

ChrisLovering (approved):
The seems to work well locally 👌

PR #573, opened 2021-08-31
Update gunicorn options for deployment

jb3:
Remove threads & max requests configuration

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -176,10 +176,7 @@ class SiteManager:
             "--preload",
             "-b", "0.0.0.0:8000",
             "pydis_site.wsgi:application",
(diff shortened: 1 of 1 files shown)


jchristgit (approved):
so realistically we want to move this configuration part into kubernetes. in my opinion, this has no place here. for instance, why are gunicorn workers checking migrations at every worker boot? this should be checked once at startup by an entrypoint script and then `exec gunicorn` and fast-booting workers excellent. this looks good regardless, thank you joe

jb3:
Well said, Christ.

ChrisLovering (approved):
👌

PR #575, opened 2021-09-02
Add Django Prometheus

jb3:
Adds Django Prometheus and installs the necessary middleware and URL paths for reporting.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ gunicorn = "~=20.0.4"
 sentry-sdk = "~=0.19"
 markdown = "~=3.3.4"
 python-frontmatter = "~=1.0"
-django-prometheus = "^2.1.0"
+django-prometheus = "~=2.1"
 
 [tool.poetry.dev-dependencies]
 coverage = "~=5.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ gunicorn = "~=20.0.4"
 sentry-sdk = "~=0.19"
 markdown = "~=3.3.4"
 python-frontmatter = "~=1.0"
+django-prometheus = "^2.1.0"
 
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -21,6 +21,7 @@ gunicorn = "~=20.0.4"
 sentry-sdk = "~=0.19"
 markdown = "~=3.3.4"
 python-frontmatter = "~=1.0"
+django-prometheus = "^2.1.0"
```
doublevcodes:
```suggestion
django-prometheus = "~=2.1"
```
This should be changed to follow the format of the other pinned dependencies.

doublevcodes (changes requested):

doublevcodes (approved):
LGTM now :shipit:

Den4200 (approved):

jchristgit (changes requested):
Yes, but also no.

This is nice. This needs a change though, because we are using multiple workers, because Python can't gobble around on more than one core in the same address space.

`gunicorn` needs to be started with a configuration file as its argument, with the following contents:
```py
from prometheus_client import multiprocess

def child_exit(server, worker):
    multiprocess.mark_process_dead(worker.pid)
```

This is step three from the Prometheus Python client's multiprocess mode considerations: https://github.com/prometheus/client_python. Step two is [taken care of](https://github.com/korfuri/django-prometheus/blob/434a3ba36bdada45c9633451f5f6cfd145814ccf/django_prometheus/exports.py#L119) by `django-prometheus`, step one I will PR into the Kubernetes repository.

jchristgit:
And as a small bonus, it would be nice if you could add a small comment in the `Dockerfile` explaining why we need `BUILDING_DOCKER=yes`.

jchristgit (commented):
Solid. Thanks!

PR #581, opened 2021-09-04
Pin platform in Dockerfile

ChrisLovering:
Some of our deps don't have wheels for atm processors.

With Mac's M1 chips becomming more common, we should make it easier for those users to build our environments.


jchristgit (approved):
Rock solid.

unknown-2

+1 on removing the pin for `.5`. Patch versions, no problem, no problem.

Xithrius (approved):
Works on my m1

jchristgit:
Thanks!

PR #582, opened 2021-09-05
Smarter Resources

### LINKED ISSUE #457: Smarter Resources: Link to GoodReads instead of Amazon.
## Updated issue by lemonsaurus
Let's update all our book links to link to GoodReads instead of Amazon. It feels icky to direct all our users to Amazon, but GoodReads is a more neutral destination which links to many different stores.

## Original issue
When on the https://pythondiscord.com/pages/resources/reading/ page, all the links are for the American amazon. 
It is easy enough for a user to find their Amazon-specific, I'm from England so I just change the .com

 amazon.com -> amazon.co.uk  
 
 Not a massive issue, but it may help users from other location than America find the book they can purchase etc.
 There may be more cases than just Book links that this occurs with.

swfarnsworth:
This PR contains the first iteration of the new resources page. I expect that a few of the tests will not pass but am otherwise ready for criticism to prepare it for production.


pydis_site/apps/redirect/redirects.yaml
```diff
@@ -87,10 +87,6 @@ resources_index_redirect:
   original_path: pages/resources/
   redirect_route: "resources:index"
 
-resources_resources_redirect:
-  original_path: pages/resources/<str:category>/
-  redirect_route: "resources:resources"
-
```
swfarnsworth: There will be other missing pages now that we've consolidated the resources into one page, though I'm not sure how to create redirects for these.

pydis_site/static/css/resources/resources.css
```diff
@@ -1,29 +1,29 @@
...
-    min-height: 15vh;
-}
-.tile.is-parent:hover .box {
-    box-shadow: 0 10px 20px rgba(0,0,0,0.19), 0 6px 6px rgba(0,0,0,0.23);
-}
-.tile.is-parent:hover {
-    padding: 0.65rem 0.85rem 0.85rem 0.65rem;
-    filter: saturate(1.1) brightness(1.1);
-}
+/*.box, .tile.is-parent {*/
```
whispersofthedawn: Is this file going to be kept? It seems to be fully commented out.

pydis_site/templates/resources/resources.html
```diff
@@ -1,90 +1,199 @@
...
+                                        <label class="checkbox">
+                                            <input class="topic" name="topic" type="checkbox" value="{{ topic }}"{% if topic == 'General' %} checked{% endif %}>
+                                            <span class="has-text-grey is-size-6">{{ topic }}</span>
+                                        </label>
+                                    </div>
+                                    {% endfor %}
+                                </div>
+                            </div>
+                            <span class="control ml-2 is-flex is-align-items-end is-justify-content-center mt-auto">
+                                <button onclick="selectAllQueryParams('topic')" class="button is-success is-small">Select All</button>
```
whispersofthedawn: This button does not seem to do anything if you click it when everything is already selected, is that intended?

whispersofthedawn (commented):
This looks pretty good to me. I don't know much Django so I may have missed some important things, but what I know looks fine.

pydis_site/apps/resources/tests/test_views.py
```diff
@@ -19,16 +19,13 @@ def test_resources_index_200(self):
...
-    @patch("pydis_site.apps.resources.views.resources_list.RESOURCES_PATH", TESTING_RESOURCES_PATH)
-    def test_invalid_resource_list_404(self):
-        """Check does site return code 404 when trying to visit invalid resource list."""
-        url = reverse("resources:resources", ("invalid",))
-        response = self.client.get(url)
-        self.assertEqual(response.status_code, 404)
+    @patch("pydis_site.apps.resources.resource_search.RESOURCES_PATH", TESTING_RESOURCES_PATH)
+    def test_filter_resource_list(self):
+        """TODO: Check that we can correctly filter resources with GET parameters."""
+        pass
```
brad90four: So this test is just going to be skipped?

pydis_site/apps/resources/tests/test_resources.py
```diff
@@ -0,0 +1,9 @@
+from django.test import TestCase
+
+
+class TestResources(TestCase):
+    """Test our resource filtering systems."""
+
+    def test_utils_to_retrieve_tags(self):
+        """Test that the utils that retrieve the tags work as intended."""
+        pass
```
brad90four: Skipping this test?

brad90four (dismissed):

swfarnsworth:
I believe all that remains is to redirect all existing resources pages to the one resource page, and to decide about the tests that aren't implemented. My filtering logic, which will probably be translated into JS (by someone who isn't me) when we make the page reactive, does not have a unit test.

pydis_site/apps/redirect/redirects.yaml
```diff
@@ -87,9 +87,45 @@ resources_index_redirect:
...
-  redirect_route: "resources:resources"
+resources_reading_redirect:
+  original_path: resources/reading/
+  redirect_route: "resources:index"
+  redirect_arguments: ["book"]
+
+resources_books_redirect:
+  original_path: resources/books/
+  redirect_route: "resources:index"
+  redirect_arguments: [ "book" ]
```
jchristgit: This should match the format of the other blocks.

```suggestion
  redirect_arguments: ["book"]
```

pydis_site/apps/resources/tests/test_resources.py
jchristgit: @lemonsaurus what should we do here?

pydis_site/apps/resources/views/resources.py
```diff
@@ -1,7 +1,119 @@
...
+        }
+
+    def get(self, request: WSGIRequest, resource_type: t.Optional[str] = None) -> HttpResponse:
+        """List out all the resources, and any filtering options from the URL."""
+        # Add type filtering if the request is made to somewhere like /resources/video.
+        # We also convert all spaces to dashes, so they'll correspond with the filters.
+        if resource_type:
+            dashless_resource_type = resource_type.replace("-", " ")
+
+            if dashless_resource_type.title() not in self.filters['Type']['filters']:
```
jchristgit:
```suggestion
            if dashless_resource_type.title() not in self.filters["Type"]["filters"]:
```

pydis_site/templates/resources/resource_box.html
```diff
@@ -9,14 +11,69 @@
...
-    {# Icons #}
-    {% for icon in resource.urls %}
-        <span class="icon is-size-4 is-medium" style="margin: 5px;">
-            <a href="{{ icon.url }}">
-                <i class="{{ icon.icon|as_icon }} is-size-3 resource-icon is-{{ icon.color }}"></i>
-            </a>
-        </span>
-    {% endfor %}
+    <div class="is-flex is-align-items-center">
+        {#  Add primary link #}
```
jchristgit:
```suggestion
        {# Add primary link #}
```

pydis_site/templates/resources/resource_box.html
```diff
@@ -9,14 +11,69 @@
...
+        {% if 'title_url' in resource %}
```
jchristgit:
```suggestion
        {% if "title_url" in resource %}
```

pydis_site/templates/resources/resource_box.html
```diff
@@ -9,14 +11,69 @@
...
+            {% for tag in resource.tags.difficulty %}
+            <span
+                class="tag resource-tag has-background-info-light has-text-info-dark ml-2 mt-2"
+                data-filter-name="difficulty"
+                data-filter-item="{{ tag|as_css_class }}"
+            >
+                <i class="{{ tag|title|get_category_icon }} mr-1"></i>
+                {{ tag|title }}
+            </span>
+            {% endfor %}
```
jchristgit, line 76: I am sort of wondering whether we can merge these together. The only thing that changes is `data-filter-name` and the thing we are iterating over, or am I misunderstanding something?

pydis_site/templates/resources/resources.html
```diff
@@ -1,90 +1,164 @@
...
+                                                <span
+                                                        class="filter-box-tag tag is-primary is-light ml-2 mt-2"
+                                                        data-filter-name="{{ filter_name|as_css_class }}"
+                                                        data-filter-item="{{ filter_item|as_css_class }}"
+                                                >
+                                                        <i class="{{ filter_item|title|get_category_icon }} mr-1"></i>
+                                                        {{ filter_item|title }}
+                                                        <button class="delete is-small is-primary has-background-primary-light"></button>
+                                                    </span>
+                                            {% endif %}
```
jchristgit, line 87: Can we maybe simplify this a bit? Maybe move the `if` condition to the `button` itself to just control its tags? Or am I missing other diffs here?

pydis_site/apps/resources/templatetags/as_css_class.py
```diff
@@ -0,0 +1,18 @@
...
+
+@register.filter
+def as_css_class(class_name: str) -> str:
+    """
+    Convert any string to a css-class name.
+
+    For example, convert
+    "Favorite FROOT_is_LEMON" to
+    "favorite-froot-is-lemon"
+    """
```
lemonsaurus: This is called `kebab-case`, I learned today. Let's rename this function to use the correct term, and maybe cover more edge cases, too.

pydis_site/apps/resources/tests/test_resources.py
lemonsaurus: We should just delete this file. There are no utils to retrieve tags anymore.

pydis_site/apps/resources/views/resources.py
```diff
@@ -1,7 +1,128 @@
...
+                    tag = tag.title()
+                    tag = tag.replace("And", "and")
+                    resource_tags[tag_type].add(tag)
+
+                # Make a CSS class friendly representation too, while we're already iterating.
+                for tag in tags:
+                    css_tag = f"{tag_type}-{tag}"
+                    css_tag = css_tag.replace("_", "-")
+                    css_tag = css_tag.replace(" ", "-")
+                    css_classes.append(css_tag)
```
lemonsaurus: Why aren't we just using the as_css_class template tag here? This is just duplicating code from there.

pydis_site/templates/resources/resource_box.html
lemonsaurus: they also all have different span classes (to give them different colors), so yeah we could do it, but we'd need to put those classes (like `has-background-info-light` and `has-text-info-dark`) into the dict we're iterating or something... Would be nice and clean, though.

pydis_site/templates/resources/resources.html
lemonsaurus: yeah so here the colors are both in the span class (`is-primary`) and in the delete button (`has-background-primary-light`), so you'd need twice as many if's, which wouldn't really be that much simpler. The only way to simplify it down to a single span would be to store all these colors somewhere in the view and make sure I can get them from filters.items, which adds more clutter to the view.. but might be a better solution. I'm not sure.

lemonsaurus (changes requested):

jchristgit:
Lemon, this is absolutely brilliant. Rest assured that I will approve this when requested.

pydis_site/templates/resources/resource_box.html
lemonsaurus: Refactoring this turns out to be quite a big job, and I'm not sure I think it's a good idea to start storing lots of CSS classes in the view logic. #wontfix.

pydis_site/templates/resources/resources.html
lemonsaurus: Same thing here - I don't wanna store these color classes in the view. It violates the [locality of reference principle](https://en.wikipedia.org/wiki/Locality_of_reference), which feels like too great a cost to pay for cleaner template code.

The classes belong with the HTML element, not in some dict in the Python view.

lemonsaurus:
@jchristgit 

All review comments addressed. Thanks for the review!

lemonsaurus (approved):
I've addressed all the issues I was able to find with my own code.

jchristgit (approved):
Thank you lemon!

image

mbaruh (approved):
By the power vested in me, this LGTM

PR #583, opened 2021-09-05
Initialise metricity at runtime

ChrisLovering:
Currently the bot cannot start in dev as the site errors, saying that metricity doesn't exist. Previously this note existing was fine, unless you needed to use metricity data. With the recent addition of django-prometheus, metricity is now required on boot.

This PR moves the init of metricity from a docker-compose volume, into running of the site. This means that external projects using site, that don't have access to the init.sql file to mount a volume, now also init metricity.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -161,7 +161,6 @@ class SiteManager:
             exists = cursor.fetchone()
             if exists:
                 # Assume metricity is already populated if it exists
-                print("Metricity already exists, not creating.")
                 return
             print("Creating metricity relations and populating with some data.")
             cursor.execute("CREATE DATABASE metricity")
@@ -171,8 +170,8 @@ class SiteManager:
             database="metricity",
             **db_connection_kwargs
         )
-        with conn.cursor() as cursor:
-            cursor.execute(open("postgres/init.sql").read())
+        with conn.cursor() as cursor, open("postgres/init.sql", encoding="utf-8") as f:
+            cursor.execute(f.read())
 
     def prepare_server(self) -> None:
         """Perform preparation tasks before running the server."""
--- a/manage.py
+++ b/manage.py
@@ -152,36 +152,21 @@ class SiteManager:
         from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
 
         db_url_parts = SiteManager.parse_db_url(os.environ["DATABASE_URL"])
-        db_connection_kwargs = {
-            "host": db_url_parts.hostname,
-            "port": db_url_parts.port,
-            "user": db_url_parts.username,
-            "password": db_url_parts.password,
-        }
-        # Connect to pysite first to create metricity db
         conn = psycopg2.connect(
-            database=db_url_parts.path[1:],
-            **db_connection_kwargs
+            host=db_url_parts.hostname,
+            port=db_url_parts.port,
+            user=db_url_parts.username,
+            password=db_url_parts.password,
+            database=db_url_parts.path[1:]
         )
+        # Required to create a db from `cursor.execute()`
         conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
 
-        with conn.cursor() as cursor:
-            cursor.execute("SELECT 1 FROM pg_catalog.pg_database WHERE datname = 'metricity'")
-            exists = cursor.fetchone()
-            if exists:
-                # Assume metricity is already populated if it exists
-                return
-            print("Creating metricity relations and populating with some data.")
-            cursor.execute("CREATE DATABASE metricity")
-        conn.close()
-
-        # Switch connection to metricity and initialise some data
-        conn = psycopg2.connect(
-            database="metricity",
-            **db_connection_kwargs
-        )
         with conn.cursor() as cursor, open("postgres/init.sql", encoding="utf-8") as f:
-            cursor.execute(f.read())
+            cursor.execute(
+                f.read(),
+                ("metricity", db_url_parts.username, db_url_parts.password)
+            )
         conn.close()
 
     def prepare_server(self) -> None:
--- a/manage.py
+++ b/manage.py
@@ -153,6 +153,8 @@ class SiteManager:
         import psycopg2
         from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
 
+        print("Initialising metricity.")
+
         db_url_parts = SiteManager.parse_db_url(os.environ["DATABASE_URL"])
         conn = psycopg2.connect(
             host=db_url_parts.hostname,
--- a/manage.py
+++ b/manage.py
@@ -173,6 +173,7 @@ class SiteManager:
                 return
             print("Creating metricity relations and populating with some data.")
             cursor.execute("CREATE DATABASE metricity")
+        conn.close()
 
         # Switch connection to metricity and initialise some data
         conn = psycopg2.connect(
@@ -181,6 +182,7 @@ class SiteManager:
         )
         with conn.cursor() as cursor, open("postgres/init.sql", encoding="utf-8") as f:
             cursor.execute(f.read())
+        conn.close()
 
     def prepare_server(self) -> None:
         """Perform preparation tasks before running the server."""
--- a/manage.py
+++ b/manage.py
@@ -142,10 +142,10 @@ class SiteManager:
     @staticmethod
     def run_metricity_init() -> None:
         """
-        Initilise metricity relations and populate with some testing data.
+        Initialise metricity relations and populate with some testing data.
 
         This is done at run time since other projects, like Python bot,
-        rely on the site initilising it's own db, since they do not have
+        rely on the site initialising it's own db, since they do not have
         access to the init.sql file to mount a docker-compose volume.
         """
         import psycopg2
--- a/manage.py
+++ b/manage.py
@@ -129,12 +129,58 @@ class SiteManager:
                 name="pythondiscord.local:8000"
             )
 
+    @staticmethod
+    def run_metricity_init() -> None:
+        """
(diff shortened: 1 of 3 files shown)


manage.py
```diff
@@ -129,12 +129,56 @@ def set_dev_site_name() -> None:
                 name="pythondiscord.local:8000"
             )
 
+    @staticmethod
+    def run_metricity_init() -> None:
+        """
+        Initilise metricity when running in debug mode.
```
doublevcodes:
```suggestion
        Initialise metricity when running in debug mode.
```

manage.py
```diff
@@ -129,12 +129,56 @@ def set_dev_site_name() -> None:
...
+
+        This is done at run time, since other projects (Like Python bot)
+        rely on the site initilising it's own db, rather than relying on
```
doublevcodes:
```suggestion
        rely on the site initialising its own db, rather than relying on
```

jchristgit (approved):
So, I would kind of argue for:

- removing the print if it already exists, to reduce noise
- `open` into a `with` block

but since this fixes the broken local dev at the moment, we should get it in ASAP.

manage.py
```diff
@@ -129,12 +129,57 @@ def set_dev_site_name() -> None:
...
+            cursor.execute("SELECT 1 FROM pg_catalog.pg_database WHERE datname = 'metricity'")
+            exists = cursor.fetchone()
+            if exists:
+                # Assume metricity is already populated if it exists
+                return
+            print("Creating metricity relations and populating with some data.")
+            cursor.execute("CREATE DATABASE metricity")
+
+        # Switch connection to metricity and initialise some data
+        conn = psycopg2.connect(
```
MarkKoz, line 159: Does the previous connection not need to be cleaned up in any way?

manage.py
```diff
@@ -129,12 +129,57 @@ def set_dev_site_name() -> None:
...
+            **db_connection_kwargs
+        )
+        conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
+
+        with conn.cursor() as cursor:
+            cursor.execute("SELECT 1 FROM pg_catalog.pg_database WHERE datname = 'metricity'")
+            exists = cursor.fetchone()
+            if exists:
+                # Assume metricity is already populated if it exists
+                return
```
MarkKoz: Can something be [done in SQL](https://stackoverflow.com/questions/18389124/simulate-create-database-if-not-exists-for-postgresql) to avoid having to switch connections?

manage.py
```diff
@@ -129,12 +129,57 @@ def set_dev_site_name() -> None:
...
+        Initilise metricity relations and populate with some testing data.
+
+        This is done at run time since other projects, like Python bot,
+        rely on the site initilising it's own db, since they do not have
+        access to the init.sql file to mount a docker-compose volume.
+        """
+        import psycopg2
+        from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
+        from urllib.parse import urlsplit
+        # The database URL has already been validated in `wait_for_postgres()`
```
MarkKoz: This is true, but also fragile since nothing is enforcing that `wait_for_postgres()` is called before this function.

manage.py
ChrisLovering: I came across that myself too. The latter part of that answer, that we'd need to use since we're not in shell or psql, became quite complicated. While I don't mind using it, this seems much easier to understand what's going on.

manage.py
ChrisLovering: ah yes, you're right. I had assumed leaving the context manager dealt with this, but it does not.

manage.py
ChrisLovering: Moved to a static method and standardised here https://github.com/python-discord/site/pull/583/commits/661454f9d53508f99a244b4a0ebf34be9b6d3008

manage.py
ChrisLovering: Resolved here https://github.com/python-discord/site/pull/583/commits/6de87e47e313b4f533e8d3adfeb0b8202e938ca0

manage.py
ChrisLovering: Hmm, I might spend a little more time on this, might have a working solution.

ChrisLovering: This should do it, actually much nicer to read the code now, just the sql file is a little more complicated to follow, not much of an issue though.
https://github.com/python-discord/site/pull/583/commits/d245bbd600811762a74544aeff6086a6c94dd7f3

manage.py
```diff
@@ -129,12 +139,43 @@ def set_dev_site_name() -> None:
                 name="pythondiscord.local:8000"
             )
 
+    @staticmethod
+    def run_metricity_init() -> None:
+        """
+        Initilise metricity relations and populate with some testing data.
```
doublevcodes:
```suggestion
        Initialise metricity relations and populate with some testing data.
```

manage.py
```diff
@@ -129,12 +139,43 @@ def set_dev_site_name() -> None:
...
+
+        This is done at run time since other projects, like Python bot,
+        rely on the site initilising it's own db, since they do not have
```
doublevcodes:
```suggestion
        rely on the site initialising it's own db, since they do not have
```

manage.py
```diff
@@ -188,6 +229,12 @@ def run_server(self) -> None:
 
 def main() -> None:
     """Entry point for Django management script."""
+    # Always run metricity init in CI
+    in_ci = os.environ.get("CI", "false").lower() == "true"
+    if in_ci:
```
doublevcodes:
```suggestion
    if os.environ.get("CI", "false").lower() == "true":
```

doublevcodes (changes requested):
Initilise :shipit:

manage.py
```diff
@@ -55,6 +54,20 @@ def __init__(self, args: List[str]):
...
+        """Validate and split the given databse url."""
+        db_url_parts = urlsplit(db_url)
+        if not all((
+            db_url_parts.hostname,
+            db_url_parts.port,
+            db_url_parts.username,
+            db_url_parts.password,
+            db_url_parts.path
+        )):
+            raise OSError("Valid DATABASE_URL environment variable not found.")
```
MarkKoz: This error message is inaccurate and the type of exception is inappropriate. It should be a `ValueError` that states the URL is invalid. A `KeyError` will already be raised by `os.environ["DATABASE_URL"]` if the variable doesn't exist.

postgres/init.sql
```diff
@@ -1,8 +1,25 @@
...
+  ELSE
+    PERFORM dblink_connect(
+        'host=localhost user=' || _user ||
+        ' password=' || _password ||
+        ' dbname=' || current_database()
+    );
+    PERFORM dblink_exec('CREATE DATABASE ' || _db);
+  END IF;
+END
+$do$;
```
MarkKoz, line 24: If you got the code from SO, then you should [attribute it](https://stackoverflow.blog/2009/06/25/attribution-required/) to the author(s).

MarkKoz (changes requested):

postgres/init.sql
MarkKoz: Well maybe [it's complicated](https://meta.stackexchange.com/questions/272956/a-new-code-license-the-mit-this-time-with-attribution-required?cb=1)...

ChrisLovering: Hmm yea that a little contradicting.
Maybe just the below at the top of the sql file will do it?
```
-- The following function is from Stack Overflow
-- https://stackoverflow.com/questions/18389124/simulate-create-database-if-not-exists-for-postgresql/36218838#36218838
-- User frankhommers (https://stackoverflow.com/users/971229/frankhommers)
```

MarkKoz: Yes, probably. I wouldn't worry about it too much as long as we at least tried to attribute :)

manage.py
MarkKoz: Oh it does state its invalid, I missed that. Well, just change OSError then I guess.

MarkKoz: I think it's a bit confusing since it says "not found" at the end.

ChrisLovering: Yea agreed, I copy-pasted it from the previous validation without reading it. Updated it in https://github.com/python-discord/site/pull/583/commits/4411ae4bbffabbf962b5270cd7aaee5b6495ab6e

postgres/init.sql
ChrisLovering: Added in https://github.com/python-discord/site/pull/583/commits/59f4a3d5a66b6d25271337627062be9344bcf285

ChrisLovering: Miss-spelled the user's name, force pushed https://github.com/python-discord/site/pull/583/commits/04ab6e6e10eace3ee103cef61f2ee94f3d2129fb lol

manage.py
```diff
@@ -55,6 +54,22 @@ def __init__(self, args: List[str]):
...
+        db_url_parts = urlsplit(db_url)
+        if not all((
+            db_url_parts.hostname,
+            db_url_parts.port,
+            db_url_parts.username,
+            db_url_parts.password,
+            db_url_parts.path
+        )):
+            raise ValueError(
+                "The DATABASE_URL environment variable is not a valid PostgreSQL database URL."
```
MarkKoz, line 69: The function is technically not restricted to only checking `DATABASE_URL` despite what the error says, but whatever; it's a minor detail.

manage.py
```diff
@@ -129,12 +141,43 @@ def set_dev_site_name() -> None:
                 name="pythondiscord.local:8000"
             )
 
+    @staticmethod
+    def run_metricity_init() -> None:
```
MarkKoz, line 145: Can you add a print somewhere in here to log that it's initialising metricity?

MarkKoz (approved):

PR #586, opened 2021-09-08
URLs: add an allow everything robots.txt

Akarys42:
This commit simply adds a `robots.txt` file from within the URL configuration.

### DIFF
--- a/pydis_site/urls.py
+++ b/pydis_site/urls.py
@@ -1,7 +1,13 @@
 from django.urls import include, path
 
 
+def robots_txt(request: WSGIRequest) -> HttpResponse:
+    """Return an allow everything robots.txtfile."""
+    return HttpResponse("User-agent: *\nAllow: /", headers={"Content-type": "text/plain"})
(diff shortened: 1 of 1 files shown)


Kronifer (approved):
Looks good!

jb3 (changes requested):
I'd rather have the robots.txt content in a multi-line string than with \n, also we shouldn't allow everything, we should disallow /admin/.

Kronifer:
Oop that probably makes sense

jchristgit:
> I'd rather have the robots.txt content in a multi-line string than with \n, also we shouldn't allow everything, we should disallow /admin/.

What's the value in disallowing `/admin/` though?

jchristgit:
@Akarys42 can you address the merge conflict and Joe's comment? It's good to me otherwise.

jchristgit (approved):
Thanks!

Akarys42:
We will be using a worker for this

PR #587, opened 2021-09-09
Create db then switch conn, rather than use dblink

ChrisLovering:
The previous implementation would succesfully create the metricity db, but the tables and rows would be inserted into pysite.

This is because dblink doesn't switch contexts within the script, as psql doesn't support this.

If we were to stick with 1 sql file, we would need to wrap every single statement into a dblink_exec() call to force psql to exec on the other db.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -151,26 +151,53 @@ class SiteManager:
         access to the init.sql file to mount a docker-compose volume.
         """
         import psycopg2
+        from psycopg2 import sql
         from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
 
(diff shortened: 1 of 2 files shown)


MarkKoz:
So, is this not actually an issue?

ChrisLovering:
> So, is this not actually an issue?

Yea, this is an issue, but has been superseded by https://github.com/python-discord/site/pull/588 since this implementation didn't work for people who also run metricity too (Since it runs it's own migrations and conflicts).

See https://canary.discord.com/channels/267624335836053506/635950537262759947/885573792376709161 for an overview of what our solution is

PR #588, opened 2021-09-09
Revert metricity init changes

ChrisLovering:
This moves the init back into a docker-compose volume.

The bot now migrates metricity by actually including metricity in the docker compose, so there is no need for the site to run migrations in the image anymore.

This PR also moves the postgres health check code from site runtime, into docker compose.

This needs to be merged at the same time as https://github.com/python-discord/bot/pull/1820

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,11 @@ services:
       POSTGRES_DB: pysite
       POSTGRES_PASSWORD: pysite
       POSTGRES_USER: pysite
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U pysite"]
+      interval: 2s
+      timeout: 1s
+      retries: 5
     volumes:
       - ./postgres/init.sql:/docker-entrypoint-initdb.d/init.sql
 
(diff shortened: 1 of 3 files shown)


mbaruh:
For context, what was the original reasoning for doing the init outside of the compose? can you link the PR?

ChrisLovering:
> For context, what was the original reasoning for doing the init outside of the compose? can you link the PR?

The original reason was because when we included site in the bot's compose file, we used the image directly, so it did not have metricity initialised. This wasn't much of an issue until recently, when we added django-prometheus which requires those relations exist at run time.

However, this wasn't a great solution, since users who actually want to run metricity needed to delete the relations before bringing up metricity, since it does it's own migrations.

https://github.com/python-discord/site/pull/583

mbaruh (dismissed):
image
Docker screams at me when I try to do anything

mbaruh (approved):
Properly reverts the previous changes and works well with the bot PR 👍

Xithrius (approved):
Yes.

PR #593, opened 2021-09-23
Direct appeals to the appeals server

ChrisLovering:
Rather than mentioning the apeals email address, we now direct users to the appeal server, which will now be used for ban appeals.


ChrisLovering:
In draft until we have a modmail plugin to control pings

Akarys42 (approved):

vcokltfre (approved):
Awesome, happy to see this happening!

PR #596, opened 2021-10-04
Fixes Test Paths For Windows

### LINKED ISSUE #566: Test Failure On Windows
## Description
The current unittests fail on windows, when using the default test command:
> python manage.py test --no-input

## Cause
The following is the failing test: `pydis_site.apps.content.tests.test_views.test_get_context_data_breadcrumbs`
https://github.com/python-discord/site/blob/512bd175cffbc3463679f3e165bea6d0a486f6fa/pydis_site/apps/content/tests/test_views.py#L163-L177

The reason is in the following line:
https://github.com/python-discord/site/blob/512bd175cffbc3463679f3e165bea6d0a486f6fa/pydis_site/apps/content/tests/test_views.py#L175

Now, `context["breadcrumb_items"][2]["path"]` would normally produce `category/subcategory`. Those who have worked with paths before know where this is going, but since we are doing a blind comparison on the string values of paths, we end up with the following assertion error:

`category/subcategory != category\\subcategory`
(left is expected, right is the output on windows)

## Solution
These sort of things wouldn't cause actual problems while running, because windows, and python, can both gracefully handle either format.

One option when it comes to fixing this test, is to pass both ends through a Path call, resulting in:
```python
        context = self.ViewClass.get_context_data()
        context["breadcrumb_items"][2]["path"] = Path(context["breadcrumb_items"][2]["path"])

        self.assertEquals(
            context["breadcrumb_items"],
            [
                {"name": PARSED_CATEGORY_INFO["title"], "path": "."},
                {"name": PARSED_CATEGORY_INFO["title"], "path": "category"},
                {"name": PARSED_CATEGORY_INFO["title"], "path": Path("category/subcategory")},
            ]
        )
```

This should work because we will be comparing two paths objects, instead of pure strings.

This isn't a real solution though, generally speaking. It will solve the problem here, but this sort of thing can slip in at any time. This specific bug appears to have existed since Kosa did the rewrite, but I'm willing to bet it has existed for much longer. Either way, it's been in for quite a while.

Now, one thing we could do is add windows to our testing CI. It's not a fun process from my experience, but I can't think of a better solution right now. Alternatively, we just completely ignore windows, though as a windows user, I have to try and advocate against that :).

I'm open to hearing other solutions.

HassanAbouelela:
Closes #566.

Fixes a test that fails on Windows due to differing path separators. This is currently the only failing test on windows.

Of note is that this doesn't implement all of 566, as I haven't included windows in the CI. Regardless, I'm pushing this now because I don't know when I'll get around to getting the CI working, and that doesn't need to block this specific bug. This will also give more time for discussion around whether Windows in CI is desirable in the first place.


ChrisLovering (approved):

vcokltfre (approved):
I actually understand this one! All looks to be good :ok_hand:

PR #607, opened 2021-10-08
Sir-Lancebot env changes

TizzySaurus:
This PR makes documentation reflect the changes made in [sir-lancebot#896](https://github.com/python-discord/sir-lancebot/pull/896), as well as adding a few missing details such as backquotes.

### DIFF
--- a/pydis_site/apps/content/resources/guides/pydis-guides/contributing/sir-lancebot.md
+++ b/pydis_site/apps/content/resources/guides/pydis-guides/contributing/sir-lancebot.md
@@ -62,7 +62,11 @@ You will need your own test server and bot account on Discord to test your chang
     * `#dev-log`
     * `#sir-lancebot-commands`
 4. Create the following roles:
-    * `@Admin`
+    * `@Owners`
+    * `@Admins`
+    * `@Moderation Team`
+    * `@Helpers`
+    * `@Core Developers`
 5. Note down the IDs for your server, as well as any channels and roles created.
     * [**Learn how to obtain the ID of a server, channel or role here.**](../setting-test-server-and-bot-account#obtain-the-ids)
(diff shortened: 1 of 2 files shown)


wookie184 (commented):
`Admin` should be the only role needed to test sir-lancebot features, so I think that's the only role we should mention on the main page (it makes sense to list the others on the reference page though).

I'd even say `Helpers` is stretching it a bit as what is necessary, although probably fair to mention so one can differentiate between `MODERATION_ROLES` and `STAFF_ROLES` in testing.

TizzySaurus:
> `Admin` should be the only role needed to test sir-lancebot features, so I think that's the only role we should mention on the main page (it makes sense to list the others on the reference page though).
> 
> I'd even say `Helpers` is stretching it a bit as what is necessary, although probably fair to mention so one can differentiate between `MODERATION_ROLES` and `STAFF_ROLES` in testing.

How exactly do you define "needed"? All of these roles are used within different checks etc. so to me that means they're all needed to get the full functionality.

HassanAbouelela:
Most of these roles are not needed to test the functionality of something. Yes they may be used in checks, but do you for instance need to test with mod,admin,etc or, would just admin suffice. Same sort of philosophy as the test server. You don't need all channels, to use the bot, so here is a shorter list of ones you do need. I think that's even more important here since we already have problems with the guides.

Unless you are adding a feature that needs that role specifically, we probably shouldn't try to make everything configurable.

TizzySaurus:
It was decided internally on Discord that this PR isn't something we're after.

We'd like to keep config to a bare minimum to make setting up the bot easier, and since the only strictly necessary role is `Admins`, that should be the only one that's configurable (as is currently the case).

PR #591, opened 2021-09-21
Restructure bot contributing guide

mbaruh:
Partially handles #574 
Restructures the bot guide to focus on getting users "through the door" first, and slowly adding additional information and options.
The guide is built in terms of showing ways to run the bot and how to achieve it, rather than listing all of the information for each configuration and environment variable right away.
That said, the attempt was to still present the reader with the full information they need, and to still be able to use as a quick reference for contributors who are already familiar with the project.

As a style choice, this guide incorporates collapsible elements, which required some CSS and JS additions. This could be resolved with the bulma-collapsible extension, but it doesn't seem to function well inside django templates.

The guide includes an optional config.yml to use with the test server template. This is a temporary measure until we get the bootstrapping application going.

This guide additionally splits the bot account creation and obtaining Discord ID's guides and they didn't seem related. The original file is kept for now as it is used by other guides at the moment.

PR #609, opened 2021-10-09
Static Previews

### LINKED ISSUE #500: Long-term: Deploy previews!
It would be amazing to have deploy previews in our PRs, but this will probably require us to do a whole bunch of very tricky stuff, so it's a really long-term goal.

## Wait, what would we have to do?

- [x] Dewikify the site
- [ ] Migrate the API to a separate app. @SebastiaanZ is already working on that, so in the near future our API will be a separate FastAPI webapp. 
- [ ] Get rid of our database. Once the APIectomy is complete, it will be trivial to remove the database. We don't need `auth` and we don't need `sites` or `contenttypes`.
- [ ] Rewrite this app to be a static-site generator app, using something like [Cactus](https://github.com/eudicots/Cactus), [Pelican](https://docs.getpelican.com/en/latest/) or [Frozen-Flask](https://pythonhosted.org/Frozen-Flask/)
- [ ] Set up Deploy Previews using something like Netlify.

So yeah, _a lot_. 😅 But maybe one day!

HassanAbouelela:
It's finally here! Static previews on the site.

This PR uses [django-distill](https://pypi.org/project/django-distill/) to build a static version of the site, which is then deployed and hosted on netlify. Netlify will give us a static preview of the main branch, and it'll also give us previews on PRs. You can find examples of what that'll look like on my fork here:

https://github.com/HassanAbouelela/site-1/pull/8
This is a static preview of the changes made in #591.

## Important!
1. The netlify site should *not* be set up until this PR is merged. The first run of the action on the main branch won't generate a preview because of that, but it shouldn't matter. I've included instructions for setting up the netlify build, and plan on doing it myself once ready.

2. This PR *will* cause action failures on all other PRs once merged, since the script `netlify-build` will not be available on those branches. It's a really easy fix though, just need to pull main into them.

3. Redirects do not work at the moment, read on below.


HassanAbouelela:
# Redirects
The situation with redirects is a little complex, so I've decided to give it it's own comment here.

I've put a lot of work into redirects (you can see my latest efforts [here](https://github.com/HassanAbouelela/site-1/blob/redirects/pydis_site/apps/redirect/urls.py). You can find those changes deployed to [this](https://615f80944a7c1b0b247418a2--pydis-static.netlify.app/) preview branch.

In short: it really does not work as intended yet, and it's already too complex, and too ugly. I'm now heavily leaning towards abandoning that work since we really don't need redirects on the static previews. The only reason I considered having them in the first place is that we use them in some places on the site, but honestly I'm considering changing up those old links to the place they currently point to.

I want opinions, but this PR should be good to merge regardless.

HassanAbouelela:
I've done some more work on the redirects. The latest changes on the branch I linked in my previous comment are deployed [here](https://616229a3a841fec5fb8bf1db--pydis-static.netlify.app/).

Now, while my changes do work, they still feel sloppy. I did a bit more digging though, and I found a lot of guides and articles still using old links, so I'm not sure what the right call is here.

There's also a question of cost, since we generate ~40 extra HTML files for redirects alone. It's probably not a big deal, but it does take us from 246 to 285 files, so it shouldn't be ignored.

All in all, I'm not sure what to do. Perhaps I'll receive some clarity in the morning? TBD

HassanAbouelela:
Decided to ignore the caching problems, since we'll only save just around 15 seconds anyways. The buildx cache does not play nice with regular docker builds, and I couldn't figure out how to run an image built by buildx. If someone else wants to give it a shot, they are more than welcome to, otherwise, I'm marking this PR as ready.

Edit 1: I've managed to get it working, saving those 15 seconds. It's not too hacky, so I've decided to leave it in.

manage.py
```diff
@@ -155,8 +173,21 @@ def main() -> None:
 
     # Pass any others directly to standard management commands
     else:
+        if _static_build := "distill" in sys.argv[1]:
```
jchristgit: Since we use this variable outside of this block, would it be more clean to define it on a separate line? Just a thought.

pydis_site/apps/content/urls.py
```diff
@@ -1,9 +1,46 @@
...
+from pathlib import Path
+
+from django_distill import distill_path
 
 from . import views
 
 app_name = "content"
+
+
+def __get_all_files(root: Path, folder: typing.Optional[Path] = None) -> list[str]:
```
jchristgit, line 11: Hmm, why the double underscores here? Would it be more clean to use `__all__` to signify what is external?

static-builds/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN poetry install --no-dev
+
+ARG git_sha="development"
+ENV GIT_SHA=$git_sha
+ENV PARENT_HOST=replace_me.host
+
+# Copy the source code in last to optimize rebuilding the image
+COPY . .
+
+RUN SECRET_KEY=dummy_value python manage.py distill-local build --traceback --force --collectstatic
```
jchristgit: Is there no way that we can re-use the main Dockerfile here?

jchristgit (approved):
This is very cool!!!!!!!!!!!! Nice work on the documentation. I have a few smaller comments, but actually they are more like questions. You can deal with them, or you can not. This looks good from an implementation perspective.

static-builds/Dockerfile
HassanAbouelela: I had initially avoided that so we can have each on a separate repo, and utilize cache better, but in it's current state, yeah it can be done.

Changed in 79259eb8

pydis_site/apps/content/urls.py
HassanAbouelela: My intention was to explicitly hide it because while it is technically a utility, it probably shouldn't be used directly from this file. I'm not sure if placing it under all conveys the same intention. It can still be imported and nothing will complain AFAIK.

What do you think?

jchristgit: Yes, that makes sense to me.

jchristgit:
Rock solid work. Thank you very much. Lemon will be very happy about this.

manage.py
HassanAbouelela: Changed in 03ad772d

HassanAbouelela:
Hey @jchristgit I've made some slight changes to the netlify build file. We now use https://nightly.site instead of a github token to increase security in our builds. Functionally, everything is still the same, and tested working on my fork.

jchristgit (approved):
@HassanAbouelela thanks! That's great news. More security 🚢🇮🇹

jb3 (approved):

PR #612, opened 2021-10-10
Adds Redirects To Static Builds

HassanAbouelela:
This PR continues the work done in #609, but adds redirects.

I've left quite a few comments there about redirects, and this is where I ended up now. It's kind of complicated, and not the most useful thing, so I decided to split it into a separate PR to help us decide what to do.


vcokltfre (approved):
With my limited Django knowledge don't take this as "everything is perfect", but rather "I don't see anything glaringly wrong with it" - so yeah, I don't see anything glaringly wrong with it

jchristgit (approved):
This is pretty genius. Thank you!

PR #615, opened 2021-10-17
FAQ: Update reference to `#dev-branding`

hedyhli:
Ref: [message in `#community-meta`](https://discord.com/channels/267624335836053506/429409067623251969/896753201645232169)

I think it should be updated so newer members won't get confused when they don't see a `#media-branding`.


Akarys42 (approved):
lgtm! thanks!

wookie184 (approved):
Thanks :)

PR #619, opened 2021-11-04
Improve Gitpod section in Sir Lancebot contributing guide

hedyhli:
**Add newlines so it's not a big unfriendly paragraph**
When I was reading (skimming) through the guide a while ago I completely missed the part about the fork and environment variables but since I was already familiar with the development process I got it working. I think that spitting them into separate paragraphs is more readable and clear.

**Fix gitpod workspace link**
Two problems:
1. `[<the-link>]()` links to the current page, not `<the-link>`
2. the link is wrong, we need to specify the full github URL not just `user/repo` else it'd show context error

**Clarification on terminal commands**
I'm not quite sure but it makes more sense like this

**Clarification on test server and bot**
We need to make it clear that a test server and bot is still needed when developing on Gitpod, else they wouldn't know what the environment variables are for.

**Add links for that ^ and for environment variables**
So users can find the corresponding guides on the two topics to set them up.


jchristgit (approved):
LGTM. Thanks!

D0rs4n (approved):

PR #621, opened 2021-11-10
Add `dm_sent` field to infractions model

### LINKED ISSUE #610: Add `dm_sent` field to infractions database
Refers to [bot#1864](https://github.com/python-discord/bot/issues/1864).

Make the necessary database changes in [this](https://github.com/python-discord/site/blob/53bbaac0eb515febb74a7a8f97dd9483df9a8568/pydis_site/apps/api/models/bot/infraction.py#L61) file in order to support infractions having a `dm_sent` field in the database which is shown upon `!infraction` command.

**RELEVANT PR TO BE MERGED WITH THOSE OF [bot#1864](https://github.com/python-discord/bot/issues/1864) AND [api#20](https://github.com/python-discord/api/issues/20).**

TizzySaurus:
Closes #610.

Exactly as the title says.

**To be merged with [bot#1951](https://github.com/python-discord/bot/pull/1951).**
Also relevant: [api#28](https://github.com/python-discord/api/pull/28)


Akarys42:
Thanks for the PR! We do not need to merge it with the API as they don't rely on each other. What we do need though is to either add bot side code to handle it at the same time, or have a sensible default value. Based on what you have so far, we should just set the default to `null` (do we even need to?).

TizzySaurus:
> Thanks for the PR! We do not need to merge it with the API as they don't rely on each other.

Will update that :+1:

> What we do need though is to either add bot side code to handle it at the same time, or have a sensible default value. Based on what you have so far, we should just set the default to `null` (do we even need to?).

Yep, the default value will be `null`. I looked through the docs and after a while of figuring out the structure etc. I found out that the default value for `BooleanField` is already `null` (actually `None`), so we don't need to specify anything.
docs

Akarys42 (dismissed):
Along with the serializer, don't forget to also add the new field to the documentation there

https://github.com/python-discord/site/blob/613e0730c5391c33d35106049bb6e8537e644446/pydis_site/apps/api/viewsets/bot/infraction.py#L61-L75

TizzySaurus:
> Along with the serializer, don't forget to also add the new field to the documentation there
> 
> https://github.com/python-discord/site/blob/613e0730c5391c33d35106049bb6e8537e644446/pydis_site/apps/api/viewsets/bot/infraction.py#L61-L75

Have done both :+1:

jchristgit (approved):
Good, concise change 👍

D0rs4n (approved):
Looks good to me, fix up the tests/linting and it's good to go!

jb3 (approved):

PR #622, opened 2021-11-11
Easy docker migrations

ChrisLovering:
This PR adds two things
1. An easy way to generate migrations running under a pure docker setup
   - Simply run `poetry run task dockermigrations` and all the migrations will be created
   - Under the hood this calls docker exec and then through the magic of volumes it appears in the local system
2. Add a persistent volume for postgres data
   - This means `docker-compose down` won't lose your db data anymore.
   - If needed you can still do `docker-compose down --volumes`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,6 +46,8 @@ services:
     volumes:
       - .:/app:ro
       - staticfiles:/var/www/static
+      - ./pydis_site/api/migrations:/app/pydis_site/api/migrations
+      - ./pydis_site/home/migrations:/app/pydis_site/home/migrations
     environment:
       DATABASE_URL: postgres://pysite:pysite@postgres:5432/pysite
       METRICITY_DB_URL: postgres://pysite:pysite@postgres:5432/metricity
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,7 @@ services:
       retries: 5
     volumes:
       - ./postgres/init.sql:/docker-entrypoint-initdb.d/init.sql
+      - pgdata:/var/lib/postgresql/data
 
   web:
     build:
@@ -57,3 +58,4 @@ services:
 
 volumes:
   staticfiles:
+  pgdata:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,7 @@ build-backend = "poetry.core.masonry.api"
 [tool.taskipy.tasks]
 start = "python manage.py run --debug"
 makemigrations = "python manage.py makemigrations"
+dockermigrations = "docker-compose exec web python /app/manage.py makemigrations"
 django_shell = "python manage.py shell"
 test = "coverage run manage.py test"
 report = "coverage report -m"


MarkKoz:
The home app's migrations should also be mounted. This seems to be the only other app that has models.

ChrisLovering:
> The home app's migrations should also be mounted. This seems to be the only other app that has models.

Not sure why I forgot you mentioning other apps having migrations yesterday. I've force pushed a new mount 👍

pyproject.toml
```diff
@@ -49,6 +49,7 @@ build-backend = "poetry.core.masonry.api"
 [tool.taskipy.tasks]
 start = "python manage.py run --debug"
 makemigrations = "python manage.py makemigrations"
+dockermigrations = "docker-compose exec web python /app/manage.py makemigrations"
```
MarkKoz, line 52: This is okay, but it would be nice if it started the service itself if it isn't already running.

MarkKoz (approved):

docker-compose.yml
```diff
@@ -46,6 +47,8 @@ services:
     volumes:
       - .:/app:ro
       - staticfiles:/var/www/static
+      - ./pydis_site/api/migrations:/app/pydis_site/api/migrations
+      - ./pydis_site/home/migrations:/app/pydis_site/home/migrations
```
jchristgit, line 51: How does this interact with filesystem permissions?

I've previously had Docker mess up my local project directory (requiring me to `chown` all of them as root) when volume mounting.

jchristgit (commented):
A question.

docker-compose.yml
ChrisLovering: Ah no idea, I have a windows host and file perms are fine, I'd appreciate you checking if you're on a linux host

pyproject.toml
jchristgit: I agree with Mark here, took me a moment to figure out why the command wasn't doing anything 🙂

docker-compose.yml
jchristgit: Tested with:
```diff
diff --git a/pydis_site/apps/api/models/bot/message.py b/pydis_site/apps/api/models/bot/message.py
index 60e2a553..8261b4fc 100644
--- a/pydis_site/apps/api/models/bot/message.py
+++ b/pydis_site/apps/api/models/bot/message.py
@@ -27,7 +27,7 @@ class Message(ModelReprMixin, models.Model):
     author = models.ForeignKey(
         User,
         on_delete=models.CASCADE,
-        help_text="The author of this message."
+        help_text="The creator of this message."
     )
     channel_id = models.BigIntegerField(
         help_text=(
```

Running migrations:

```py
$ poetry run task dockermigrations
Migrations for 'api':
  pydis_site/apps/api/migrations/0075_auto_20211127_1117.py
    - Alter field author on deletedmessage
  File "/app/manage.py", line 195, in <module>
    main()
  File "/app/manage.py", line 184, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 328, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 369, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 83, in wrapped
    res = handle_func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/commands/makemigrations.py", line 182, in handle
    self.write_migration_files(changes)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/commands/makemigrations.py", line 220, in write_migration_files
    with open(writer.path, "w", encoding='utf-8') as fh:
OSError: [Errno 30] Read-only file system: '/app/pydis_site/apps/api/migrations/0075_auto_20211127_1117.py'
```

In the container, the mountpoint is mounted read-write:

```sh
/dev/mapper/vg0-home on /app/pydis_site/api/migrations type ext4 (rw,relatime,discard)
```

On the host, the mountpoint is moutned read-write:

```sh
/dev/mapper/vg0-home on /home type ext4 (rw,relatime,discard)
```

I do find it a bit unfortunate that this only uses the default "auto" migration names generated by Django. Is there a way to pass this arguments, or even make it mandatory?

pyproject.toml
ChrisLovering: Yea, that makes sense. We have 3 options here `docker-compose up` or `docker-compose up --no-recreate` or `docker-compose up --force-recreate`.

The first checks if there are any changes to pick up, if there are it recreates the containers then attaches to them.
`--no-recreate` attaches to the currently running containers and doesn't do any changes.
`--force-recreate` forces a recreate on all containers then attaches.
I'm leaning more towards `--no-recreate`.

IE for `--no-recreate` this task would become `docker-compose up web -d --no-recreate && docker-compose exec web python /app/manage.py makemigrations`

docker-compose.yml
ChrisLovering: Hmm alright I can investigate this perms issue a little later, I'm a little lost at first glance.
> Is there a way to pass this arguments, or even make it mandatory?

Yea, I can add `--name` on the end of this task, which would make the user to do `poetry run task dockermigrations <name>`. I only did it this was for consistency with the `makemigrations` task.

Forcing a named migration probably is best here.

jchristgit: Fully agree 👍

jchristgit: @ChrisLovering hi, just checking in to see whether you've found out anything new about the permissions issue?

ChrisLovering: Other than a bit of time the evening that you posted this I haven't had a chance to look at this no.

Between my limited knowledge of linux permissions, and lack of time to learn, I'm tempted to mark this PR up for grabs.

MarkKoz: I checked `/proc/mounts` for read-only mounts and `/app` was listed in there:

```
root@70b6714b9032:/app# grep "\sro[\s,]" /proc/mounts
sysfs /sys sysfs ro,nosuid,nodev,noexec,relatime 0 0
cgroup /sys/fs/cgroup cgroup2 ro,nosuid,nodev,noexec,relatime 0 0
/dev/sda3 /app ext4 ro,relatime,discard 0 0
proc /proc/bus proc ro,nosuid,nodev,noexec,relatime 0 0
proc /proc/fs proc ro,nosuid,nodev,noexec,relatime 0 0
proc /proc/irq proc ro,nosuid,nodev,noexec,relatime 0 0
proc /proc/sys proc ro,nosuid,nodev,noexec,relatime 0 0
proc /proc/sysrq-trigger proc ro,nosuid,nodev,noexec,relatime 0 0
tmpfs /proc/asound tmpfs ro,relatime,inode64 0 0
tmpfs /proc/acpi tmpfs ro,relatime,inode64 0 0
tmpfs /proc/scsi tmpfs ro,relatime,inode64 0 0
tmpfs /sys/firmware tmpfs ro,relatime,inode64 0 0
```

MarkKoz: Oh it's cause `.:/app:ro` is in the compose file. I guess it's not possible for a subdirectory to be rw but the parent to be ro?

ChrisLovering: It works on like this on windows, where the more specific volume can be rw

MarkKoz: No... I can manually `touch /app/pydis_site/api/migrations/x.py`.

MarkKoz: I can also write to that directory through Python.

pyproject.toml
MarkKoz: You could just use `docker-compose run --rm web makemigrations`.

jchristgit: @ChrisLovering @MarkKoz was there ever a resolution to this topic?

docker-compose.yml
jchristgit: I'm also pretty sure that a subdirectory can be mounted as read-write. In Kubernetes for instance it's common to have the root directory mounted read-only and then a read-write subdirectory, so that should be fine.

@MarkKoz does this work for you?

MarkKoz: @jchristgit Last time I was debugging this, it was not working. I still have no idea why.

jchristgit: Hold on, I think I just found out why by accident. I found this directory and tried to kill it:

```
(site-_f36zgtP-py3.9) jc@zulu:~/workspace/pydis-site$ rm -rf pydis_site/api/
rm: cannot remove 'pydis_site/api/migrations': Permission denied
```

@ChrisLovering @MarkKoz Are we missing the `apps` in the path here? The mounts from the previous posts suggest it.

MarkKoz: Yes, it does seem like it's missing. Looking at the repo, it should be `pydis_site/apps/api/migrations` However, I am not sure how that affects permissions or why it's missing.

pyproject.toml
MarkKoz: My suggested command should resolve Chris's concerns. I have used this command and it works well. It should also work if there's a container already running, since it will just start its own and remove it when its done.

PR #627, opened 2021-11-22
Don't always calc activity blocks

ChrisLovering:
https://github.com/python-discord/bot/pull/1971 must be merged before this.

We only truly care about how many activity blocks a user has when they have a small number of messages, as we only use this for the voice gate.

If a user has more than say 1k messages, then we really don't need to calculate their activity blocks, as it's quite an expensive query with many messages.

Also, for the tests we only actually care that thee key:value pairs are equal, the order of them isn't actually important.

The naming of `assertCountEqual` is a little misleading, since it actually tests that the first sequence contains the same elements as second, regardless of their order. See https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual

### DIFF
--- a/pydis_site/apps/api/tests/test_users.py
+++ b/pydis_site/apps/api/tests/test_users.py
@@ -428,6 +428,25 @@ class UserMetricityTests(AuthenticatedAPITestCase):
             "activity_blocks": total_blocks
         })
 
+    def test_get_metricity_data_over_1k(self):
+        # Given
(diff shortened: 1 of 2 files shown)


jchristgit (approved):
LGTM.

I added the "do not merge" label to prevent somebody from merging it before the bot PR is merged.

Akarys42 (approved):

PR #628, opened 2021-11-23
Alter message query to leverage index

ChrisLovering:
Previously this query would convert each row to an array just to check if it matched or not. By changing EXCLUDE_CHANNELS to a tuple instead of a list, it doesn't get passed as an array, so we can do a simple NOT IN check.

This will also allow us to add an index with this condition to speed it up further.


Akarys42 (approved):
Let's squeeze that speed fruit. Should it be merged yet?

ChrisLovering:
> Let's squeeze that speed fruit. Should it be merged yet?

yup, it's good to be merged. I've already created the index in the db while perf testing

jchristgit (approved):

PR #632, opened 2021-12-11
Use new approx message count view

ChrisLovering:
We have added a new view to metricity that will keep track of an approximate message count, updating every 10 seconds.

By doing this, we avoid running a query against the whole message table every time we want to get a user's messages.


ChrisLovering:
marking this as needs devops, as we are working out a way to refresh this materialised view on a schedule.

jchristgit (approved):

HassanAbouelela (approved):

PR #636, opened 2021-12-16
Always include metricity message blocks

ChrisLovering:
Thanks to a recent database maintenance (https://pythondiscord.freshstatus.io/incident/139811) querying out metricity message data is far cheaper. So there is no longer a reason to only fetch blocks if the member has a low message count.


onerandomusername (approved):
lgtm

jb3 (approved):

jchristgit (approved):
Thanks!

PR #637, opened 2021-12-30
Fix faulty regex filters in migration 59

mbaruh:
Migration 59 populated the filters table with regex filters, some of which were faulty because the `\b` character wasn't escaped. This is fixed in this PR by making all patterns raw strings.


Akarys42 (approved):
Making *all of them* raw was maybe overkill, but it sure does work haha

ChrisLovering (approved):

mbaruh:
Gotta burn the bugs with fire @Akarys42

D0rs4n (approved):

PR #641, opened 2022-01-10
Add netcup logo to page footer

ChrisLovering:
netcup are now sponsoring us with servers! This PR adds netcup to both the footer and the sponsers list on the homepage.

We still have credit with linode, so they will be staying for now.

I have also updated some of the css to make the logos line up a bit better, and move the django footer image to an svg.


jb3 (approved):
LGTM! preview looks great

HassanAbouelela (approved):

PR #644, opened 2022-01-23
Return random off topic names which are Active only

RohanJnr:
### Background
- PR #508 added the `active` attribute to the `off topic channel names` table.
- It also added the ability to send PATCH and PUT requests to modify the instance attributes.

### Changes
- This PR filters the off topic channels names and only sends the `active names` as response to an incoming `GET` request to `bot/off-topic-channel-names?random_items=x`
- Test cases were modified which were relevant to the above change.


jchristgit (approved):
Thank you Rohan! Concise and tested. Excellent!

Xithrius (commented):
Looks good, but I am not confident in my abilities on site. We shall wait for another person approves.

Xithrius (approved):
yep

PR #646, opened 2022-01-26
Fix pynews config in example file

ChrisLovering:
The python news cog relies on the python_news section of the config being filled in via a named alias. Since we were using the `DEV_` prefix, this meant it was not being populated, leading to the cog throwing an error on boot.

This PR also adds the new incidents webhook that is used for message unfurling.


ToxicKidz (approved):
LGTM

minalike (approved):
Can confirm these changes resolved errors when using the `config.yml` provided on the public contributing guide.

PR #647, opened 2022-01-27
Use voice_mute and voice_ban to determine voice_gate

ChrisLovering:
Previously only voice_ban was used, but now having either of these infractions should mean the user is blocked from reciving the role.


lemonsaurus (approved):
This is quite an excellent PR, and I can't find any problems with it.

jchristgit (approved):

PR #653, opened 2022-02-02
Google Colab resource: Rewrite to emphasize GPU computation

swfarnsworth:
Previously, it was unclear why this was a data science resource. This version states that Colab offers GPU computation and that it can be used for deep learning, which is the reason. Also fixes abrupt change in sentence structure ("a ... Jupyter Notebook which supports e.g. !apt-get"; "e.g." unexpected).

### DIFF
--- a/pydis_site/apps/resources/resources/google_colab.yaml
+++ b/pydis_site/apps/resources/resources/google_colab.yaml
@@ -1,7 +1,7 @@
-description: Google Colab is a high-powered custom version of Jupyter Notebook which supports e.g.
-  !apt-get to install arbitrary Debian packages to the runtime, which is very generous with CPU and memory,
-  and well-integrated with Google Drive.
-  You can share your Colab Notebooks with other people and work collaboratively.
+description: Google Colab is a custom version of Jupyter Notebook that runs code in the cloud, with support for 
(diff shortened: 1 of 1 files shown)


ChrisLovering (approved):

pydis_site/apps/resources/resources/google_colab.yaml
```diff
@@ -1,7 +1,7 @@
-description: Google Colab is a high-powered custom version of Jupyter Notebook which supports e.g.
-  !apt-get to install arbitrary Debian packages to the runtime, which is very generous with CPU and memory,
-  and well-integrated with Google Drive.
-  You can share your Colab Notebooks with other people and work collaboratively.
+description: Google Colab is a custom version of Jupyter Notebook that runs code in the cloud, with support for 
+  Debian package installation with `apt-get`. You can share your Colab Notebooks with other people and work collaboratively.
```
ChrisLovering: Actually, just noticed this \`apt-get` won't format it as code, not sure if we have another way to do this.

swfarnsworth: We can just delete that part and say "support for Debian packages".

PR #654, opened 2022-02-04
Replace FA pro issue icon with the regular icon

ChrisLovering:
We stopped using FA pro, as we were using an ex-admin's person FA pro subscription, which we didn't control.

Before:
image

After:
image


lemonsaurus (approved):
Hi, my name is Leon Sandøy and I approve of this code.

D0rs4n (approved):

PR #660, opened 2022-02-13
Smarter Resources - Search bar!

### LINKED ISSUE #659: Smarter Resources: Search bar
Let's create a reactive search bar that can be used to filter the resources on the /resources page.

We can build this with some fairly simple jQuery, `.filter`ing all currently visible resource cards. This will extend the current functionality very naturally.

## Specs
- The search bar should do a search for every keypress.
- The search should be fuzzy. `aks` should match `ask`.
- When typing a search term, a `Search: <search term>` tag should appear in the tag pool at the top of the filter sidebar.
- Pressing enter in the search box does nothing.
- Searching also needs to update the URL, just like filters. That way, we'll be able to link any search result.

I plan to implement this change myself.

lemonsaurus:
Implements a search feature on the resources page
search final

You can test it on the deploy preview!
https://deploy-preview-660--pydis-static.netlify.app/

## Features
- Uses [fuzzysort](https://github.com/farzher/fuzzysort) to do fuzzy matching against resource titles.
- Fully reactive! ✨ Every keypress triggers a search.
- Updates the URL to match the search query, and loads the search on page load when a `?search=` parameter is provided.
   live updating query
- Will match even if the user has a typo in the search query.
   image
- Adds a tag to the filter tag pool whenever a search is live.
   search tag

## Kaizen
- Fixes redirects to /pages/resources/communities, these are used in the `!guild` tag on the bot.
- Ensures that resource cards are always fullwidth. Previously they would only fill the width needed for the longest description of the visible resource cards.
- Update the URL for Two Scoops of Django. The previous one was broken.


pydis_site/templates/resources/resource_box.html
```diff
@@ -2,7 +2,7 @@
 {% load to_kebabcase %}
 {% load get_category_icon %}
 
-<div class="box resource-box {{ resource.css_classes }}">
+<div class="box resource-box {{ resource.css_classes }}" name="{{ resource.name }}">
```
jchristgit: Shouldn't we use something like `x-data-name` instead?

pydis_site/templates/resources/resources.html
```diff
@@ -16,6 +16,7 @@
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.6.0/jquery.min.js"></script>
     <script defer src="{% static "js/resources/resources.js" %}"></script>
     <script defer src="{% static "js/collapsibles.js" %}"></script>
+    <script src="https://rawgit.com/farzher/fuzzysort/master/fuzzysort.js"></script>
```
jchristgit: Can we add this to our staticfiles instead, or pin it to a ref?

lemonsaurus: Resolved by [f5f2310](https://github.com/python-discord/site/pull/660/commits/f5f231078ab6f6de6d0161372d00c57f26acba9e)

pydis_site/templates/resources/resource_box.html
lemonsaurus: Resolved by [33e8f0c](https://github.com/python-discord/site/pull/660/commits/33e8f0cb477547bbef7564a83353d2379caa9e0e)

jchristgit (approved):
This is great lemon. Stunning work on the Smarter Resources. I don't understand anything about JavaScript or CSS or HTML for that matter, but I think this is great.

whispersofthedawn (approved):
This looks wonderful! I don't know much about Django, but from what I do know about JS, that all looks great. The search experience is flawless, too.

pydis_site/apps/resources/resources/python_discord_videos.yaml
```diff
@@ -1,3 +1,4 @@
+name: Python Discord YouTube channel
```
MarkKoz: It's a title, right? Then this needs to be title case
```suggestion
name: Python Discord YouTube Channel
```

pydis_site/templates/resources/resources.html
```diff
@@ -27,18 +28,33 @@
             <div class="column filtering-column is-one-third">
                 <div class="content is-justify-content-center">
                     <nav id="resource-filtering-panel" class="panel is-primary">
-                        <p class="panel-heading has-text-centered" id="filter-panel-header">Filter Resources</p>
+                        <p class="panel-heading has-text-centered" id="filter-panel-header">Filter resources</p>
```
MarkKoz, line 31:
```suggestion
                        <p class="panel-heading has-text-centered" id="filter-panel-header">Filter Resources</p>
```

MarkKoz (changes requested):
The close button doesn't look great when the bar is really wide
bild

I think this is actually a problem with the bar's width, as it shouldn't get that wide in the first place. It seems to try to match the width of the line below it, but when a 3rd line gets added, it gets too wide:

bild

bild

pydis_site/templates/resources/resources.html
lemonsaurus: I feel strongly that this should not be titlecase, but if more people feel otherwise I'll change it.

MarkKoz (approved):
I've gone over the code and played around with it on netlify. Looks good!

PR #663, opened 2022-02-19
Add support for BumpedThreads to be stored in site

ChrisLovering:
Following our move to use Redis as just a cache, this PR allows the site to store a list of threads that need to be bumped.

The bot will interact with this within the ThreadBumper cog.

This also moves the imports of FilterList around, so that they're alphabetised.


jchristgit (approved):
Thanks!

pydis_site/apps/api/models/bot/bumped_thread.py
```diff
@@ -0,0 +1,22 @@
+from django.core.validators import MinValueValidator
+from django.db import models
+
+from pydis_site.apps.api.models.mixins import ModelReprMixin
+
+
+class BumpedThread(ModelReprMixin, models.Model):
+    """A list of thread IDs to be bumped."""
+
+    thread_id = models.BigIntegerField(
```
MarkKoz, line 10: Would it make more sense to use a pg array instead, or do frequent membership checks also need to happen?

pydis_site/apps/api/viewsets/bot/bumped_thread.py
```diff
@@ -0,0 +1,65 @@
...
+    ### GET /bot/bumped-threads
+    Returns all BumpedThread items in the database.
+
+    #### Response format
+    >>> [
+    ...     {
+    ...         'thread_id': "941705627405811793",
+    ...     },
+    ...     ...
+    ... ]
```
MarkKoz: This works, but it's clumsy. It should be flattened to an array. And why is the ID returned as a string? Or is that just a documentation error?

pydis_site/apps/api/viewsets/bot/bumped_thread.py
```diff
@@ -0,0 +1,65 @@
...
+    - 200: returned on success
+    - 401: returned if unauthenticated
+
+    ### GET /bot/bumped-threads/<thread_id:int>
+    Returns a specific BumpedThread item from the database.
+
+    #### Response format
+    >>> {
+    ...     'thread_id': "941705627405811793",
+    ... }
```
MarkKoz: Again, it's clumsy. Just return the value directly (that's still valid JSON by the way). There is only ever going to be one field, so there's no point in having an object with a field name.

pydis_site/apps/api/viewsets/bot/bumped_thread.py
```diff
@@ -0,0 +1,65 @@
...
+    - 200: returned on success
+    - 404: returned if a BumpedThread with the given thread_id was not found.
+
+    ### POST /bot/bumped-threads
+    Adds a single BumpedThread item to the database.
+
+    #### Request body
+    >>> {
+    ...    'thread_id': int,
+    ... }
```
MarkKoz, line 42: Similar to the GET request, it should be possible to just directly send the integer value. Hell, we could even avoid JSON and do url-encoded params or maybe `POST /bot/bumped-threads/<thread_id:int>` if DRF supports that for post requests and it's not deemed unidiomatic.

MarkKoz (changes requested):

pydis_site/apps/api/viewsets/bot/bumped_thread.py
ChrisLovering: Good point, I'll flatten it. Yea, it's a doc error and will fix.

Would this be the best approach to do so? My django knowledge is limited atm

```py
    def list(self, request: HttpRequest, *args, **kwargs) -> list[int]:
        """Returns all BumpedThread items in the database as a flat list."""
        queryset = self.queryset.values_list("thread_id", flat=True)
        return Response(queryset)
```

pydis_site/apps/api/models/bot/bumped_thread.py
ChrisLovering: Yea, membership checks will happen quite often to check when a thread is archived if it's in the bump list.

pydis_site/apps/api/viewsets/bot/bumped_thread.py
MarkKoz: My first thought is to look into customising the serialiser.

jchristgit: Is it possible that we may add a field here in the future, such as `'last_bumped_at'` or `'bumped_by`?

pydis_site/apps/api/viewsets/bot/bumped_thread.py
ChrisLovering: https://www.django-rest-framework.org/api-guide/routers/#simplerouter suggests that POST isn't supported with a lookup path `POST /bot/bumped-threads/<thread_id:int>` only PUT or PATCH would be.

However, query params would work. I'm not sure how much better this is over just using request body though since it would be adding `?thread_id=int` to the url instead of adding `{thread_id: int}` to the request body

MarkKoz: Never mind then. In practice, it doesn't matter since for aiohttp a dictionary needs to be passed either way (unless we put it as a string in the url which can be kinda yucky)

ChrisLovering:
Force push was to rebase onto main. Only conflict was in the migration, so I updated it to depend on the latest in main.

pydis_site/apps/api/viewsets/bot/bumped_thread.py
ChrisLovering: https://github.com/python-discord/site/pull/663/commits/3e9557056c06a39c077b76d718eb35b99a365711

pydis_site/apps/api/viewsets/bot/bumped_thread.py
ChrisLovering: https://github.com/python-discord/site/pull/663/commits/a0180619e77d884aeedb1a89748c0115c8ec8c56

ChrisLovering: Ah, this commit will require tests since I added a custom retrieve, will fix

pydis_site/apps/api/serializers.py
```diff
@@ -41,6 +42,33 @@ class Meta:
...
 
+class ListBumpedThreadSerializer(ListSerializer):
+    """Custom ListSerializer to override to_representation() when list views are triggered."""
+
+    def to_representation(self, objects: list[BumpedThread]) -> int:
+        """
+        Used by the `ListModelMixin` to return just the list of bumped thread ids.
+
+        We want to only return the thread_id attribute, hence it is unnecessary
+        to create a nested dictionary.
```
MarkKoz: Please avoid using "we", "us", "I", etc. in documentation.
```suggestion
        Only the thread_id field is useful, hence it is unnecessary to create a nested dictionary.
```

MarkKoz (changes requested):
I think your original intention was to only test the custom serialisers, but it seems like you're testing other methods as well. If you were trying to test everything, then you have not covered all methods.

ChrisLovering:
> I think your original intention was to only test the custom serialisers, but it seems like you're testing other methods as well. If you were trying to test everything, then you have not covered all methods.

I'm not sure I follow, I tested unauthed requests for each endpoint. Then I also have a test for all the custom stuff I added, the list serializer and the custom retrieve method I added.

MarkKoz:
> > I think your original intention was to only test the custom serialisers, but it seems like you're testing other methods as well. If you were trying to test everything, then you have not covered all methods.
> 
> I'm not sure I follow, I tested unauthed requests for each endpoint. Then I also have a test for all the custom stuff I added, the list serializer and the custom retrieve method I added.

There is no authenticated test for POST (create).

jchristgit:
Mark, I place my trust in you putting this into master, so I assigned you.

ChrisLovering:
> > > I think your original intention was to only test the custom serialisers, but it seems like you're testing other methods as well. If you were trying to test everything, then you have not covered all methods.
> > 
> > 
> > I'm not sure I follow, I tested unauthed requests for each endpoint. Then I also have a test for all the custom stuff I added, the list serializer and the custom retrieve method I added.
> 
> There is no authenticated test for POST (create).

Yea, I didn't bother for create and delete, since both are using the normal serializer and the built-in mixins from DRF.

MarkKoz:
Okay I guess this is ready. I will test this tomorrow as a sanity check and then it's good to go.

MarkKoz (approved):

PR #666, opened 2022-02-20
Prevent race condition with duplicate infractions

### LINKED ISSUE #665: IntegrityError: duplicate key value violates unique constraint "unique_active_infraction_per_type_per_user"
Sentry Issue: [SITE-22](https://sentry.io/organizations/python-discord/issues/2889692012/?referrer=github_integration)

```
UniqueViolation: duplicate key value violates unique constraint "unique_active_infraction_per_type_per_user"
DETAIL:  Key (user_id, type)=(938288196100423800, ban) already exists.

  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)

IntegrityError: duplicate key value violates unique constraint "unique_active_infraction_per_type_per_user"
DETAIL:  Key (user_id, type)=(938288196100423800, ban) already exists.

(21 additional frame(s) were not displayed)
...
  File "django/db/backends/utils.py", line 66, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "django/db/backends/utils.py", line 75, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "django/db/utils.py", line 90, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
```

jchristgit:
DRF's `UniqueTogetherValidator` validates uniqueness by querying the
database before running the actual insert. This is not, has not, and
will never be valid, unless you happen to run a single worker, on a
single thread, and your single worker running on a single thread is the
only client for the database, in which case it may be valid. For any
other cases, it's invalid, and it has never been valid. PostgreSQL spits
out an `IntegrityError` for us if we have a duplicate entry, and
PostgreSQL is the only valid and correct thing to trust here.

The `UniqueTogetherValidator` is removed, and an existing test case
calling into this validator to check for uniqueness is removed.
Furthermore, to work around a Django quirk, `transaction.atomic()` is
added to prevent one `subTest` from messing with another.

Closes #665.

### DIFF
--- a/pydis_site/apps/api/viewsets/bot/infraction.py
+++ b/pydis_site/apps/api/viewsets/bot/infraction.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 
+from django.db import IntegrityError
 from django.db.models import QuerySet
 from django.http.request import HttpRequest
 from django_filters.rest_framework import DjangoFilterBackend
@@ -271,3 +272,20 @@ class InfractionViewSet(
         """
         self.serializer_class = ExpandedInfractionSerializer
         return self.partial_update(*args, **kwargs)
+
+    def create(self, request: HttpRequest, *args, **kwargs) -> Response:
+        """
+        Create an infraction for a target user.
+
+        Called by the Django Rest Framework in response to the corresponding HTTP request.
+        """
+        try:
+            return super().create(request, *args, **kwargs)
+        except IntegrityError:
+            raise ValidationError(
+                {
+                    'non_field_errors': [
+                        'This user already has an active infraction of this type.',
+                    ]
+                }
+            )
--- a/pydis_site/apps/api/serializers.py
+++ b/pydis_site/apps/api/serializers.py
@@ -156,13 +156,6 @@ class InfractionSerializer(ModelSerializer):
             'hidden',
             'dm_sent'
         )
-        validators = [
-            UniqueTogetherValidator(
-                queryset=Infraction.objects.filter(active=True),
-                fields=['user', 'type', 'active'],
-                message='This user already has an active infraction of this type.',
-            )
-        ]
 
     def validate(self, attrs: dict) -> dict:
         """Validate data constraints for the given data and abort if it is invalid."""
--- a/pydis_site/apps/api/tests/test_infractions.py
+++ b/pydis_site/apps/api/tests/test_infractions.py
@@ -3,6 +3,7 @@ from datetime import datetime as dt, timedelta, timezone
 from unittest.mock import patch
 from urllib.parse import quote
 
+from django.db import transaction
 from django.db.utils import IntegrityError
 from django.urls import reverse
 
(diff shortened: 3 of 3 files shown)


ChrisLovering (approved):
Thank you Johannes, this is great. I notice the `FilterListSerializer` also uses the `UniqueTogetherValidator`, due to your PR description I assume this should be changed (in a separate PR) too?

jchristgit:
> Thank you Johannes, this is great. I notice the `FilterListSerializer` also uses the `UniqueTogetherValidator`, due to your PR description I assume this should be changed (in a separate PR) too?

Yes, although over there it's less critical as it's not edited as often as infractions, is it?

ChrisLovering:
> > Thank you Johannes, this is great. I notice the `FilterListSerializer` also uses the `UniqueTogetherValidator`, due to your PR description I assume this should be changed (in a separate PR) too?
> 
> Yes, although over there it's less critical as it's not edited as often as infractions, is it?

Yea, it's definitely not as critical. It's quite rare for two requests to add the same item to the filter list would be sent close enough together to trigger this race condition.

pydis_site/apps/api/viewsets/bot/infraction.py
```diff
@@ -271,3 +272,20 @@ def partial_update_expanded(self, *args, **kwargs) -> Response:
...
+
+        Called by the Django Rest Framework in response to the corresponding HTTP request.
+        """
+        try:
+            return super().create(request, *args, **kwargs)
+        except IntegrityError:
+            raise ValidationError(
+                {
+                    'non_field_errors': [
+                        'This user already has an active infraction of this type.',
```
MarkKoz, line 288: Is it safe to assume all `IntegrityError`s here will be due to a duplicate infraction? Isn't an `IntegrityError` a fairly generic error?

ChrisLovering: It'll be whenever the relational integrity of the database fails. EG duplicate key, foreign key check or any other check added to the table.

So IG this could also be raised if a foreign key being inserted doesn't exist in the related table.

jchristgit: I thought about checking the constraint name here, but (apart from forcing an insert of an ID, which we don't do), we only have a single unique constraint here

MarkKoz: It's not too troublesome to check the constraint name, right? Do you think it'd be better to do this so this code is "future proof" against future constraints we may add? Or am I being overly cautious?

jchristgit: The thing is that the constraint name isn't directly included as an attribute on the error, so we'd need to either just do an "in" check on the string or parse it ourselves, and I didn't quite like that approach...

MarkKoz (approved):
Excellent. Thank you for teaching me something new about race conditions with databases. Great catch with the test transaction too!

pydis_site/apps/api/viewsets/bot/infraction.py
MarkKoz: Yeah, screw it. This is fine.

PR #668, opened 2022-02-20
Update Name & Profile Policy

minalike:
Following today's admin meeting, this PR updates our rules to better reflect our current moderation practices in regards to user profiles. The "Nickname Policy" has also been renamed to "Name & Profile Policy."


ChrisLovering:
It looks like the numbering is messed up.
image
If you add an extra blank line before the first number it should fix it.

ChrisLovering (approved):

PR #669, opened 2022-02-21
Assert we're dealing with the unique constraint

jchristgit:


ChrisLovering (approved):
Nice, looks good 👍

pydis_site/apps/api/viewsets/bot/infraction.py
```diff
@@ -281,11 +281,19 @@ def create(self, request: HttpRequest, *args, **kwargs) -> Response:
...
-                }
-            )
+        except IntegrityError as err:
+            # We need to use `__cause__` here, as Django reraises the internal
+            # UniqueViolation emitted by psycopg2 (which contains the attribute
+            # that we actually need)
+            #
+            # _meta is documented and mainly named that way to prevent
+            # name clashes: https://docs.djangoproject.com/en/dev/ref/models/meta/
+            if err.__cause__.diag.constraint_name == Infraction._meta.constraints[0].name:
```
MarkKoz, line 291: Relying on the position of the constraint seems fragile, especially without test cases for this. What do you think about assigning the constraint name to a constant variable, so we can avoid relying on position? I'm not sure if Django's metadata system supports accessing or even defining arbitrary attributes in the `Meta` class.

MarkKoz (approved):
Nice

jchristgit:
Thanks

PR #682, opened 2022-02-24
Add channel ID to message deletion logs

### LINKED ISSUE #620: Show channel info in message logs
Currently when messages are deleted by the anti-spam filter or by the clean command, a message log is created with info of of the user who sent it, at what time, and the contents of each message.

With recent developments, these two features now work across multiple channels, but the logs don't include that information making them sometimes hard to follow. 

Including the channel ID for each message seems to make sense, but at the same time we navigate channels using their names. Perhaps we include both?

TizzySaurus:
Closes #620.

The channel ID is now shown in message deletion logs.

Zoomed in on one log message:
image

Fullscreen page:
image

### DIFF
--- a/pydis_site/templates/staff/logs.html
+++ b/pydis_site/templates/staff/logs.html
@@ -18,8 +18,12 @@
         <div class="discord-message">
             <div class="discord-message-header">
                 <span class="discord-username"
-                      style="color: {{ message.author.top_role.colour | hex_colour }}">{{ message.author }}</span><span
-                    class="discord-message-metadata has-text-grey">{{ message.timestamp }} | User ID: {{ message.author.id }}</span>
+                      style="color: {{ message.author.top_role.colour | hex_colour }}">{{ message.author }}
+                </span>
+                <span class="discord-message-metadata has-text-grey">
+                    User ID: {{ message.author.id }}<br>
+                    {{ message.timestamp }} (Channel ID: {{ message.channel_id }})
+                </span>
             </div>
             <div class="discord-message-content">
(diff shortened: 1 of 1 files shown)


jchristgit (approved):
Thanks!

minalike (commented):
Tested and it works fine and renders as you'd expect, but semantically wouldn't it make sense to not have line break within a `span` element, and rather split it into 2 `span`s?

One for the user ID, a line break, and a separate one for the line after with the timestamp and channel ID. 

I would suggest:

```html
            <div class="discord-message-header">
                <span class="discord-username" style="color: {{ message.author.top_role.colour | hex_colour }}">
                      {{ message.author }}
                </span>
                <span class="discord-message-metadata has-text-grey">
                    User ID: {{ message.author.id }}
                </span>
                <br>
                <span class="discord-message-metadata has-text-grey">
                    {{ message.timestamp }} (Channel ID: {{ message.channel_id }})
                </span>
            </div>
```

and in `logs.css` adjust the margin so it's only present on the right:

```css
.discord-message-metadata {
    font-size: 0.75rem;
    font-weight: 400;
    margin: 0 .3rem 0 0;
}
```

But I realize this may be overkill, and there may also be a better way of doing this.

MarkKoz:
On the principle of semantics, I agree with mina. A `span` is generally not seen as something that is multi-line. However, in practice, I am not so sure that it matters. If I were the one to implement this, I would likely follow mina's suggestion. However, I am overall indifferent, so I would leave this decision to the author's discretion.

minalike (approved):

jchristgit:
Thanks!

PR #686, opened 2022-03-03
Add a README for the API app

jchristgit:


pydis_site/apps/api/README.md
```diff
@@ -0,0 +1,69 @@
...
+guide](https://www.django-rest-framework.org/tutorial/quickstart/) is a great
+resource to get started.
+
+## Directory structure
+
+Let's look over each of the subdirectories here:
+
+- `migrations` is the standard Django migrations folder. You usually won't need
+  to edit this manually, as `python manage.py makemigrations` handles this for
+  you in case you change our models.
```
janine9vn: Would it be worth mentioning here our policy of renaming migrations to be human-readable/more informative? That might be diving too deep into our policy.

janine9vn (approved):
Fantastic and helps explain how our `api` folder is setup with wonderful resources as well. 

I have a small comment/note, but otherwise this makes it much more approachable with clear points/resources that people can run with and learn more about if they're feeling a bit lost.

MarkKoz (approved):
Maybe the migrations and models could use some more explanation (i.e. what those terms actually mean), but it could be covered elsewhere in a more general area I suppose.

pydis_site/apps/api/README.md
jchristgit: Yes!!!!!!!!!!!!

MarkKoz: Is this the only app for which migrations are relevant? If not, we may want to have a more general documentation page to explain this, which ties into my review comment.

jchristgit: I added this in 073e09a0. Thanks for the review!

jchristgit:
> Maybe the migrations and models could use some more explanation (i.e. what those terms actually mean), but it could be covered elsewhere in a more general area I suppose.

For what it's worth, I don't think we should need to explain what these mean. We require basic Django knowledge, and if you've worked through the tutorial, you know what these mean. I don't want to duplicate Django documentation here either.

jchristgit:
Thanks!

PR #689, opened 2022-03-07
Add support for storing AoC related data in site

D0rs4n:
## Summary

I added support for storing AoC related data in site

## Description of changes

Added two new endpoints, with relating ViewSets and models

### I confirm I have:

- [x] Joined the [Python Discord community](discord.gg/python)
- [x] Read the [Code of Conduct](https://www.pydis.com/pages/code-of-conduct) and agree to it
- [x] I have discussed implementing this feature on the relevant service (Discord, GitHub, etc.)


pydis_site/apps/api/models/bot/aoc_completionist_block.py
```diff
@@ -0,0 +1,21 @@
...
+        on_delete=models.CASCADE,
+        help_text="The user that is blocked from getting the AoC Completionist Role"
+    )
+
+    is_blocked = models.BooleanField(
+        default=True,
+        help_text="Whether this user is actively being blocked "
+                  "from getting the AoC Completionist Role",
+        verbose_name="Blocked"
+    )
```
jchristgit, line 22: Wouldn't every user in this table be blocked? Should we add / use a reason field instead?

pydis_site/apps/api/viewsets/bot/aoc_completionist_block.py
```diff
@@ -0,0 +1,69 @@
...
+    - 404: returned if an AoC completionist block with the given user__id was not found.
+
+    ### POST /bot/aoc-completionist-blocks
+    Adds a single AoC completionist block
+
+    #### Request body
+    >>> {
+    ... 'user': int,
+    ... 'is_blocked': bool
+    ... }
```
jchristgit: nit:
```suggestion
    >>> {
    ...     'user': int,
    ...     'is_blocked': bool
    ... }
```

pydis_site/apps/api/viewsets/bot/aoc_link.py
```diff
@@ -0,0 +1,69 @@
...
+    - 200: returned on success
+    - 404: returned if an AoC account link with the given user__id was not found.
+
+    ### POST /bot/aoc-account-links
+    Adds a single AoC account link block
+
+    #### Request body
+    >>> {
+    ... 'user': int,
+    ... 'aoc_username': str
```
jchristgit: Nit:
```suggestion
    ...     'user': int,
    ...     'aoc_username': str
```

jchristgit (approved):
One minor question and a few nitpicks. The question is probably not as useful. Thanks!

pydis_site/apps/api/models/bot/aoc_completionist_block.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class AocCompletionistBlock(ModelReprMixin, models.Model):
+    """A Discord user blocked from getting the AoC completionist Role."""
+
+    user = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        help_text="The user that is blocked from getting the AoC Completionist Role"
+    )
```
MarkKoz: This should probably be a `OneToOneField` instead. A many `AocCompletionistBlock`-to-one `User` relationship does not make sense since each user can only be blocked once.

pydis_site/apps/api/models/bot/aoc_completionist_block.py
MarkKoz: Yeah, I think it can be assumed the user is blocked by virtue of just existing in this table. Admittedly, it is kind of strange to have a table with just one field but on the other hand `is_blocked` seems redundant.

pydis_site/apps/api/models/bot/aoc_link.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class AocAccountLink(ModelReprMixin, models.Model):
+    """An AoC account link for a Discord User."""
+
+    user = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        help_text="The user that is blocked from getting the AoC Completionist Role"
+    )
```
MarkKoz: This should be a `OneToOneField`. I realise this is using the Discord User model rather than the Django user model, but I believe this is still applicable https://docs.djangoproject.com/en/3.2/topics/auth/customizing/#extending-the-existing-user-model

pydis_site/apps/api/models/bot/aoc_link.py
```diff
@@ -0,0 +1,20 @@
+from django.db import models
+
+from pydis_site.apps.api.models.bot.user import User
+from pydis_site.apps.api.models.mixins import ModelReprMixin
+
+
+class AocAccountLink(ModelReprMixin, models.Model):
+    """An AoC account link for a Discord User."""
```
MarkKoz, line 8: Can only linked users be blocked from being a completionist? Is that separate from this table because we need to be able to keep a user blocked if they unlink their account?

If that is the case, it may be better to merge the tables, and instead of deleting the record when the user unlinks, just set `aoc_username` to null/empty. This way, the block will remain for unlinked users. What are everyone's thoughts on this alternative?

MarkKoz (changes requested):

pydis_site/apps/api/models/bot/aoc_completionist_block.py
D0rs4n: I see the concern, however there is no reason provided on the bot side, as a golden mean I could implement a deactivate block command which sets the `is_blocked` field to False whilst keeping a record of users who have been blocked previously.

pydis_site/apps/api/models/bot/aoc_completionist_block.py
D0rs4n: Oh, yes. I didn't catch it. Thanks!

pydis_site/apps/api/models/bot/aoc_link.py
D0rs4n: No, I don't believe so, this way Admins can block users even if they haven't linked their AoC account.
Otherwise, they would be limited to wait until they link their account, and then block them.
Update: Sorry, I see your alternative now. I was trying to keep the tables as close to the code as possible, as they are two unlinked commands

pydis_site/apps/api/models/bot/aoc_completionist_block.py
janine9vn: It might make sense to keep the column in case we need to unblock someone. Altering a value in a row is ever so slightly easier than deleting a row. I also wouldn't be opposed to a reason field, so there's some context around it.

D0rs4n: Alright, that might need some modification on the bot site, although.
But, minor ones, only.

D0rs4n: Added in b93dce5

jchristgit (approved):
⭐⭐⭐⭐⭐⭐

pydis_site/apps/api/models/bot/aoc_completionist_block.py
MarkKoz: If we want to keep a reason around, then the `is_blocked` field makes more sense. However, this is starting to look more like an infraction at this point.

D0rs4n: It might look like one, but admins might not want to block someone and give them an infraction the same time.
Especially, because that's on Events Bot, (used to be on Sir-Lanebot)

pydis_site/apps/api/models/bot/aoc_link.py
MarkKoz: > I was trying to keep the tables as close to the code as possible

I'm not sure if I agree with this approach going forward with other migrations from redis to the API. The design that makes sense for redis doesn't necessarily make sense for the API. I'd rather not have the API be beholden to the choices that were made for redis. This is more of a general comment; don't take it to mean that my alternative suggestion in this case was definitely the better choice.

If we want to make these migrations quick, then that's okay. But we should revisit it eventually then.

D0rs4n: Regardless of the general approach proposed, I'd still part with using two tables, to keep things consistent, and easier to handle on the bot side, as they're two distinct commands.

pydis_site/apps/api/models/bot/aoc_completionist_block.py
```diff
@@ -0,0 +1,26 @@
...
+
+    is_blocked = models.BooleanField(
+        default=True,
+        help_text="Whether this user is actively being blocked "
+                  "from getting the AoC Completionist Role",
+        verbose_name="Blocked"
+    )
+    reason = models.TextField(
+        null=True,
```
MarkKoz, line 24: Better to use `blank` instead of `null`. But maybe we should leave it `null` since that seems to be what other endpoints use for similar fields (and change them all later in a separate PR if we want)?

https://docs.djangoproject.com/en/3.2/ref/models/fields/#null

D0rs4n: Yes, I used Infractions (the model) as a reference, to keep things consistent

pydis_site/apps/api/serializers.py
```diff
@@ -250,6 +252,26 @@ class Meta:
...
 
 
+class AocCompletionistBlockSerializer(ModelSerializer):
+    """A class providing (de-)serialization of `AocCompletionistBlock` instances."""
+
+    class Meta:
+        """Metadata defined for the Django REST Framework."""
+
+        model = AocCompletionistBlock
+        fields = ("user", "is_blocked", "reason")
```
MarkKoz, line 262: I'm not sure, but I think this is redundant since it's listing all fields. I believe leaving the fields unspecified would be equivalent.

pydis_site/apps/api/viewsets/bot/aoc_completionist_block.py
```diff
@@ -0,0 +1,72 @@
...
+    ## Routes
+
+    ### GET /bot/aoc-completionist-blocks/
+    Returns all the AoC completionist blocks
+
+    #### Response format
+    >>> [
+    ...     {
+    ...         "user": 2,
+    ...         "is_blocked": False
```
MarkKoz:
```suggestion
    ...         "is_blocked": False,
```

pydis_site/apps/api/viewsets/bot/aoc_completionist_block.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+    ### GET /bot/aoc-completionist-blocks/<user__id:int>
+    Retrieve a single Block by User ID
+
+    #### Response format
+    >>>
+    ...     {
+    ...         "user": 2,
+    ...         "is_blocked": False
```
MarkKoz:
```suggestion
    ...         "is_blocked": False,
```

pydis_site/apps/api/viewsets/bot/aoc_completionist_block.py
```diff
@@ -0,0 +1,72 @@
...
+    - 200: returned on success
+    - 404: returned if an AoC completionist block with the given user__id was not found.
+
+    ### POST /bot/aoc-completionist-blocks
+    Adds a single AoC completionist block
+
+    #### Request body
+    >>> {
+    ...     "user": int,
+    ...     "is_blocked": bool
```
MarkKoz:
```suggestion
    ...     "is_blocked": bool,
```

pydis_site/apps/api/serializers.py
D0rs4n: Doesn't this makes it a bit easier to understand?
In my personal opinion it makes it easier to understand what those serializers are for, and that's our aim I believe.

pydis_site/apps/api/viewsets/bot/aoc_completionist_block.py
```diff
@@ -0,0 +1,72 @@
...
+    #### Status codes
+    - 204: returned on success
+    - 404: if the AoC Completionist block with the given user__id does not exist
+
+    """
+
+    serializer_class = AocCompletionistBlockSerializer
+    queryset = AocCompletionistBlock.objects.all()
+    filter_backends = (DjangoFilterBackend,)
+    filter_fields = ("user__id",)
```
MarkKoz: Probably want to be able to filter by `is_blocked` too.

pydis_site/apps/api/serializers.py
MarkKoz: I don't know. Maybe. But it's also yet another thing to maintain when a field gets changed. On the other hand, maybe we don't want new fields to implicitly be included in the serializer. Just a comment so I don't feel strongly either way.

pydis_site/apps/api/viewsets/bot/aoc_completionist_block.py
```diff
@@ -0,0 +1,72 @@
...
+    ...     "reason": string
+    ... }
+
+    #### Status codes
+    - 204: returned on success
+    - 400: if one of the given fields is invalid
+
+    ### DELETE /bot/aoc-completionist-blocks/<user__id:int>
+    Deletes the AoC Completionist block item with the given `user__id`.
+    #### Status codes
```
MarkKoz:
```suggestion
    Deletes the AoC Completionist block item with the given `user__id`.

    #### Status codes
```

pydis_site/apps/api/viewsets/bot/aoc_link.py
```diff
@@ -0,0 +1,71 @@
...
+    ...     'aoc_username': str
+    ... }
+
+    #### Status codes
+    - 204: returned on success
+    - 400: if one of the given fields is invalid
+
+    ### DELETE /bot/aoc-account-links/<user__id:int>
+    Deletes the AoC account link item with the given `user__id`.
+    #### Status codes
```
MarkKoz:
```suggestion
    Deletes the AoC account link item with the given `user__id`.

    #### Status codes
```

pydis_site/apps/api/viewsets/bot/aoc_link.py
```diff
@@ -0,0 +1,71 @@
...
+    ### GET /bot/aoc-account-links<user__id:int>
+    Retrieve a AoC account link by User ID
+
+    #### Response format
+    >>> [
+    ...     {
+    ...         "user": 2,
+    ...         "aoc_username": "AoCUser1"
+    ...     },
+    ... ]
```
MarkKoz: Does this really still return a list for just 1 user? I am pretty sure the default DRF behaviour is to return a single object for /endpoint/id requests.

pydis_site/apps/api/viewsets/bot/aoc_link.py
```diff
@@ -0,0 +1,71 @@
...
+    >>> [
+    ...     {
+    ...         "user": 2,
+    ...         "aoc_username": "AoCUser1"
+    ...     },
+    ...     ...
+    ... ]
+
+
+    ### GET /bot/aoc-account-links<user__id:int>
```
MarkKoz: Surely it's meant to be this?
```suggestion
    ### GET /bot/aoc-account-links/<user__id:int>
```

pydis_site/apps/api/viewsets/bot/aoc_link.py
```diff
@@ -0,0 +1,71 @@
...
+
+    #### Status codes
+    - 204: returned on success
+    - 400: if one of the given fields is invalid
+
+    ### DELETE /bot/aoc-account-links/<user__id:int>
+    Deletes the AoC account link item with the given `user__id`.
+    #### Status codes
+    - 204: returned on success
+    - 404: if the AoC account link with the given user__id does not exist
```
MarkKoz: Some minor things in the docs like `user__id` not always being surrounded by backticks or some sentences not having full stops, but I will leave it up to you if you want to be bothered with fixing that.

MarkKoz: Another thing is inconsistent use of tense/mood.

pydis_site/apps/api/viewsets/bot/aoc_link.py
D0rs4n: Yes, I was under the impression that it does, but after testing it out yesterday, it seemed to return a list still.

D0rs4n: Don't mind me, you're right! :D 
Not sure why I updated in the end.

pydis_site/apps/api/viewsets/bot/aoc_link.py
```diff
@@ -55,14 +54,14 @@ class AocAccountLinkViewSet(
...
-    - 400: if one of the given fields is invalid
+    - 400: if one of the given fields was invalid
 
     ### DELETE /bot/aoc-account-links/<user__id:int>
     Deletes the AoC account link item with the given `user__id`.
 
     #### Status codes
-    - 204: returned on success
-    - 404: if the AoC account link with the given user__id does not exist
+    - 204: returned on succes
```
MarkKoz: This was correct originally.
```suggestion
    - 204: returned on success
```

MarkKoz (changes requested):

pydis_site/apps/api/viewsets/bot/aoc_link.py
```diff
@@ -29,20 +29,19 @@ class AocAccountLinkViewSet(
...
 
     #### Response format
-    >>> [
+    >>>
     ...     {
     ...         "user": 2,
     ...         "aoc_username": "AoCUser1"
-    ...     },
-    ... ]
+    ...     }
```
MarkKoz: Shouldn't be indented as far.

MarkKoz (changes requested):

MarkKoz:
Are you not planning on unit tests by the way? Not sure if there is much to gain anyway since this is mostly relying on DRF.

D0rs4n:
Initially I was not, I checked out a similar migration PR (namely the PR made by @ChrisLovering ) to see if there were tests.
But yes, these things are strongly reliant on DRF, and I did not introduce things such as complicated constraints, or custom actions.

MarkKoz (approved):

Issue #695, opened 2022-03-13
Migrating #discord-bots pins to the site


Shom770:
This issue is for the ongoing project to migrate all the pinned messages in #discord-bots to the website under the [guides](https://www.pythondiscord.com/pages/guides/python-guides/).

# Pins to be Migrated
If any of the pins below are checked off (checkmark next to the pin), a pull request has already been opened for that pin, and the pull request will be linked.

- [x] [Messages with Colors](https://discord.com/channels/267624335836053506/343944376055103488/928909158764396554) (Krypton) (#691)
- [x] [Discord Embed Limits](https://discord.com/channels/267624335836053506/343944376055103488/849552077092552704) (Krypton) (#690)
- [x] [Custom Help Commands](https://discord.com/channels/267624335836053506/343944376055103488/874712618827403375) (Raven) (#698)
- [x] [VPS services](https://discord.com/channels/267624335836053506/343944376055103488/874667287918026802) (Python) (#697)
- [x] [Replit vs Heroku](https://discord.com/channels/267624335836053506/343944376055103488/869084309334663169) (Raven) this is to be added to the above VPS page
- [x] [Discord bot token information](https://discord.com/channels/267624335836053506/343944376055103488/845290598394494997) (Tizzy) (#701)
- [x] [Why JSON is unsuitable as a database](https://discord.com/channels/267624335836053506/343944376055103488/822824159091884042) (Xithrius) (#700)
- [x] [Fixing SSL Certificate Issue](https://discord.com/channels/267624335836053506/343944376055103488/716675308253610074) (Stella) (#796)
- [x] [Setting different statuses on your bot](https://discord.com/channels/267624335836053506/343944376055103488/730574118658900039) (iamericfletcher) (#709)
- [x] [on_command_error eating unhandled errors](https://discord.com/channels/267624335836053506/343944376055103488/796212280244305970) (Atom) (#765)

# Additional Topics
The following topics below are additional topics that can be added to the site, regarding discord bots.
- [x] Tutorial on subclassing Bot (#702) (Waiting for author)
- [x] Tutorial on subclassing Context (#703) (Needs review)
- [x] DevOps guide for Discord Bots (#746) (Waiting for author)
- [x] Hikari guide (#738) (Closed due to lack of review)
- [x] Application Commands Guide (#733) (Waiting for author)

# Tags
Per request, we also want to scope the addition of new tags as part of this project.
These tags should follow the same traditional pattern as any other tag that we would add.
[...]

PR #700, opened 2022-03-21
Don't do JSON as a database, kids.

Xithrius:
It won't do you well.

Finishes off one of the check marks for #695.


jchristgit (approved):
This migrates the pin, so I'm approving.

That said, I do think the content is a bit weird, since the second paragraph makes the following points:

- "No redundancy built into JSON"
-  "No built in ways to keep it safe and backed up"
- "No built in encryption", with some inclusion of GDPR

This, to me, sounds like standard databases offer any of that, out of the box. Unfortunately there's no magic setting or function in PostgreSQL where I can go `SELECT spin_up_slaves(3)`, `SELECT perform_encrypted_redundant_backup()` and `SELECT pg_encrypt_harddrive()`, those are all still left to the operator.

Regardless, this makes some good points, so let's get it in.

jchristgit:
Let's maybe revise the content later, but migrate the pin here first.

pydis_site/apps/content/resources/guides/python-guides/why-not-json-as-database.md
```diff
@@ -0,0 +1,12 @@
...
+
+JSON, unlike relational databases, has no way to store relational data, which is a very commonly needed way of storing data. Relational data, as the name may suggest, is data that relates to other data. For example if you have a table of users and a table of servers, the server table will probably have an owner field, where you'd reference a user from the users table. [This is only relevant for relational data]
+
+JSON is primarily a KV (key-value) format, for example {"a":"b"} where a is the key and b is the value, but what if you want to search not by that key but by a sub-key? Well, instead of being able to quickly use var[key], which in a Python dictionary has a constant return time (for more info look up hash tables), you now have to iterate through every object in the dictionary and compare to find what you're looking for. Most relational database systems, like MySQL, MariaDB, and PostgreSQL have ways of indexing secondary fields apart from the primary key so that you can easily search by multiple attributes.
```
Bluenix2: Yeah I still think these should be split into multiples lines capping at like 120 which is what Python Discord uses for code

Bluenix2 (approved):
That was my only complaint. Thanks for the PR Xithrius (and Mina for finishing it 💯)!

PR #711, opened 2022-04-12
add article discussing free hosting services

Diabolical5777:


sue-m-oh:
@Robin5605 @Diabolical5777 and I have discussed adding this content as a section to the existing page which recommends [VPS Services](https://www.pythondiscord.com/pages/guides/python-guides/vps-services/).

Having this content on a page where we recommend good VPS options is a better option than having it on its own page because ultimately we want to fast-track the users to the good stuff.

@Robin5605 would you please take on this task?

CC: @Shom770

sue-m-oh:
It was [decided](https://discord.com/channels/267624335836053506/635950537262759947/963783266232598569) that @Diabolical5777 would continue the work on this instead.

PR #717, opened 2022-04-21
Add support for BIG SOLIDUS unicode characters for off topic names

minalike:
We must add additional support for these unicode characters in off-topic names.


ChrisLovering (approved):
Looks good 👌 Very important change, thanks

shenanigansd (approved):

PR #718, opened 2022-04-22
Invert Deleted Message Log Order

HassanAbouelela:
Inverts the order messages are displayed in on the logs page to match the order they would be displayed in on discord (and how they would be read).


ChrisLovering (approved):
thank

minalike (approved):
Works beautifully

PR #719, opened 2022-04-27
Force line breaks at arbritrary points for really long words

minalike:
Currently if a message contains a very long word the content doesn't wrap and extends off the page.
image

This PR forces line breaks in what normally are unbreakable words.
image


HassanAbouelela (approved):

GDWR (approved):
Looks like it will do the trick, nice one.

PR #722, opened 2022-06-13
Resources: Add The Algorithms, remove Atom

### LINKED ISSUE #180: Algorithms Github (Python)
### Resource type

Community

### Pricing

Free

### Where can the resource be found?

https://github.com/TheAlgorithms/Python

### What sort of content is in the resource? What areas of Python does it cover?

Inside the Github repository there are a list of algorithms implemented in Python.

### Please give any additional information about the resource you think is worth including, as well as any potential limitations

_No response_

### LINKED ISSUE #185: Resources: Atom (the editor) is being sunset
The Atom editor [is being sunset in December](https://github.blog/2022-06-08-sunsetting-atom/).

My recommendation is that we simply delete it from the resources page. We already list Mu-Editor, PyCharm, Spyder, Sublime, Thonny, and VSC. I'm not sure which audience Atom is intended to appeal to, but I suspect part of the reason for them sunsetting Atom is that its audience might have shifted to one of the other editors that we feature.

Does anyone know what the value proposition of Atom is, as compared to the other editors? If there is one, I suppose we could change the description to mention that it's going to be sunset and see what happens (Senjan pointed out that it might be forked). Otherwise, I don't think this resource should be featured to reduce the frequency of option paralysis for our users.

swfarnsworth:
Adds The Algorithms as a new resource, and removes Atom. Reasons for these changes can be found in the two issues that this closes. Also adds the PyCharm logo to that resource.

I do not have an environment set up to run the site locally, so I would appreciate if the reviewer could verify the changes.

Closes python-discord/meta#185
Closes python-discord/meta#180


pydis_site/apps/resources/templatetags/get_category_icon.py
```diff
@@ -21,6 +21,7 @@
     "Paid": "fa-dollar-sign",
     "Podcast": "fa-microphone-alt",
     "Project Ideas": "fa-lightbulb-o",
+    "Security": "lock-alt",
```
sue-m-oh: Can someone tell me what this does? I am unfamiliar. (Where do I check that this change worked?)

minalike (approved):
I've checked out your PR and ran site locally, everything looks good. 

Just not sure how to check what was changed by the change to `get_category_icon.py`.

pydis_site/apps/resources/templatetags/get_category_icon.py
MarkKoz: They're the icons at the bottom right in the following example 
bild

To confirm it works, find a resource that has the security category and confirm that at the bottom right you see a lock icon next to the security category for that resource.

sue-m-oh: Replaced with `fa-solid fa-lock` and confirmed it works 
image

MarkKoz (approved):

shenanigansd (approved):

PR #726, opened 2022-06-19
Add Summer Code Jam 2022 banner to frontpage.

kosayoda:


mbaruh:
Looks good to me. ~~One issue is that there is a "qualifier open" card and then a separate "sign up now"~~

~~After the qualifier you can't sign up anymore so it probably shouldn't say that~~

mbaruh (approved):
Thanks!

ChrisLovering (approved):
good

PR #729, opened 2022-06-22
Update Discord logo in navbar to use new Discord branding

lemonyte:
Updated the Discord icon and text in the navbar. Color is not final.


lemonyte:
Color variants shown below (new blurple, old blurple, black) with the current unmodified SVG for reference. The second one violates Discord's branding policy because it has been recolored to the old blurple color (pre-May 13th 2021 rebrand).

ChrisLovering (approved):

PR #734, opened 2022-06-28
Rework and reorder CJ9 framework descriptions

Bluenix2:
This PR, simply put, reorders the CJ9 frameworks to be in a more familiar order based on recommendation:
- FastAPI and Starlette are both ASGI frameworks many should have heard of
- websockets is the only client library on the list
- Django channels is kept on the list but I did not purposefully give it a different position
- wsproto should be last, only for those who wish to use it

Additionally, the descriptions have been reworded to shine more light on the differences between the frameworks and their websocket aspect of it.

*Flask-socketIO is currently up for discussion so I have not touched it.*

### DIFF
--- a/pydis_site/templates/events/pages/code-jams/9/frameworks.html
+++ b/pydis_site/templates/events/pages/code-jams/9/frameworks.html
@@ -94,6 +94,10 @@
                 <p class="is-italic">ASGI is the specification your favourite frameworks are built on. This is not a framework in of itself, but listed here for completeness.</p>
             </div>
         </div>
+        <div class="card-footer">
+            <a href="https://asgi.readthedocs.io/en/latest/" class="card-fotter-item"><i class="fa-clipboard-list"></i>Specification</a>
+        </div>
+    </div>
 
     <div class="card mb-4">
         <div class="card-content">
--- a/pydis_site/templates/events/pages/code-jams/9/frameworks.html
+++ b/pydis_site/templates/events/pages/code-jams/9/frameworks.html
@@ -94,7 +94,7 @@
             </div>
         </div>
         <div class="card-footer">
-            <a href="https://asgi.readthedocs.io/en/latest/" class="card-fotter-item"><i class="fas fa-clipboard-list"></i>&ensp;Specification</a>
+            <a href="https://asgi.readthedocs.io/en/latest/" class="card-footer-item"><i class="fas fa-clipboard-list"></i>&ensp;Specification</a>
         </div>
     </div>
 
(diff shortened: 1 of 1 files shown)


Bluenix2:
image
image

mbaruh:
Maybe it shouldn't be in italics anymore? Since they're no longer quotes

Bluenix2:
Unfortunately, it appears that Netlify was ratelimited which caused the deploy and check to fail.

pydis_site/templates/events/pages/code-jams/9/frameworks.html
```diff
@@ -16,39 +16,53 @@
...
         </div>
     </div>
+
     <div class="card mb-4">
         <div class="card-content">
             <div class="content">
                 <p class="subtitle">Django Channels</p>
-                <p class="is-italic">Channels is a project that takes Django and extends its abilities beyond HTTP - to handle WebSockets, chat protocols, IoT protocols, and more.
-                    It’s built on a Python specification called ASGI.
+                <>Django Channels adds WebSocket-support to Django - built on ASGI like other web frameworks.
```
ChrisLovering:
```suggestion
                Django Channels adds WebSocket-support to Django - built on ASGI like other web frameworks.
```

ChrisLovering (dismissed):
Looks good, other than the floating tag.

Updated screenshot below. All links work as expected. 

image

pydis_site/templates/events/pages/code-jams/9/frameworks.html
Bluenix2: This is supposed to be a paragraph tag that is supposed to be closed below. It appears that I accidentally also removed the `p`.

I can fix this later on.

pydis_site/templates/events/pages/code-jams/9/frameworks.html
```diff
@@ -17,6 +17,16 @@
         If there is a library not listed below that you think should be here, you're welcome to discuss it with the Events Team over at <a href="https://discord.gg/HnGd3znxhJ">the server</a>.
     </p>
 
+    <div class="notification is-info is-light">
+        <p>Most of the below frameworks implement what is called the ASGI Specification.
+            This specification documents how the frameworks should interact with ASGI servers and call the frameworks.
```
kosayoda:
```suggestion
            This specification documents how the frameworks should interact with ASGI servers.
```
This reads better to me personally, thoughts?

kosayoda (changes requested):
One small thing, otherwise LGTM

pydis_site/templates/events/pages/code-jams/9/frameworks.html
Bluenix2: I agree, commited! 👍

kosayoda (approved):
Thanks a ton!

PR #740, opened 2022-07-11
Update Django Version To 4.0

HassanAbouelela:
This PR started out as something completely different, but I kept getting stuck because our dependencies are so outdated Now this PR only bumps the django version to 4.0. Most things are self-explanatory, and those that aren't are documented in their relevant commits. I've left this PR in a planning stage for now to allow for feedback on 7b40cd8. It should still be ready for a review.

This PR also kaizens in a switch to httpx from requests. It's a drop in replacement, but provides a nicer interface for some things. Refer to c24cccc. It also kaizens in an update for some lintning dependencies, because loosely-versioned dependencies and `poetry update` do not go together well.


HassanAbouelela:
So far everything looks good, going to continue to 4.0 now

HassanAbouelela:
**Background on 7b40cd8**

I've worked on this for a while, so here's a bit more information on why that part of the test suite was dropped. After some debugging, The custom setup for migration tests instantiates a few model instances. Going from django `3.1.14` (what we were on) to `3.2.0` (the version directly after) introduces behavior that breaks this. At some point between the setup and the actual test the (already created) instances' `_meta` model gets updated to the latest model, which breaks everything.

@D0rs4n identified that it might be related to the deprecation notice under [this](https://docs.djangoproject.com/en/4.0/topics/testing/tools/#django.test.TestCase.setUpTestData) method, and suggested using `setUpClass` instead. That did resolve the issues with the migration models, and broke the rest of the test suite.

I think we could in theory get this working, but I'm not sure if we necessarily should. It seems to me that these tests are technically fascinating but practically useless. Why are we testing that a historical migration, which is not going to change. If we know it works as expected, then these tests are probably redundant.

I'm hoping Sebastiaan could provide some context on why it was added in the first place. If it's still relevant, I'll work on reintroducing it.

cc @SebastiaanZ @mbaruh

HassanAbouelela:
Added the do not merge tag. I think it's best if dev-ops is on hand for this. We should create a DB backup before deploying.

SebastiaanZ:
> I'm hoping Sebastiaan could provide some context on why it was added in the first place. If it's still relevant, I'll work on reintroducing it.

It was introduced because we were writing code that would directly affect import production data, namely a data migration that would directly touch all of our infraction records.

This allowed is to test the code that we were writing.

These days, there are a few packages available that basically implement this functionality without having to do it yourself, like `django-test-migrations`.

The migrations we are testing here have been rolled out to all environments, so I'd say it's okay to remove the tests for them.

If we start writing custom/data migrations again, it would be trivial to introduce the aforementioned package. That means our custom test class can go as well.

(The breakage would be trivial to fix if we decide otherwise. Just let me know, and I can update it.)

ChrisLovering (approved):
httpx good, tests bad

jb3 (approved):
I love it. I love everything about it.

PR #741, opened 2022-07-12
Remove IP Specification From Compose Ports

HassanAbouelela:
This PR originally disabled postgres' durable optimizations for a significant speedup in testing, but more experimentation proved that it was the port changes that had the most effect. There is more info below, though the reason this works remains a bit mysterious.


docker-compose.yml
```diff
@@ -42,12 +53,15 @@ services:
     depends_on:
       postgres:
         condition: service_healthy
+      test-postgres:
+        condition: service_started
```
MarkKoz: Does this mean it's going to run two instances of the db?

.github/workflows/lint-test.yaml
```diff
@@ -97,13 +97,13 @@ jobs:
...
           python manage.py makemigrations --check
-          python manage.py migrate
           coverage run manage.py test --no-input
           coverage report -m
         env:
           CI: True
-          DATABASE_URL: postgres://pysite:pysite@localhost:7777/pysite
-          METRICITY_DB_URL: postgres://pysite:pysite@localhost:7777/metricity
+          DATABASE_URL: postgres://pysite:pysite@localhost:7778/pysite
+          TEST_DB_URL: postgres://pysite:pysite@localhost:7778/pysite
```
MarkKoz: What's the point of separate URLs if they're identical anyway? Do we _need_ the flexibility to make them different if we want?

docker-compose.yml
HassanAbouelela: Yes, it does. There's the option of not adding this to the depends-on and letting people manually invoke it if we don't want that.

.github/workflows/lint-test.yaml
HassanAbouelela: They're both needed as one is used in the migrations, and one is used in the tests. Making them one environment variable would be annoying as locally and in the compose they aren't going to refer to the same thing.

docker-compose.yml
MarkKoz: Well, Docker Compose is intended to only be used for development purposes, right? So why don't we just always disable durability? I don't think anyone would cry over development data being lost in the rare case of power loss or a crash.

HassanAbouelela: That's a fair point, I don't see a good reason not to.

.github/workflows/lint-test.yaml
MarkKoz: Oh sorry, I see that they are different by default in docker-compose.yml. I confused this file with docker-compose.yml.

HassanAbouelela:
While trying to merge both databases as per Mark's suggestion, I made an interesting discovery. At first, the merge seemed to slow the tests again, despite nothing having really changed. Upon closer inspection, the port I defined on the service, was slightly different from the one we currently have (`127.0.0.1:port:port` current vs `port:port` my change).

I can't explain it, and it's possible I have something messed up in my networking, but I can reproducibly demonstrate it makes an order of magnitude difference. @MarkKoz also reproduced the same thing on his machine. This might also explain why the CI already runs in 20 seconds, since that doesn't use the port definition from the compose.

Either way, I've reverted the changes I've made now, and only kept the relevant ones.

docker-compose.yml
```diff
@@ -8,12 +8,12 @@
 # and additionally use the Django development server which is
 # unsuitable for production.
 
-version: "3.6"
+version: "3.8"
```
MarkKoz, line 11: For future reference, there isn't really a point in changing this as its deprecated

https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element

> A Compose implementation SHOULD NOT use this version to select an exact schema to validate the Compose file, but prefer the most recent schema at the time it has been designed.

Maybe there are still some tools that are not respecting this advice?

pyproject.toml
```diff
@@ -51,7 +51,7 @@ start = "python manage.py run --debug"
 makemigrations = "python manage.py makemigrations"
 django_shell = "python manage.py shell"
 test = "coverage run manage.py test"
-coverage = "coverage run manage.py test --no-input; coverage report -m"
+coverage = "coverage run manage.py test --no-input && coverage report -m"
```
MarkKoz, line 54: Would be nice if this ran migrations too.

MarkKoz (approved):

docker-compose.yml
HassanAbouelela: My compose does use the behavior you described above, so I'd be happy to remove it. The reason I bumped it is because our current compose is using features that weren't even introduced in 3.6, and I wanted to fix the linting issues.

MarkKoz: Does `docker-compose` still rely on it (notice the dash)? Maybe wait until that's completely phased out? I don't really know. It doesn't hurt to leave it for now I guess.

pyproject.toml
HassanAbouelela: That sounds fine. Would you prefer the initial approach of putting it in the manage.py, or should we do it here?

MarkKoz: Doing it in `manage.py` would be better so it still works without reliance on Poetry's task.

MarkKoz: And you could also avoid useless stuff like collecting static files.

Edit: never mind that doesn't happen to begin with. I just got confused when viewing your old diff.

HassanAbouelela: Added in 98fa159

ChrisLovering (approved):

PR #742, opened 2022-07-12
Add GitHub Artifacts API Route

### LINKED ISSUE #728: Deploy Preview Failures
We get occasional CI failures from the netlify preview deploys due to GitHub rate limits. [Example](https://app.netlify.com/sites/pydis-static/deploys/62b21f7a0849350009d6ce49):

```
  File "static-builds/netlify_build.py", line 122, in <module>
    download_artifact(*get_build_artifact())
  File "static-builds/netlify_build.py", line 36, in get_build_artifact
    pull_request.raise_for_status()
  File "/opt/buildhome/python3.8/lib/python3.8/site-packages/httpx/_models.py", line 1423, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: 403 Client Error: rate limit exceeded for url: https://api.github.com/repos/python-discord/site/pulls/727
For more information check: https://httpstatuses.com/403
```

This issue is meant for finding workarounds that allow us to avoid these issues, with the following limitations in mind:
1. We can't use action secrets, since those are not [accessible to external contributors](https://github.community/t/can-pull-requests-change-actions/143120).
2. We can't use any sensitive secrets in the netlify build, since the build script from the PR is executed with full trust. Assume any secrets in netlify are publicly visible.

HassanAbouelela:
## Description
This PR adds an API route, which uses a [github app](https://docs.github.com/en/developers/apps/getting-started-with-apps/about-apps) (not to be confused with OAuth apps) to return a download URL for the static preview artifact. This should allow us to get around the rate limiting issues that we've been experiencing in #728.

**A bit more info on my approach:**
Q. Why a GitHub app?
A. It provides the most granularity in terms of permissions. We can get very specific about which resources we need, what type of access (read, write), and on which repos it's installed. It requires more work in the codebase than something like PATs, but it's about a million times safer.

Q. Why on the site, instead of a cloudflare worker?
A. The approach used here would've been the same had we used CF workers, but keeping it in site and keeping it in python makes it easier to debug and maintain as we have a lot more Python developers who can help keep the script running.

## Testing the PR
This PR is expected to fail actions, since it relies on a (till this is merged) non-existent API route. I've deployed the site, and tested everything out ([PR](https://github.com/HassanAbouelela/pydis-site/pull/4), [site](https://app.netlify.com/sites/clever-klepon-54aabb/deploys/)). I've taken down the site for now, but I can throw it back up if someone needs it for testing. The netlify build script does not need to be tested *on* netlify, if you don't want to.

## Roll-out plan
As mentioned, I expect the action to fail on this PR. Once the branch is ready, I'll update the command and environment variables in netlify, then we'll merge and hopefully see some green ticks. We also need to roll out the changes to bot-core, but that can be done in a separate PR.


pydis_site/apps/api/github_utils.py
```diff
@@ -0,0 +1,184 @@
...
+            "Could not find the requested repository. Make sure the application can access it."
+        )
+
+    except BaseException as e:
+        # Close the client if we encountered an unexpected exception
+        client.close()
+        raise e
+
+
+def check_run_status(run: dict) -> str:
```
ChrisLovering: Would it be a good idea to dataclass/namedtuple for this run arg? Just so that we can know what to expect in the data, and maybe error at creation time if it's missing a field we expect?

ChrisLovering (approved):
From your testing it clearly works, the tests seem to cover quite a bit too.

Just have a question, that could be a nice to have/not needed, but shouldn't block this PR.

pydis_site/apps/api/github_utils.py
HassanAbouelela: I think that sounds reasonable, though the class would be pretty bare bones because the documentation is too. I'd probably just include the fields we do use because it's also a very large model.

HassanAbouelela: 37001bc

pydis_site/apps/api/github_utils.py
```diff
@@ -0,0 +1,209 @@
...
+    created_at: str
+    status: str
+    conclusion: str
+    artifacts_url: str
+
+    @classmethod
+    def from_raw(cls, data: dict[str, typing.Any]):
+        """Create an instance using the raw data from the API, discarding unused fields."""
+        return cls(**{
+            key.name: data[key.name] for key in dataclasses.fields(cls)
```
jchristgit, line 73: this is such a cool trick

pydis_site/apps/api/github_utils.py
```diff
@@ -0,0 +1,209 @@
...
+                    # We've found the correct repository, and it's accessible with the current auth
+                    return client
+
+        raise NotFoundError(
+            "Could not find the requested repository. Make sure the application can access it."
+        )
+
+    except BaseException as e:
+        # Close the client if we encountered an unexpected exception
+        client.close()
```
jchristgit, line 144: Shouldn't this be in a `finally`? Otherwise we leak, unless httpx closes it in `__del__`.

jchristgit: This is also how it's done further below

pydis_site/apps/api/tests/test_github_utils.py
```diff
@@ -0,0 +1,285 @@
...
+class GeneralUtilityTests(unittest.TestCase):
+    """Test the utility methods which do not fit in another class."""
+
+    def test_token_generation(self):
+        """Test that the a valid JWT token is generated."""
+        def encode(payload: dict, _: str, algorithm: str, *args, **kwargs) -> str:
+            """
+            Intercept the encode method.
+
+            The result is encoded with an algorithm which does not require a PEM key, as it may
```
jchristgit, line 26: is that why the addition of `*.pem` in `.gitignore`?

jchristgit (approved):
This looks good. But shouldn't the endpoint have some authorization / whitelisting / ratelimiting?

HassanAbouelela:
@jchristgit I wanted to add auth, but because we are running a file hosted in the repo, netlify is not a trusted environment. If someone wanted to be malicious, they can print out environment vars and get the key.

I suppose we could still do auth and accept that it can be leaked and we'll have to regenerate it sometimes. How would rate limiting look?

pydis_site/apps/api/github_utils.py
HassanAbouelela: The client only needs to be closed if we encounter an error. If we don’t, the client is returned to the caller to be used for all further interactions. The close below is what closes it in case of success.

pydis_site/apps/api/tests/test_github_utils.py
HassanAbouelela: Yeah private key encoding is mandated for GitHub apps. In prod, we can deploy it as an environment variable, but this is more convenient for local testing.

jchristgit:
> netlify is not a trusted environment. If someone wanted to be malicious, they can print out environment vars and get the key.

Great! That sounds very sane and secure.

> I suppose we could still do auth and accept that it can be leaked and we'll have to regenerate it sometimes. How would rate limiting look?

Hm, not much value in case it can be leaked. Rate limiting - Maybe this? https://www.django-rest-framework.org/api-guide/throttling/ It's overkill though. We don't need it.

HassanAbouelela:
The necessary environment variables have been deployed to site and netlify, this PR should now be good to merge.

PR #743, opened 2022-07-14
Make finalizing codejam 9 changes

Bluenix2:
This PR makes some last changes to the code jam 9 approved frameworks and rules which have been discussed these last days. See the commit history for a detailed list of changes.

### DIFF
--- a/pydis_site/templates/events/pages/code-jams/9/frameworks.html
+++ b/pydis_site/templates/events/pages/code-jams/9/frameworks.html
@@ -85,20 +85,6 @@
     <div class="card mb-4">
         <div class="card-content">
             <div class="content">
-                <p class="subtitle">wsproto</p>
-                <p>wsproto is a pure-Python WebSocket protocol stack written to be as flexible as possible by having the user build the bridge to the I/O.
-                </p>
-            </div>
-        </div>
-        <div class="card-footer">
-            <a href="https://python-hyper.org/projects/wsproto/en/stable" class="card-footer-item"><i class="fas fa-book"></i>&ensp;Documentation</a>
-            <a href="https://github.com/python-hyper/wsproto" class="card-footer-item"><i class="fab fa-github"></i>&ensp;GitHub</a>
-        </div>
-    </div>
-
-    <div class="card mb-4">
-        <div class="card-content">
-            <div class="content">
                 <p class="subtitle">Starlite</p>
                 <p>Starlite is a light and flexible ASGI API framework, using Starlette and Pydantic as foundations.
                 </p>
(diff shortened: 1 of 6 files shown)


pydis_site/templates/events/pages/code-jams/9/frameworks.html
```diff
@@ -71,70 +71,70 @@ <h3 id="approved-frameworks"><a href="#approved-frameworks">Approved Frameworks<
     <div class="card mb-4">
         <div class="card-content">
             <div class="content">
-                <p class="subtitle">Django Channels</p>
-                <p>Django Channels adds WebSocket-support to Django - built on ASGI like other web frameworks.
+                <p class="subtitle">AioHTTP</p>
+                <p>AioHTTP provides both a client and server WebSocket implementation, while avoiding callback-hell.
```
Numerlor: The capitalization here looks weird, IO is an abbreviation that should also be capitalized. But looking through the docs, only the title of the main page had AIOHTTP capitalized while all other uses are lowercase
```suggestion
                <p class="subtitle">aiohttp</p>
                <p>aiohttp provides both a client and server WebSocket implementation, while avoiding callback-hell.
```

Numerlor (changes requested):
https://github.com/python-discord/site/blob/7511c6d4aff6f3ad31ccb43c00226a9187d7ece1/pydis_site/templates/events/pages/code-jams/9/_index.html#L27-L28 is also currently missing strike through for dates that have passed

Bluenix2:
Thank you for bringing that up, I was considering opening another PR for those changes, but I might as well include that here on second consideration. I have now commited that

pydis_site/templates/events/pages/code-jams/9/rules.html
```diff
@@ -12,6 +12,15 @@
 {% block event_content %}
 <ol>
     <li><p>Your solution must use one of the approved frameworks (a list will be released soon). It is not permitted to circumvent this rule by e.g. using the approved framework as a wrapper for another framework.</p></li>
+    <li>
+        <p><strong>Your solution must only communicate through WebSockets</strong>. Any communication between the client and server (or any two server workers), must utilize WebSockets.</p>
+
+        <p>
+            Exceptions to this rule are made for resources such as databases and files, albeit excluding usage other than for storage.
+            For example, you may use PostgreSQL as a database but not its `NOTIFY` command.
+            Lastly, working with subprocesses (through stdin/stdout or <code>multiprocessing.Pool()</code>/<code>concurrent.futures.ProcessPoolExecutor()(</code>) is also exempted from this rule.
```
kosayoda:
```suggestion
            Lastly, working with subprocesses (through stdin/stdout or <code>multiprocessing.Pool()</code>/<code>concurrent.futures.ProcessPoolExecutor()</code>) is also exempted from this rule.
```

kosayoda (changes requested):
Everything else looks great!

kosayoda (approved):

pydis_site/templates/events/pages/code-jams/9/rules.html
```diff
@@ -12,6 +12,15 @@
 {% block event_content %}
 <ol>
     <li><p>Your solution must use one of the approved frameworks (a list will be released soon). It is not permitted to circumvent this rule by e.g. using the approved framework as a wrapper for another framework.</p></li>
+    <li>
+        <p><strong>Your solution must only communicate through WebSockets</strong>. Any communication between the client and server (or any two server workers), must utilize WebSockets.</p>
+
+        <p>
+            Exceptions to this rule are made for resources such as databases and files, albeit excluding usage other than for storage.
+            For example, you may use PostgreSQL as a database but not its `NOTIFY` command.
+            Lastly, working with subprocesses (through stdin/stdout or <code>multiprocessing.Pool()</code>/<code>concurrent.futures.ProcessPoolExecutor()</code>) is also exempted from this rule.
```
wookie184: Pending discussion on the server, this wording feels a little clumsy. Would this be better?
```suggestion
            An exception to this rule is that communication with databases and files is allowed for accessing resources or for storage purposes. 
            For example, you may use PostgreSQL as a database but not its `NOTIFY` command.
            Working with subprocesses (through stdin/stdout or <code>multiprocessing.Pool()</code>/<code>concurrent.futures.ProcessPoolExecutor()</code>) is also allowed.
```

(If you prefer your wording then `exempted` should be changed to `exempt`)

Bluenix2: I was kind of going back- and forth- on something like this, but couldn't quite formulate it like you just did. I'll apply that.

wookie184 (approved):
Thanks :D

PR #751, opened 2022-07-27
Add a field to track the time an infraction was last applied

ChrisLovering:
A default is set for backwards compatibility with bot version that don't explicitly give a value.

This is required to assist with https://github.com/python-discord/bot/issues/2130 and https://github.com/python-discord/bot/issues/2217 so that we can store when it was last edited. By comparing this date to the expiry date, we can get the absolute duration, rather than a rough duration which exhibits the issues above.


HassanAbouelela (approved):
Code looks valid

ChrisLovering:
The migration has been tested against prod, it worked at took <1s locally.

wookie184:
What is the purpose of this?

ChrisLovering:
> What is the purpose of this?

Sorry, I left out a lot of context as we were discussing in #dev-contrib. I've updated the PR description with some of that context.

ionite34 (approved):
Added field appears correct for supporting new duration calculations, and does not affect current API.

PR #762, opened 2022-08-12
Static Builds Small Improvements

HassanAbouelela:
This includes a few small bugfixes and improvements to #742. The first bugfix is for scenarios where the build action bypasses the maximum deadline specified by the API. This PR adds the missing raise_for_status which should display a better error message in that scenario. This PR also increases the timeout when downloading artifacts, since in the case of botcore it can go to ~100MB. The deadline for the API is also increased since bot-core can take upwards of 7 minutes in worst case scenarios.

### DIFF
--- a/static-builds/netlify_build.py
+++ b/static-builds/netlify_build.py
@@ -15,6 +15,18 @@ from urllib import parse
 
 import httpx
 
+
+def raise_response(response: httpx.Response) -> None:
+    """Raise an exception from a response if necessary."""
+    if response.status_code // 100 != 2:
+        try:
+            print(response.json())
+        except json.JSONDecodeError:
+            pass
+
+    response.raise_for_status()
+
+
 if __name__ == "__main__":
     owner, repo = parse.urlparse(os.getenv("REPOSITORY_URL")).path.lstrip("/").split("/")[0:2]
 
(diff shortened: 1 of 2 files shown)


ChrisLovering (approved):

static-builds/netlify_build.py
```diff
@@ -29,24 +41,18 @@
...
     while response.status_code == 202:
         print(f"{response.json()['error']}. Retrying in 10 seconds.")
         time.sleep(10)
         response = httpx.get(download_url, follow_redirects=True)
 
+    raise_response(response)
     url = response.json()["url"]
     print(f"Downloading build from {url}")
-    zipped_content = httpx.get(url, follow_redirects=True)
+    zipped_content = httpx.get(url, follow_redirects=True, timeout=3 * 60)
```
Xithrius, line 55: Do we want to wait 10 minutes here as well?

HassanAbouelela: The wait here is only for the download, so it should in theory only take a few seconds. 10 minutes might be a little over the top, but we've got plenty of build minutes to spare, so I don't mind either way.

Xithrius: 3 minutes will do, I needed confirmation between the two wait times.

Xithrius (approved):

PR #763, opened 2022-08-13
List Bot Tags On Site

### LINKED ISSUE #687: Page listing all tags and their content
I think it would be beneficial to have a page with toggle dropdowns listing all the tags we have and their contents for easy viewing.

We have the `!tags` on the server, but to read their contents you would have to run `!tag-name` on each individual tag or browse to our bot repository and read the markdown files directly. Both are not quite the smooth user experience I would want the folks on our server to have.

We can can build this page somewhat dynamically as it can pull from the specific bot repository on GitHub itself and use a jinja template to build the page from the markdown files.

HassanAbouelela:
Closes #687.

I implemented this feature in the most efficient way (in terms of network requests) possible. We download the entire bot project as a tarfile, extract all the tags from it, and cache them to the database.

Ideally, I'd also like to add a bit more flavor-info to the pages, such as when they were last updated, and the authors that have worked on them, but the only way to get that data from the API is to request data for a specific commit, which means making ~100 requests in total. We can't do that without being rate-limited, but it could be something to consider. I also considered downloading the git repo (as opposed to the content only), but it doesn't seem the API returns that, we'd need to clone with git.

Speaking of the 100 tags, I've gone through quite a few and they all look fine. They all should *render* fine, since the static deployment hits them all, but there could be some formatting oddities that I have missed. Feel free to tell me if so.


jahrules:
> Closes #687.
> 
> I implemented this feature in the most efficient way (in terms of network requests) possible. We download the entire bot project as a tarfile, extract all the tags from it, and cache them to the database.
> 
> Ideally, I'd also like to add a bit more flavor-info to the pages, such as when they were last updated, and the authors that have worked on them, but the only way to get that data from the API is to request data for a specific commit, which means making ~100 requests in total. We can't do that without being rate-limited, but it could be something to consider. I also considered downloading the git repo (as opposed to the content only), but it doesn't seem the API returns that, we'd need to clone with git.
> 
> Speaking of the 100 tags, I've gone through quite a few and they all look fine. They all should _render_ fine, since the static deployment hits them all, but there could be some formatting oddities that I have missed. Feel free to tell me if so.

What about the tag "rule <#>"

ChrisLovering:
> What about the tag "rule <#>"

Those aren't tags, they're a command in and of themselves. Also, the rules are already listed in our website (it's what the rules command links to).

pydis_site/apps/content/utils.py
```diff
@@ -27,8 +39,111 @@ def get_categories(path: Path) -> Dict[str, Dict]:
...
     """Get all page names and their metadata at a category path."""
+    # Special handling for tags
+    if path == Path(__file__).parent / "resources/tags":
+        tags = {}
+        for tag in get_tags():
+            content = frontmatter.parse(tag.body)[1]
+            if len(content) > 100:
+                # Trim the preview to a maximum of 100 visible characters
+                # This causes some markdown to break, but we ignore that
+                content = content[:100] + "..."
```
ChrisLovering:
```suggestion
                content = content[:100] + "…"
```
😡

pydis_site/apps/content/views/tags.py
```diff
@@ -0,0 +1,56 @@
...
+            raise Http404
+
+        context = super().get_context_data(**kwargs)
+        context["page_title"] = tag.name
+        body = frontmatter.parse(tag.body)
+        content = body[1]
+
+        # Check for tags which can be hyperlinked
+        start = 0
+        while match := COMMAND_REGEX.search(content, start):
```
ChrisLovering: Would a `re.finditer` be more readable here? Rather than managing the iteration ourselves with `match.end()`.

pydis_site/apps/content/utils.py
```diff
@@ -27,8 +39,111 @@ def get_categories(path: Path) -> Dict[str, Dict]:
...
+
+
+def get_category_pages(path: Path) -> dict[str, dict]:
     """Get all page names and their metadata at a category path."""
+    # Special handling for tags
+    if path == Path(__file__).parent / "resources/tags":
+        tags = {}
+        for tag in get_tags():
+            content = frontmatter.parse(tag.body)[1]
+            if len(content) > 100:
```
ChrisLovering: I think we can get away with more than 100 chars in the preview, even on mobile, something ~300 is likely fine.
The problems arise when we start grabbing the first parts of a codeblock, and then we don't close them.
Is there a way for us to detect if we're "inside" a codeblock after trimming, and close it before the ellipses? 

Also, if the start of tags include a long hyperlink, the link length is counted towards this 100 character limit. This leads to some tags looking far shorter than other (example below). I wonder if there is anything we can do about this?

image

ChrisLovering (changes requested):
Just a couple a comments here.

Another observation is that this list is _very_ long. A paginated and/or searchable list could go a long way here.
I don't think this should be a blocker on this PR directly, but it should be something we look into in future.

pydis_site/apps/content/utils.py
HassanAbouelela: Hahaha, I'm actually working on making the cropping here better, just keep your eyes peeled.

pydis_site/apps/content/views/tags.py
HassanAbouelela: Yes, that's what I originally did, but how do you do the substitution? Substituting by index would not work after the first match since the index would've moved, and substituting with something like re.sub wouldn't work either since the substitution still matches, so you'll end up on the first URL again.

pydis_site/apps/content/utils.py
HassanAbouelela: It's not easy to fix this, at least in python, without effectively writing a full markdown interpreter. What I'm doing instead is re-writing this to do the splitting in javascript, where the text is already separated out in a convenient manner.

pydis_site/apps/content/views/tags.py
ChrisLovering: I made this suggestion without even looking at what this was doing. re.sub could be used here afaik, if you make the replacement a func rather than a string.
```py
def linkify_tag_ref(match: re.Match) -> str:
    link = reverse("content:tag", kwargs={"name": match.group("name")})
    return f"[{match.group()}]({link})"

...

content = COMMAND_REGEX.sub(linkify_tag_ref, content):
```

pydis_site/apps/content/models/tag.py
```diff
@@ -0,0 +1,17 @@
...
+        help_text="The date and time this data was last fetched.",
+        auto_now=True,
+    )
+    name = models.CharField(
+        help_text="The tag's name.",
+        primary_key=True,
+        max_length=50,
+    )
+    body = models.TextField(help_text="The content of the tag.")
+    url = models.URLField(help_text="The URL to this tag on GitHub.")
```
Numerlor: does this need to be its own field instead of using the name and a configured github url?

Numerlor (changes requested):
It looks like this is currently missing proper support for tag groups (tags in directories under `tags/`), as there can be multiple tags with the same name under different dirs, and the github url template skips groups completely.

pydis_site/apps/content/models/tag.py
HassanAbouelela: It's a remnant from when I was still using the API and it'd return the URL. If we leave things as is, I'll drop it, but it might come in handy for the groups.

HassanAbouelela:
@Numerlor I was not aware of the tag groups, is there an example of them currently?

Numerlor:
> @Numerlor I was not aware of the tag groups, is there an example of them currently?

I don't think any are currently used or that there are any real examples, but they're created through the directories I mentioned:
```
tags/
├── tag_group1/
│   └── tag.md
├── tag_group2/
│   └── another_tag.md
└── plain_tag_without_group.md
```
this structure would create two tag groups `tag_group1` and `tag_group2` with `tag.md` and `another_tag.md` in them respectively

HassanAbouelela:
Thanks for the info, I'll work on implementing it

pydis_site/apps/content/views/tags.py
HassanAbouelela: 74226511

pydis_site/apps/content/utils.py
HassanAbouelela: f2ad3eed

pydis_site/apps/content/utils.py
HassanAbouelela: Stale due to f2ad3eed

HassanAbouelela:
Tag groups added in 45cdb27a. Due to the way tags are loaded in static previews, it was pretty easy to add in some under a fake group. Look for "very-cool-group" (these are sorted alphabetically like all the other tags, but we can place tag groups at the top of the list if we want to).

Live example: https://deploy-preview-763--pydis-static.netlify.app/pages/tags/very-cool-group/

pydis_site/apps/content/models/tag.py
HassanAbouelela: Ended up making it a property in b4911d0

pydis_site/apps/content/migrations/0002_remove_tag_url_tag_group.py
```diff
@@ -0,0 +1,22 @@
...
+        migrations.RemoveField(
+            model_name='tag',
+            name='url',
+        ),
+        migrations.AddField(
+            model_name='tag',
+            name='group',
+            field=models.CharField(help_text='The group the tag belongs to.', max_length=50, null=True),
+        ),
+    ]
```
ChrisLovering: Do we really need a second migration file for un-merged changes?

Probably better to merge the two migrations together, so that there's only 1 migration to run after merge.

Numerlor:
> Ideally, I'd also like to add a bit more flavor-info to the pages, such as when they were last updated, and the authors that have worked on them, but the only way to get that data from the API is to request data for a specific commit, which means making ~100 requests in total. We can't do that without being rate-limited, but it could be something to consider. I also considered downloading the git repo (as opposed to the content only), but it doesn't seem the API returns that, we'd need to clone with git.

This should be relatively simple to cache; the `contents` api on a directory returns all the files with their git blob SHA. If that is saved the individual requests for commits to a file can only be done if the new SHA differs from the saved one. Full content to the contents endpoint could also be avoided as it provides the ETag header but the download size there should be fairly small

HassanAbouelela:
The cache solution sounds smart, I'll look into that. In terms of downloading, I'm not sure there's a good way to bulk download files without downloading the entire repo, did you have a solution in mind?

pydis_site/apps/content/migrations/0002_remove_tag_url_tag_group.py
HassanAbouelela: I figured it was easiest to write, and easiest to use for reviewers, but yeah we can merge them.

Numerlor:
> The cache solution sounds smart, I'll look into that. In terms of downloading, I'm not sure there's a good way to bulk download files without downloading the entire repo, did you have a solution in mind?

I think the best solution is to just use the raw download urls from the contents api call. I don't believe count into the rate limits either so the only disadvantage is that is needs to be individual requests instead of bulk for the changed files, but that should only be a disadvantage when filling up the cache, as after that the changes will be mostly to individual files.

The GraphQL api can fetch all the files and their contents together, but I'm not sure if that can then be further filtered when not all files are necessary, and it needs auth

HassanAbouelela:
I added in the commit stuff, I think it turned out decently.

pydis_site/apps/content/migrations/0002_remove_tag_url_tag_group.py
HassanAbouelela: f2374900c4c83097c105b56de02ea82d66bd9466

pydis_site/apps/content/models/tag.py
```diff
@@ -0,0 +1,75 @@
+import collections.abc
+import json
+
+from django.db import models
+
+
+class Commit(models.Model):
+    """A git commit."""
```
ChrisLovering: Possibly worth changing this to specify that it's a commit from the bot repo, since we're hard coding the URL

pydis_site/apps/content/models/tag.py
```diff
@@ -0,0 +1,75 @@
...
+    URL_BASE = "https://github.com/python-discord/bot/commit/"
+
+    sha = models.CharField(
+        help_text="The SHA hash of this commit.",
+        primary_key=True,
+        max_length=40,
+    )
+    message = models.TextField(help_text="The commit message.")
+    date = models.DateTimeField(help_text="The date and time the commit was created.")
+    author = models.TextField(help_text="The person(s) who created the commit.")
```
ChrisLovering: Is it worth pluralising this to make it clear it's a list of authors, rather than a single author. From the format_users function, it seems that this is a serialised list of dicts? If so, probably worth mentioning that in the help text too, maybe even a comment describing the structure, or a link to the docs.

pydis_site/apps/content/models/tag.py
```diff
@@ -0,0 +1,75 @@
...
+        """The URL to the commit on GitHub."""
+        return self.URL_BASE + self.sha
+
+    def lines(self) -> collections.abc.Iterable[str]:
+        """Return each line in the commit message."""
+        for line in self.message.split("\n"):
+            yield line
+
+    def format_users(self) -> collections.abc.Iterable[str]:
+        """Return a nice representation of the user(s)' name and email."""
```
ChrisLovering: Should this be format_authors, rather than format_users, so that we're not referring to it in 2 different ways

pydis_site/apps/content/models/tag.py
```diff
@@ -0,0 +1,75 @@
...
+        for author in json.loads(self.author):
+            yield f"{author['name']} <{author['email']}>"
+
+
+class Tag(models.Model):
+    """A tag from the python-discord server."""
```
ChrisLovering: Since we're hard coding this to the bot repo, should this instead be
```suggestion
    """A tag from the python-discord bot repository."""
```

pydis_site/apps/content/models/tag.py
```diff
@@ -0,0 +1,75 @@
...
+        auto_now=True,
+    )
+    sha = models.CharField(
+        help_text="The tag's hash, as calculated by GitHub.",
+        max_length=40,
+    )
+    last_commit = models.ForeignKey(
+        Commit,
+        help_text="The commit this file was last touched in.",
+        null=True,
```
ChrisLovering, line 58: How can this be nullable?

pydis_site/apps/content/urls.py
```diff
@@ -29,18 +29,39 @@ def __get_all_files(root: Path, folder: typing.Optional[Path] = None) -> list[st
...
+        views.TagView.as_view(),
+        name="tag",
+        distill_func=get_all_tags
+    ),
     distill_path(
         "<path:location>/",
         views.PageOrCategoryView.as_view(),
         name='page_category',
         distill_func=get_all_pages
-    ),
```
ChrisLovering: Is this diff needed?

pydis_site/apps/content/utils.py
```diff
@@ -1,22 +1,50 @@
...
 import yaml
 from django.http import Http404
+from django.utils import timezone
 from markdown.extensions.toc import TocExtension
 
+from pydis_site import settings
+from .models import Commit, Tag
 
-def get_category(path: Path) -> Dict[str, str]:
+TAG_CACHE_TTL = datetime.timedelta(hours=1)
```
ChrisLovering, line 20: This is fine as is. However, what are your thoughts on using the TTLCache from [cachetools](https://pypi.org/project/cachetools/)? Just mentioning it was it's simply a deco around the function, as shown in the 3rd function [here](https://github.com/tkem/cachetools/#cachetools)

pydis_site/apps/content/views/tags.py
```diff
@@ -0,0 +1,104 @@
...
+import markdown
+from django.conf import settings
+from django.http import Http404
+from django.urls import reverse
+from django.views.generic import TemplateView
+
+from pydis_site.apps.content import utils
+from pydis_site.apps.content.models import Tag
+
+COMMAND_REGEX = re.compile(r"`*!tags? (?P<name>[\w\d-]+)`*")
```
ChrisLovering: Tag groups are separated by spaces, rather than dashes. The below may work, but haven't tested if it causes any other oddities due to the space.
```suggestion
COMMAND_REGEX = re.compile(r"`*!tags? (?P<name>[\w\d ]+)`*")
```

ChrisLovering (changes requested):
A couple of comments, only the regex one is required, the rest are just observations.

Haven't reviewed js, css or html directly, but it looks good in the app itself.

pydis_site/apps/content/models/tag.py
HassanAbouelela: I called it author to follow the naming scheme of the API object, but it probably makes more sense to properly title it. In terms of the documentation, there's also the option of making the author another model, though it really only carries two important bits of info. Another option is to encode the author as a single string from the get-go if we wanted to avoid converting to and from JSON.

pydis_site/apps/content/models/tag.py
HassanAbouelela: The initial value can be null, since the data is only populated when requesting a single tag page (to avoid making 100 requests at once). Once the tag is accessed for the first time (since the last update), it's commit data is saved for future access.

HassanAbouelela: Relevant code in the get_tag method:
https://github.com/python-discord/site/blob/92a42694b6ad1a29e5a21e0b3e57639528837113/pydis_site/apps/content/utils.py#L245-L248

pydis_site/apps/content/urls.py
HassanAbouelela: It wasn't intentional, I'll see about removing it.

pydis_site/apps/content/views/tags.py
HassanAbouelela: Hmm, that dash is for tags, not tag groups, so I think it's still necessary (for instance `args-kwargs`). I hadn't considered tag groups at all when writing this, but I don't think they're separated by spaces either, I think the tag code does not support that. The relevant code for parsing tag commands is:

https://github.com/python-discord/bot/blob/dc9cb7910040f32d91aa546eec23b942a5a72203/bot/exts/info/tags.py#L66-L68


which leads me to believe that tags currently have to follow the `(tag-group)? tag-name` structure. I think it's simplest to write a regex which captures everything after `!tag` and handle the logic in the replacement function.

pydis_site/apps/content/utils.py
HassanAbouelela: What would we apply the cache to here? The code checks the TTL against items from the database.

pydis_site/apps/content/models/tag.py
HassanAbouelela: There's also the option of just not caring, it's unusual for the commit author to not be the committer as well, so it's not very important info to display.

pydis_site/apps/content/models/tag.py
HassanAbouelela: Yeah that's a fair point, c0384c62

pydis_site/apps/content/models/tag.py
HassanAbouelela: c0384c62

pydis_site/apps/content/urls.py
HassanAbouelela: c0384c62

pydis_site/apps/content/utils.py
ChrisLovering: Ah, this was my missunderstanding, I thought the ttl was used for an lru cache, but now I see what you mean. Ignore this.

pydis_site/apps/content/views/tags.py
ChrisLovering: Yea that seems reasonable, when i said separated by spaces I meant the group and the tag itself, I forgot tags could have hyphens in them too.

pydis_site/apps/content/models/tag.py
ChrisLovering: yea I don't think we need to go as far as another model, I was just surprised when reviewing that a field titled `author` was actually a serialized json structure _possibly_ containing multiple authors with additional info.

pydis_site/apps/content/views/tags.py
HassanAbouelela: Actually it might be pretty difficult to accurately discern what's a tag group, and what's a tag with a normal sentence after it (for instance this tag link at the bottom of [args-kwargs](https://deploy-preview-763--pydis-static.netlify.app/pages/tags/args-kwargs/) which has a sentence after it). The only reliable way would be to either make a request to the page (reverse does not validate kwargs), or to use the get_tag function, but that can make an extra request (commit data) if the tag is not part of a group, neither of which are ideal solutions.

HassanAbouelela: This regex achieves the first part of that:
```regex
`*!tags? (?P<first>[\w-]+)(?P<second> [\w-]+)?`*
```

It might optionally match a second word after the first one, and if it does, that could be a tag group combination

ChrisLovering: Seems like it would do the job, obviously we'd need to update the tag model and the logic to get the url. So I'd be happy to merge this and have that as a separate PR if needed. Especially since we don't have any tag groups atm anyway.

ChrisLovering (approved):

pydis_site/apps/content/views/tags.py
HassanAbouelela: I wrote the code for it, that was pretty simple, but I'm having unrelated trouble getting the test suite running. I'll probably have the fix out tomorrow

HassanAbouelela: Fixed in efe784fe

pydis_site/apps/content/utils.py
HassanAbouelela: Fair enough, thanks for the library suggestion though, it looks cool 😀

pydis_site/apps/content/models/tag.py
HassanAbouelela: 2343f7ab

pydis_site/apps/content/urls.py
```diff
@@ -29,14 +29,35 @@ def __get_all_files(root: Path, folder: typing.Optional[Path] = None) -> list[st
...
+
+def get_all_pages() -> DISTILL_RETURN:
     """Yield a dict of all page categories."""
     for location in __get_all_files(Path("pydis_site", "apps", "content", "resources")):
         yield {"location": location}
 
 
+def get_all_tags() -> DISTILL_RETURN:
+    """Return all tag names and groups in static builds."""
+    groups = {None}
```
jchristgit, line 45: A comment couldn't hurt here as to why we start with `{None}`. I can understand it, as Chris has lovingly explained it to me.

pydis_site/apps/content/utils.py
```diff
@@ -27,8 +55,258 @@ def get_categories(path: Path) -> Dict[str, Dict]:
...
+            files = [entry]
+
+        for file in files:
+            hashes[file["name"]] = file["sha"]
+
+    # Download the files
+    tar_file = client.get("/repos/python-discord/bot/tarball")
+    tar_file.raise_for_status()
+
+    client.close()
```
jchristgit: Why not use `with`?

pydis_site/apps/content/utils.py
```diff
@@ -27,8 +55,258 @@ def get_categories(path: Path) -> Dict[str, Dict]:
...
+        except Tag.DoesNotExist:
+            # The tag is not in the database yet,
+            # pretend it's previous state is the current state
+            old_tag = tag
+
+        if old_tag.sha == tag.sha and old_tag.last_commit is not None:
+            # We still have an up-to-date commit entry
+            tag.last_commit = old_tag.last_commit
+
+        tag.save()
```
jchristgit: This part does not make sense to me. If the hash of the file changes, shouldn't we update it here? We don't seem to throw it away in that case either, but just stop updating.

pydis_site/apps/content/utils.py
```diff
@@ -27,8 +55,258 @@ def get_categories(path: Path) -> Dict[str, Dict]:
...
+
+    # Drop old, unused commits
+    Commit.objects.filter(tag__isnull=True).delete()
+
+
+def get_tags() -> list[Tag]:
```
jchristgit, line 196: This typehint is wrong, because `return Tag.objects.all()` returns a `QuerySet`, **not** a `list[Tag]`. We could use `Iterable[Tag]` or possibly even `Container[Tag]`, but `list[Tag]` is wrong.

pydis_site/apps/content/utils.py
```diff
@@ -27,8 +55,258 @@ def get_categories(path: Path) -> Dict[str, Dict]:
...
+    an extra request will be made to sync the information.
+
+    The tag name and group must match. If only one argument is provided in the path,
+    it's assumed to either be a group name, or a no-group tag name.
+
+    If it's a group name, a list of tags which belong to it is returned.
+    """
+    path = path.split("/")
+    if len(path) == 2:
+        group, name = path[0], path[1]
```
jchristgit:
```suggestion
        group, name = path
```

pydis_site/apps/content/utils.py
```diff
@@ -27,8 +55,258 @@ def get_categories(path: Path) -> Dict[str, Dict]:
...
+                groups[tag.group]["description"].append(tag.name)
+
+    # Flatten group description into a single string
+    for group in groups.values():
+        # If the following string is updated, make sure to update it in the frontend JS as well
+        group["description"] = "Contains the following tags: " + ", ".join(group["description"])
+        data.append(group)
+
+    # Sort the tags, and return them in the proper format
+    return {tag["title"]: tag for tag in sorted(data, key=lambda tag: tag["title"].lower())}
```
jchristgit: Use `str.casefold()` instead of `.lower()` for key functions.

```suggestion
    return {tag["title"]: tag for tag in sorted(data, key=lambda tag: tag["title"].casefold())}
```

jchristgit: This may cause one of the 200 dictator linters to throw up due to line length.

pydis_site/apps/content/views/tags.py
```diff
@@ -0,0 +1,124 @@
...
+
+from pydis_site.apps.content import utils
+from pydis_site.apps.content.models import Tag
+
+# The following regex tries to parse a tag command
+# It'll read up to two words seperated by spaces
+# If the command does not include a group, the tag name will be in the `first` group
+# If there's a second word after the command, or if there's a tag group, extra logic
+# is necessary to determine whether it's a tag with a group, or a tag with text after it
+COMMAND_REGEX = re.compile(r"`*!tags? (?P<first>[\w-]+)(?P<second> [\w-]+)?`*")
```
jchristgit, line 19: This is a great comment, and I love it very much.

jchristgit (changes requested):
Legend man. LEGEND!

I have small comments. If the comments are introspected, I am approving.

pydis_site/apps/content/utils.py
HassanAbouelela: We don't fetch the hashes here as documented in the approach methodology in the PR body since it would be too expensive in terms of ratelimits. There are ~100 tags, and it doesn't make sense to make an additional 100 requests to pre-load all the commits. When someone requests a tag, that commit data is fetched and saved, otherwise no commit data will be available until then.

jchristgit (approved):
Great Behaviour From Joe Banks In The Dog Park Today.

pydis_site/apps/content/views/tags.py
HassanAbouelela: Thank you :D

pydis_site/apps/content/utils.py
HassanAbouelela: 426fc71b

pydis_site/apps/content/utils.py
HassanAbouelela: 153059c

pydis_site/apps/content/utils.py
HassanAbouelela: e8b3745c

pydis_site/apps/content/urls.py
HassanAbouelela: 3940e34b

PR #765, opened 2022-08-15
Migrate on_command_error pin

Robin5605:
As part of the migration of the #discord-bots pinned content from discord to the site, this PR migrates the pin regarding `on_command_error` of the discord.py library "eating" (silencing) unhandled errors.

This PR also adds a few extra items, including:
- General markdown improvements, formatting
- Fixing a few inaccuracies (e.g generic `commands.BagArgument` has been changed to `commands.MemberNotFound`
- An `elif` statement handling another error, showing users that they can handle any number of errors they want to
- Showing how to use specific properties of the errors `commands.MemberNotFound.argument`, `commands.MissingRequiredArgument.param`
- Implementing the `on_command_error` function with and without a bot subclass, as the function in itself does not do anything, and needs to be bound to the bot somehow


jchristgit (changes requested):
This looks good, but I think we should be using four-space indents here, shouldn't we?

Xithrius (approved):

PR #768, opened 2022-08-18
Fix Warnings & Enable In Tests

HassanAbouelela:
While working on upgrading some of our dependencies, I started running into issues due to us continuing to use deprecated features. This includes:
- django-filter. The history behind django-filter, and our usage off it is a little complicated. I tried to summarize in 6fe491c, but I can elaborate/explain further if needed. The warnings go from deprecations to removal in the current latest version, which is where I ran into issues.
- Django 5.x: django actually emits a helpful deprecation warning for features that are removed in 5.x. There are a few that have been added since the initial upgrade, so those were updated here. They are very simple things
- Unittest's `assertEquals` has been deprecated since 3.2, in favor of `assertEqual`.

With these changes made, there were no other warnings remaining, so I took the liberty of enabling warnings in CI, as a relatively harmless method of detecting some of these things earlier.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -7,6 +7,7 @@ from pathlib import Path
 import django
 from django.contrib.auth import get_user_model
 from django.core.management import call_command, execute_from_command_line
+from django.test.utils import ignore_warnings
 
 DEFAULT_ENVS = {
     "DJANGO_SETTINGS_MODULE": "pydis_site.settings",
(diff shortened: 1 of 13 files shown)


ChrisLovering (approved):

jchristgit (approved):

manage.py
```diff
@@ -154,7 +155,14 @@ def run_server(self) -> None:
     def run_tests(self) -> None:
         """Prepare and run the test suite."""
         self.prepare_environment()
-        call_command(*sys.argv[1:])
+        # The way we run tests in CI, and whitenoise's behavior raise a warning
+        # since the static root folder does not need to exist
+        # Reference: https://github.com/evansd/whitenoise/issues/215
```
MarkKoz: Can you rephrase this? The clause before the comma seems incomplete.

MarkKoz (changes requested):
Good initiative.

manage.py
HassanAbouelela: Fair enough, 74326128. Thanks for the review!

MarkKoz (approved):

PR #769, opened 2022-08-18
Update To 3.10 And Bump Dependencies

HassanAbouelela:
~Blocked by #768, which this branch is based on.~

This PR updates the python version to 3.10, as we've done elsewhere in our repositories, and bumps and locks all dependencies to latest versions, using exact specifiers. Using exact specifiers is part of a larger push by the dev-core team for better dependency management, along with the addition of dependabot.

To keep things as simple as possible to review, the commits are organized as follows:
- 8210299 updates all normal dependencies, which thanks to PR #768 requires no changes beyond changing the version numbers. This is missing one important update however, which is for django 4.1.1. That is postponed due to a very critical bug in 4.1.1 which breaks a lot of our models. The issue is linked in the code base.
- c175fc76 bumps coverage. The breaking change in v6 is dropping support for older python versions, so it requires no real changes. It did introduce a new warning, which is triggered due to a misconfiguration in our code. I fixed the misconfiguration in this commit.
- e26d6b8c bumps dev-dependencies, and similar to the other bump commit, it only requires updating the version numbers. flake8-bandit was disabled for the reason written in the pyproject file.
- a5fb5bf1 bumps the python version to 3.10. This is mostly just changing a digit here and there, as well as some doc changes. I kaizened in a doc change for bot which also recently upgraded to 3.10.
- 98cea83f adds dependabot
- e6601506 finally removed the coveralls dependency. This is not a silver bullet which resolves all the issues we've had with coveralls, but it does decouple it from our code, which should alleviate some difficulties. This change also has the nice benefit of showing up on commits as well (alongside the other action names).


HassanAbouelela:
Just opened this PR to keep track of CI as I make the upgrade, it's not ready yet.

HassanAbouelela:
Django distill seems to be failing on update from 2.9 -> 2.10. I've narrowed it down to this commit https://github.com/meeb/django-distill/commit/5abe55b78d2024df4f3150b0e16459759cb329b1. I'll investigate later.

HassanAbouelela:
Nearly everything is ready now, just need to figure out why coveralls isn't working again

ChrisLovering (approved):
This is great scaleios, thanks!

jchristgit (approved):

HassanAbouelela:
I've gone ahead and cleaned history, merged the main branch, and updated all dependencies again (including both dependencies which were waiting on fixes upstream).

This also means @jchristgit review should count since I snapped your changes out of existence 😀, or you can review again if you like

I've noticed that flake is failing locally when run through pre-commit (something we wouldn't see in CI), so I'm working on finding a fix for that now

HassanAbouelela:
Alright, everything seems to be working, but the test suite took 4+ minutes in CI (still taking the correct amount of time locally)...

HassanAbouelela:
I've narrowed down the test slowness to the redirects tests, specifically on resource redirects which take 15 seconds per redirect in the django render function (only if running through the django test runner).

Downgrading sentry from 1.9.9 to 1.9.8 fixes the issue. Based on their issue tracker, 1.9.9 is broken big time.

ChrisLovering (approved):

PR #789, opened 2022-10-27
Messages in past n days endpoint

### LINKED ISSUE #783: Endpoint for getting activity from metricity
This issue is required for https://github.com/python-discord/bot/issues/2302

The endpoint should take a list of user IDs and a number of days (or just a timestamp?),  and return the number of messages sent by each user from now up to the date/interval provided. This data is available through metricity.

The reason I suggest an endpoint taking multiple user IDs at once is that I assume it would be more efficient as a single query, as we may want to use this endpoint on all nominated users at once (e.g. to check for users in the talent pool that haven't been active for a long time so can be removed).

wookie184:
A GET endpoint that takes data from the body is a bit dodgy which made this a pain 😩 . Am open to any improvements, I considered using a POST but not sure it's worth changing at this point.

Closes #783


jchristgit:
This was still being worked out re: conversation in #dev-core, right?

wookie184:
Yeah, I think I'll just change it to be a POST endpoint.

wookie184:
Should be ready for review now

wookie184:
Thoughts on changing this to return 0 for users with no messages?

I didn't do it originally because it would also return 0 for *invalid* IDs which could potentially be misleading, but now i'm thinking it would probably be more useful if it just returned 0 for the ID if nothing was found.

shtlrs:
I guess we could have the user ids as comma-separated-values, but we'd be running the risk of having too long query strings :/
I think it's fine to have this as a POST endpoint, even though it "sounds" wrong.

The only way to make this a GET would be to fetch the data of one user, but it would basically be a waste of bandwidth.

I'm saying this because I don't know what was discussed in #dev-core, so just shooting out ideas.

shtlrs:
> Thoughts on changing this to return 0 for users with no messages?
> 
> I didn't do it originally because it would also return 0 for _invalid_ IDs which could potentially be misleading, but now I'm thinking it would probably be more useful if it just returned 0 for the ID if nothing was found.

I think it would be a good idea to return 0 for users with no messages.

Currently, if at least one id is invalid, the request won't be successful and will return with the encountered errors.

image

Unless there's a detail that I'm missing

wookie184:
I was referring to the case where we can't tell if the ID is correct or not. For cases where it's not an integer or is negative the API user has clearly done something wrong so I think it makes sense to return an error response.

The issue is that with the current query we can't tell between users with no messages in that period and users that don't exist (but with the ID in the correct form). I think i'll change it so that does return 0 anyway though, as it's probably more useful.

shtlrs:
> I was referring to the case where we can't tell if the ID is correct or not. For cases where it's not an integer or is negative the API user has clearly done something wrong so I think it makes sense to return an error response.
> 
> The issue is that with the current query we can't tell between users with no messages in that period and users that don't exist (but with the ID in the correct form). I think i'll change it so that does return 0 anyway though, as it's probably more useful.

When would we have the case of a non existent ID ? 
Because what I'm wondering is, if the user is a member of server but not found in the database, we would have a delta between the number of items returned by the server & the number of items requested.

And if the user is not in the database, then the query won't return any information about them since they don't exist.

What I'm thinking is, if we "desperately" want that information, we can have some sort of query post-processing before the API returns a response, and non existent user will have a value of -1 for example.

wookie184:
This is ready for review now.

> What I'm thinking is, if we "desperately" want that information, we can have some sort of query post-processing before the API returns a response, and non existent user will have a value of -1 for example.

We don't need that information currently, so it's probably not worth overcomplicating it.

jchristgit (approved):
Legend man. LEGEND.

jchristgit:
This is what the American country has been looking forward to for years. As a single nation, we are finally able to provide a detailed and in-depth aggregation of information on American citizens, which further improves our commitment to a brighter future for our citizens. It is imperative to be aware of our environmental impact, and I am hereby signing this pull request into master.

Xithrius (approved):
I see no issues with the current implementation of this feature.

PR #792, opened 2022-11-06
Add "thread_id" as a column in the nomination table

### LINKED ISSUE #2304: Nominations: Link previous nomination threads
It would be nice if we could add links to previous nomination threads to both !tp history and newly created nomination threads. As it stands, we have to search for the username in the discord thread history (and this isn't particularly great UX nor reliable). 

Storing the thread ID for a nomination sounds like a good first step (and probably worth committing immediately even before other features are complete?), then searching through threads by username to attempt to rediscover old threads. Since usernames are changeable, we might also want to detect "desynced" talentpool history, where we know there have been X previous nominations, but only found less-than-X threads. 

We could also search through threads by time--we know when previous nominations have _ended_, but we don't know when they start, so we can use this also to help figure out which threads might be relevant by `archive_timestamp` or `last_pin_timestamp`, the latter of which might actually be extremely reliable (given reviews are auto-pinned in their thread).

shtlrs:
**Note to core devs**
This one needs to be merged along with this [PR](https://github.com/python-discord/bot/pull/2319)

This is one of the changes required to close [this](https://github.com/python-discord/bot/issues/2304)

We want to be able to traceback threads of earlier nominations for historical purposes.

The purpose is to attach the nomination vote's thread id along with each nomination & display them later on when the history of a particular user is demanded.


shtlrs:
@wookie184 Here's the first iteration of #2304

However, I wanted your opinion on when should we allow editing the `thread_id` of a nomination.
And my guess is, it should follow the rest, e.g.:
1. It shouldn't be allowed when we're ending a nomination
2. It shouldn't be allowed on an inactive nomination

Let me know if I missed something

wookie184:
> However, I wanted your opinion on when should we allow editing the `thread_id` of a nomination. And my guess is, it should follow the rest, e.g.:
> 
>     1. It shouldn't be allowed when we're ending a nomination
> 
>     2. It shouldn't be allowed on an inactive nomination
> 
> 
> Let me know if I missed something

Hmm, I hadn't thought about validation. I don't really mind and don't think it matters too much, although those two things sound good for now. We can always make it more/less strict if we need it.

pydis_site/apps/api/viewsets/bot/nomination.py
```diff
@@ -282,12 +287,15 @@ def partial_update(self, request: HttpRequest, *args, **kwargs) -> Response:
             )
 
         # This is actually covered, but for some reason coverage don't think so.
```
wookie184: I think this comment can be removed, as coverage seems to detect the code fine now.

wookie184 (commented):
I haven't tried it out yet but this looks great, thanks!

pydis_site/apps/api/viewsets/bot/nomination.py
shtlrs: That's true, I thought I removed it.
Yoinked !

pydis_site/apps/api/models/bot/nomination.py
```diff
@@ -35,6 +35,11 @@ class Nomination(ModelReprMixin, models.Model):
         default=False,
         help_text="Whether a review was made."
     )
+    thread_id = models.BigIntegerField(
+        help_text="The nomination vote's thread id.",
+        null=True,
+        blank=True
```
jchristgit: `blank=True` only makes sense on string fields, as far as I'm aware? https://docs.djangoproject.com/en/4.1/ref/models/fields/#blank

shtlrs: You're correct, it's mostly used for validation purposes.

Having it as nullable should suffice. 

I'll remove it

jchristgit (approved):
Thanks!

wookie184 (approved):
LGTM

wookie184:
marked do not merge pending bot PR, not sure if it would actually cause any issues merging before but easiest to wait rather than risk it.

Xithrius (approved):
Extra approval for good luck

PR #808, opened 2022-12-02
Rename vps_services.md to vps-services.md

Robin5605:
Rename `vps_services.md` to `vps-services.md` (notice the dash) to not have to update old messages linking here
[Link to discussion](https://discord.com/channels/267624335836053506/551789653284356126/1048372967597424690)


minalike (approved):
I also think hyphens are just nicer than underscores when it comes to URLs, but that's just me.

PR #824, opened 2022-12-31
Add jump_url field to infraction model

### LINKED ISSUE #801: Add field to Infraction model for jump link to infraction message.
Relevant bot issue: [bot#2329](https://github.com/python-discord/bot/issues/2329)

vivekashok1221:
Closes #801 .

Related bot issue: python-discord/bot#2329
Related bot PR: python-discord/bot#2372


pydis_site/apps/api/migrations/0086_merge_20221231_0822.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0085_add_thread_id_to_nominations'),
+        ('api', '0085_infraction_jump_url_text'),
+    ]
+
+    operations = [
```
shtlrs: Is this file supposed to exist ?
It looks like it came from a merge, but it's strange since the two dependencies have the same identifier.
I think deleting these 2 & making a new migration would be cleaner

pydis_site/apps/api/models/bot/infraction.py
```diff
@@ -69,6 +69,14 @@ class Infraction(ModelReprMixin, models.Model):
         help_text="Whether a DM was sent to the user when infraction was applied."
     )
 
+    jump_url_text = models.CharField(
```
shtlrs: Wouldn't it be better to store the raw URL in the database ? And do the formatting on the bot's side ?

shtlrs (changes requested):
This is purely a code review, will test later!
Thanks for the PR vivek ! :D

pydis_site/apps/api/migrations/0086_merge_20221231_0822.py
vivekashok1221: This file's the result of running `python manage.py makemigrations --merge`.

shtlrs: Ah I see, I didn't know this existed.
Don't you think it'd be cleaner to have a separate one, with a different id ? instead of this merged format ?

pydis_site/apps/api/models/bot/infraction.py
vivekashok1221: I was initially doing what you described but changed to this because it looked cleaner bot-side. But I'll do what you're suggesting because it will be more modular.

pydis_site/apps/api/migrations/0086_merge_20221231_0822.py
vivekashok1221: Similar merged migration files already exist in the project so imo it's fine.

shtlrs: Sure, that's one way to view this.
Even though I don't fully agree, since having something that can or could've been improved doesn't mean we should do it again.
I think the conflict happened because of the migration I did on the nominations table, which has already been merged into main 2 days ago.
And the conflict happened just on the naming, and not on the model itself, and I still think it should be done on its own.

We can wait for others to see what they think about this.

vivekashok1221: Migration files are not for humans anyway. However, I've gone ahead and generated a new migration file.

pydis_site/apps/api/models/bot/infraction.py
vivekashok1221: Done.

pydis_site/apps/api/models/bot/infraction.py
```diff
@@ -69,6 +69,14 @@ class Infraction(ModelReprMixin, models.Model):
         help_text="Whether a DM was sent to the user when infraction was applied."
     )
 
+    jump_url = models.CharField(
+        default='',
+        max_length=88,
```
wookie184, line 75: This is pretty tight, I can't think of any valid reason it should be longer than this so I guess it's fine though...

wookie184 (approved):
LGTM

pydis_site/apps/api/models/bot/infraction.py
```diff
@@ -69,6 +69,14 @@ class Infraction(ModelReprMixin, models.Model):
         help_text="Whether a DM was sent to the user when infraction was applied."
     )
 
+    jump_url = models.CharField(
+        default='',
```
ChrisLovering: What was the reason behind using an empty string here, rather than making it nullable?

vivekashok1221: Replied here: https://github.com/python-discord/bot/pull/2372#discussion_r1118095635

jchristgit (approved):
🚢 🇮🇹

pydis_site/apps/api/models/bot/infraction.py
vivekashok1221: I've now replaced CharField with URLField and added `null=True` in https://github.com/python-discord/site/pull/824/commits/ed3577652d620057cc87bbc74775548066cfdc8f.

PR #837, opened 2023-01-14
Make dependabot check for github actions updates

wookie184:
Noticed that there were some warnings to update in some workflow runs, so seems worth automating.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot


ChrisLovering (approved):

Xithrius (approved):
very neat

PR #854, opened 2023-01-18
Stop using deprecated set-output command in actions

wookie184:
GitHub has deprecated the old format, so I've updated to use the new syntax:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,7 +20,7 @@ jobs:
         id: sha_tag
         run: |
           tag=$(cut -c 1-7 <<< $GITHUB_SHA)
-          echo "::set-output name=tag::$tag"
+          echo "tag=$tag" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


ChrisLovering (approved):
Looks good. We should migrate to reusable flows soon though, see https://github.com/python-discord/black-knight/tree/main/.github/workflows for reference

MarkKoz (approved):

wookie184:
> Looks good. We should migrate to reusable flows soon though, see https://github.com/python-discord/black-knight/tree/main/.github/workflows for reference

Ah, yeah, that does seems nice. 

Also, I wonder if it would be worth having a repo of workflows we can reuse across the org. The build/deploy/status_embed workflows seem very similar wherever they're used. Not sure if the differences are enough that in practice it would be difficult though.

PR #860, opened 2023-01-24
yaml for python.org tutorial

### LINKED ISSUE #852: Add the Python Org's Python Tutorial to Resources
The resources page does not include the tutorial linked on the python.org site [here](https://docs.python.org/3/tutorial/)

I think we should include it as a free tutorial source.

Assile:
Possible fix for #852


brad90four (approved):
Looks good to me and addresses all of the information that should be present.

wookie184 (approved):
Looks good, thanks!

PR #861, opened 2023-01-28
New Filtering System

### LINKED ISSUE #479: New Filter schemas

mbaruh:
This pull request applies the necessary changes to the database and API to support the new filtering system on the bot.

For more information on how the information in these tables will be used, check out the bot PR: https://github.com/python-discord/bot/pull/2390

> **Warning** This PR must be merged at the same time as the bot PR.

closes #479.

## Scheme
image

* There is a unique constraint on combinations of name, list_type in FilterList.
* There is a unique constraint on combinations of content, additional_field, filter_list and the settings (everything below filter_list) in Filter.
  * This uses the NULLS NOT DISTINCT feature introduced in psql 15, as two filters in the same filter list with null for some setting have in practice the same value (the value for that setting will be taken from the same filter list).

This scheme is different from the one in the related issue, and is the result of implementing the suggestions made in #482.

## Serializers
The two models use custom serializers in order to provide the settings in a format the bot can work with.
* All settings are placed under one field, `"settings"`.
* Some settings are grouped together, e.g the four settings relating to channels and categories are under the `"channel_scope"` group.


pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,260 @@
...
+    infraction_reason = models.CharField(
+        max_length=1000,
+        help_text="The reason to give for the infraction.",
+        blank=True,
+        null=False
+    )
+    infraction_duration = models.DurationField(
+        null=False,
+        help_text="The duration of the infraction. Null if permanent."
+    )
```
wookie184: "null if permanent" is impossible given "null=False`

I see the difficulty since for `FilterBase` null means "not overwritten" so we can't really make use of it for other meanings. I guess you've already sorted this on the bot side so the help text just needs changing

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,260 @@
...
+    enabled = models.BooleanField(
+        help_text="Whether this filter is currently enabled.",
+        null=True
+    )
+    send_alert = models.BooleanField(
+        help_text="Whether an alert should be sent.",
+        null=True
+    )
+
+    # Check FilterList model for information about these properties.
```
wookie184: I'm not sure I see anything in the `FilterList` model that isn't here? Help text here seems fine though, fine to leave exact specification to the bot side.

pydis_site/apps/api/viewsets/bot/filters.py
```diff
@@ -0,0 +1,499 @@
...
+    #### Status codes
+    - 200: returned on success
+    - 400: if one of the given fields is invalid
+
+    ### DELETE /bot/filter/filters/<id:int>
+    Deletes the Filter item with the given `id`.
+
+    #### Status codes
+    - 204: returned on success
+    - 404: if a tag with the given `id` does not exist
```
wookie184:
```suggestion
    - 404: if a Filter with the given `id` does not exist
```

pydis_site/apps/api/migrations/0086_new_filter_schema.py
```diff
@@ -0,0 +1,171 @@
...
+from datetime import timedelta
+
+import django.contrib.postgres.fields
+from django.apps.registry import Apps
+from django.core.validators import MinValueValidator
+from django.db import migrations, models
+import django.db.models.deletion
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+import pydis_site.apps.api.models.bot.filters
```
wookie184: Is there a reason this isn't just `import pydis_site.apps.api.models` since that's what's being accessed? Same for other migrations.

Or just a `from` import to get the names directly?

pydis_site/apps/api/viewsets/bot/filters.py
```diff
@@ -0,0 +1,499 @@
...
+    #### Status codes
+    - 200: returned on success
+    - 400: if one of the given fields is invalid
+
+    ### DELETE /bot/filter/filter_lists/<id:int>
+    Deletes the FilterList item with the given `id`.
+
+    #### Status codes
+    - 204: returned on success
+    - 404: if a tag with the given `id` does not exist
```
wookie184:
```suggestion
    - 404: if a FilterList with the given `id` does not exist
```
Caught red handed!

wookie184 (changes requested):
I've not looked at the bot side yet so this is just comments while I was understanding how things work. Looks really good though.

pydis_site/apps/api/serializers.py
```diff
@@ -141,30 +144,270 @@ class Meta:
...
+            field_args['allow_empty'] = True
+        extra_kwargs[field] = field_args
+    return extra_kwargs
+
+
+def get_field_value(data: dict, field_name: str) -> Any:
+    """Get the value directly from the key, or from the filter list if it's missing or is None."""
+    if data.get(field_name):
+        return data[field_name]
+    return getattr(data["filter_list"], field_name)
```
GDWR:
```suggestion
def get_field_value(data: dict, field_name: str) -> Any:
    """Get the value directly from the key, or from the filter list if it's missing or is None."""
    if data.get(field_name) is not None:
        return data[field_name]
    return getattr(data["filter_list"], field_name)
```

This should be checked to be `None`, to avoid a value being falsy or even `False`. Take the example of a filter with the value `enabled=False`. in this case, the `filter_list`'s `enabled` field would be used instead.

ChrisLovering: wouldn't we want `is not None` here? So if it's not None return it, else return the filter list default

GDWR: Maybe, lol. You are correct. I've amended the suggestion. Thankyou.

mbaruh: Good spot [be854fa](https://github.com/python-discord/site/pull/861/commits/be854fa3d34dac7b4b9e96b3736dd61d972f1b79)

pydis_site/apps/api/viewsets/bot/filters.py
mbaruh: 👀 4eecda9

pydis_site/apps/api/migrations/0086_new_filter_schema.py
mbaruh: I have no idea, maybe because it was a semi-generated migration 4eecda9

pydis_site/apps/api/viewsets/bot/filters.py
mbaruh: 4eecda9

pydis_site/apps/api/models/bot/filters.py
mbaruh: This code has gone through so many iteration I have no clue why this was written. 4eecda9

pydis_site/apps/api/models/bot/filters.py
mbaruh: Good spot, should be 0. 4eecda9

pydis_site/apps/api/migrations/0087_unique_constraint_filters.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0086_new_filter_schema'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
```
shtlrs: Shouldn't we make this a reversible operation ?

I say that because all the default migrations are mostly reversible, and for this case, we'd need to provide a an sql script to the `reverse_sql` kwarg

pydis_site/apps/api/migrations/0088_unique_filter_list.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+    rich_embed.save()
+
+    discord_token = filter_.objects.create(
+        content="discord_token",
+        filter_list=list_,
+        filter_dm=False,
+        remove_context=True,
+        dm_content=(
+            "I noticed you posted a seemingly valid Discord API "
```
shtlrs: Looks like the bot is self aware :p

pydis_site/apps/api/migrations/0089_antispam_filter_list.py
```diff
@@ -0,0 +1,52 @@
+from datetime import timedelta
+
+from django.apps.registry import Apps
+from django.db import migrations
+
+import pydis_site.apps.api.models.bot.filters
+
+
+def create_antispam_list(apps: Apps, _):
+    """Create the 'unique' FilterList and its related Filters."""
```
shtlrs:
```suggestion
    """Create the 'antispam' FilterList and its related Filters."""
```

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,259 @@
...
+    )
+    infraction_channel = models.BigIntegerField(
+        validators=(
+            MinValueValidator(
+                limit_value=0,
+                message="Channel IDs cannot be negative."
+            ),
+        ),
+        help_text="Channel in which to send the infraction.",
+        null=False
```
shtlrs, line 65: Just curious, are you passing `False` to  the `null` kwarg for explicitness ? because they default to `False` IIRC

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,259 @@
...
+        help_text="Whether an alert should be sent.",
+    )
+    # Where a filter should apply.
+    enabled_channels = ArrayField(
+        models.CharField(max_length=100),
+        help_text="Channels in which to run the filter even if it's disabled in the category."
+    )
+    disabled_channels = ArrayField(
+        models.CharField(max_length=100),
+        help_text="Channels in which to not run the filter."
```
shtlrs:
```suggestion
        help_text="Channels in which to not run the filter even if it's enabled in the category"
```
I suppose ?

pydis_site/apps/api/serializers.py
```diff
@@ -141,30 +144,270 @@ class Meta:
...
+CHANNEL_SCOPE_FIELDS = (
+    "disabled_channels",
+    "disabled_categories",
+    "enabled_channels",
+    "enabled_categories"
+)
+MENTIONS_FIELDS = ("guild_pings", "dm_pings")
+
+
+def _create_filter_meta_extra_kwargs() -> dict[str, dict[str, bool]]:
```
shtlrs: We can probably condense this one & `_create_filter_list_meta_extra_kwargs` like this
```py
def _create_meta_extra_kwargs(for_filter_list: bool) -> dict[str, dict[str, bool]]:
    """Create the extra kwargs of the for either FilterList serializer's Meta class."""
    extra_kwargs = {}
    for field in SETTINGS_FIELDS:
        field_args = {} if for_filter_list else field_args = {'required': False, 'allow_null': True}
        if field in ALLOW_BLANK_SETTINGS:
            field_args['allow_blank'] = True
        if field in ALLOW_EMPTY_SETTINGS:
            field_args['allow_empty'] = True
        extra_kwargs[field] = field_args
    return extra_kwargs
 ```

and then call them with the correct value of `for_filter_list` in the meta classes

File: pydis_site/apps/api/serializers.py [hunk 13]
```diff
@@ -141,30 +144,270 @@ class Meta:
...
+            raise ValidationError(
+                "Infraction type is required with infraction duration or reason."
+            )
+
+        if (
+            set(get_field_value(data, "disabled_channels"))
+            & set(get_field_value(data, "enabled_channels"))
+        ):
+            raise ValidationError(
+                "You can't have the same value in both enabled and disabled channels lists."
```
shtlrs: It would be painful to send two big lists of enabled & disabled channels that **might** have the common values.

Wouldn't it be better to return the ids that were found in both of them ?

So something like

```py
        if (
            common_channels := set(get_field_value(data, "disabled_channels"))
            & set(get_field_value(data, "enabled_channels"))
        ):
            raise ValidationError(
                f"You can't have the same value in both enabled and disabled channels lists: {list(common_channels)}"
            )
```

Same would go for categories

shtlrs (changes requested):
This looks superb ! 
Just a couple of questions & nitpicks here & there.

I have done some basic api testing and I haven't really noticed anything just yet.
I'll come back to approve this once I'm done with the bot testing part :D

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,259 @@
...
+    )
+    remove_context = models.BooleanField(
+        help_text=(
+            "Whether this filter should remove the context (such as a message) "
+            "triggering it."
+        ),
+        null=False
+    )
+    bypass_roles = ArrayField(
+        models.CharField(max_length=100),
```
jchristgit, line 86: Shouldn't this be a big integer field - because we're dealing with snowflakes?

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,259 @@
...
+            MinValueValidator(
+                limit_value=0,
+                message="Channel IDs cannot be negative."
+            ),
+        ),
+        help_text="Channel in which to send the infraction.",
+        null=False
+    )
+    guild_pings = ArrayField(
+        models.CharField(max_length=100),
```
jchristgit, line 68: Shouldn't this be a big integer field - because we're dealing with snowflakes?

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,259 @@
...
+    enabled = models.BooleanField(
+        help_text="Whether this filter is currently enabled.",
+        null=False
+    )
+    send_alert = models.BooleanField(
+        help_text="Whether an alert should be sent.",
+    )
+    # Where a filter should apply.
+    enabled_channels = ArrayField(
+        models.CharField(max_length=100),
```
jchristgit, line 99: Same snowflake comment applies here and to the following three fields.

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,259 @@
...
+        validators=(
+            MinValueValidator(
+                limit_value=0,
+                message="Channel IDs cannot be negative."
+            ),
+        ),
+        help_text="Channel in which to send the infraction.",
+        null=True
+    )
+    guild_pings = ArrayField(
```
jchristgit, line 182: Here the same questions as to BigIntegerField applies as with the model above.

pydis_site/apps/api/models/bot/filters.py
```diff
@@ -0,0 +1,259 @@
...
+    """
+
+    class Meta:
+        """Metaclass Filter to set the unique constraint."""
+
+        constraints = (
+            UniqueConstraint(
+                fields=tuple(
+                    [field.name for field in FilterBase._meta.fields
+                     if field.name not in ("id", "description", "created_at", "updated_at")]
```
jchristgit: Nitpick, you don't need the `[]` here and can avoid building the intermediate list :)
```suggestion
                fields=tuple(
                    field.name for field in FilterBase._meta.fields
                     if field.name not in ("id", "description", "created_at", "updated_at")
```

pydis_site/apps/api/serializers.py
```diff
@@ -141,30 +144,270 @@ class Meta:
...
+
+# Required fields for custom JSON representation purposes
+BASE_FILTER_FIELDS = (
+    'id', 'created_at', 'updated_at', 'content', 'description', 'additional_field'
+)
+BASE_FILTERLIST_FIELDS = ('id', 'created_at', 'updated_at', 'name', 'list_type')
+BASE_SETTINGS_FIELDS = (
+    "bypass_roles",
+    "filter_dm",
+    "enabled",
```
jchristgit: There is some inconsistency here with the quotation style between above and below. Personally I would think the fields here and below should be single quoted as well, or is there a reason why these differ?

pydis_site/apps/api/serializers.py
```diff
@@ -141,30 +144,270 @@ class Meta:
...
+            field_args['allow_empty'] = True
+        extra_kwargs[field] = field_args
+    return extra_kwargs
+
+
+def get_field_value(data: dict, field_name: str) -> Any:
+    """Get the value directly from the key, or from the filter list if it's missing or is None."""
+    if data.get(field_name) is not None:
+        return data[field_name]
+    return getattr(data["filter_list"], field_name)
```
jchristgit: A tiny bit more quoting inconsistency here.

pydis_site/apps/api/serializers.py
```diff
@@ -141,30 +144,270 @@ class Meta:
...
+            set(get_field_value(data, "disabled_channels"))
+            & set(get_field_value(data, "enabled_channels"))
+        ):
+            raise ValidationError(
+                "You can't have the same value in both enabled and disabled channels lists."
+            )
+
+        if (
+            set(get_field_value(data, "disabled_categories"))
+            & set(get_field_value(data, "enabled_categories"))
```
jchristgit: I believe these should be single quotes.

jchristgit (changes requested):
This looks super good. Thank you for your hard work.

I have some smaller comments: they are mostly related to code style - quotation style mainly. Apart from that, this looks good to merge to me.

pydis_site/apps/api/migrations/0087_unique_constraint_filters.py
mbaruh: How would I reverse it?

pydis_site/apps/api/models/bot/filters.py
mbaruh: I honestly don't remember.

pydis_site/apps/api/models/bot/filters.py
mbaruh: Roles can be specified as names as well.

pydis_site/apps/api/models/bot/filters.py
mbaruh: Pings can be specified as role names as well.

pydis_site/apps/api/models/bot/filters.py
mbaruh: Channels and categories can be specified by their names as well.

pydis_site/apps/api/models/bot/filters.py
mbaruh: Same as the other guild_pings

pydis_site/apps/api/serializers.py
mbaruh: I think it was just created by different people. I don't mind aligning all of them to one style or another.

mbaruh: What makes you prefer single quotes over double quotes here?

pydis_site/apps/api/models/bot/filters.py
jchristgit: Names can change. What about a filter on one of the off-topic channels?

pydis_site/apps/api/models/bot/filters.py
jchristgit: Is a ping a user mention? Those can definitely change. Or what are we referencing in here?

pydis_site/apps/api/serializers.py
jchristgit: I use single quotes for "data" like dictionary keys, and double quotes for longer user-facing strings like log messages. But it's really personal preference: doesn't matter much

pydis_site/apps/api/models/bot/filters.py
mbaruh: It's either an ID or a role name. The meaning is resolved by the bot.

pydis_site/apps/api/models/bot/filters.py
mbaruh: Then for those you can use the channel ID. Name support was added to match the current filters which don't trigger in categories called "CODE JAM". I expect that we will mostly use IDs, but it might come in handy in additional scenarios in the future.

pydis_site/apps/api/migrations/0087_unique_constraint_filters.py
shtlrs: It should be done by adding `reverse_sql="ALTER TABLE api_filter DROP CONSTRAINT unique_filters"` kwarg to the `RunSQL` instance.

The reason I'm asking this is because when testing locally, and then wanting to move to a different branch to test something else, you'll need to apply migrations up until some version prior to yours, which will result in a 
`django.db.migrations.exceptions.IrreversibleError`

This is ok to happen locally, as you can always drop the container & spin up a new one. 

It's also fair to day that I'm not sure this is something that'll ever happen in production, so we might see what others have to say about it before we action it.

mbaruh: I added it anyway, it can't hurt [1385060](https://github.com/python-discord/site/pull/861/commits/138506045acd1cf464dc63d4465d36208801a3ab)

However if you need to switch to another branch then I'm not sure if some of the other migrations are reversible. I use a separate clone for this project.

shtlrs: I was referring to all the filter migrations, but I don't think this'll ever be a concern.
Thanks for doing this already, and i think it's fine to skip the rest.

pydis_site/apps/api/migrations/0089_antispam_filter_list.py
mbaruh: [d656584](https://github.com/python-discord/site/pull/861/commits/d6565845ec07b9d76949da36505b3f53e402d230)

pydis_site/apps/api/models/bot/filters.py
mbaruh: [d656584](https://github.com/python-discord/site/pull/861/commits/d6565845ec07b9d76949da36505b3f53e402d230)

pydis_site/apps/api/serializers.py
mbaruh: [Merge the extra kwargs creation functions](https://github.com/python-discord/site/pull/861/commits/43913623f87329b51cd2a6793e843c30368698aa)

pydis_site/apps/api/serializers.py
mbaruh: [Specify the common elements in the validation errors](https://github.com/python-discord/site/pull/861/commits/36bca58ff336f9d4b797a2c76f08775f9de7e9a7)

pydis_site/apps/api/models/bot/filters.py
mbaruh: [Use consistent quoting style](https://github.com/python-discord/site/pull/861/commits/91b89475913210400cf39884efe37ab5552efbf7)

pydis_site/apps/api/serializers.py
mbaruh: [Use consistent quoting style](https://github.com/python-discord/site/pull/861/commits/91b89475913210400cf39884efe37ab5552efbf7)

pydis_site/apps/api/serializers.py
mbaruh: [Use consistent quoting style](https://github.com/python-discord/site/pull/861/commits/91b89475913210400cf39884efe37ab5552efbf7)

pydis_site/apps/api/serializers.py
mbaruh: [Use consistent quoting style](https://github.com/python-discord/site/pull/861/commits/91b89475913210400cf39884efe37ab5552efbf7)

jchristgit (approved):
Thank you!

wookie184 (approved):
The changes since my last review look good, and the bot side works, so this seems good, thanks!

shtlrs (approved):
One more for extra luck 🍀

PR #879, opened 2023-02-20
Reverse timeline entries so newest entries come first

ichard26:
We agreed internally that this would look better, and especially as I add more entries.

This was automated using this script:

```python
import dataclasses, sys, textwrap
from pathlib import Path

import bs4


@dataclasses.dataclass
class TimelineEntry:
    title: str
    date: str


html_doc = Path(sys.argv[1]).read_text("utf-8")
soup = bs4.BeautifulSoup(html_doc, "html.parser")

entries = []
for item in soup.find_all(class_="cd-timeline__block"):
    entries.append(TimelineEntry(
        title=str(item.find("h2").string),
        date=str(item.find(class_="cd-timeline__date").string),
    ))

entry_html_pieces = []
doc_lines = html_doc.splitlines()
line_index = 0
for entry in entries:
    entry_lines = []
    for line_index, line in enumerate(doc_lines[line_index:], start=line_index):
        if "cd-timeline__block" in line:
            break
    for line_index, line in enumerate(doc_lines[line_index:], start=line_index + 1):
        entry_lines.append(line)
        if line == "      </div>":
            break
    entry_html_pieces.append("\n".join(entry_lines))
    if entry.title not in entry_html_pieces[-1]:
        assert False, f"entry title ({entry.title}) not found!"

print("\n\n".join(entry_html_pieces))
```
I gave it the pre-existing HTML and then copy n' pasted the output in-between the non-timeline-entry elements.

### DIFF
--- a/pydis_site/templates/home/timeline.html
+++ b/pydis_site/templates/home/timeline.html
@@ -13,627 +13,25 @@
   <section class="cd-timeline js-cd-timeline">
     <div class="container max-width-lg cd-timeline__container">
       <div class="cd-timeline__block">
-        <div class="cd-timeline__img cd-timeline__img--picture">
-          <img src="{% static "images/timeline/cd-icon-picture.svg" %}" alt="Picture">
-        </div>
-
-        <div class="cd-timeline__content text-component">
-          <h2>Python Discord is created</h2>
-          <p class="color-contrast-medium"><strong>Joe Banks</strong> becomes one of the owners around 3 days after it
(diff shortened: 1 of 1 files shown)


ichard26:
> **Note**

This may be easier to review locally with `git show` as the diff will contain no red or green lines, indicating that the contents weren't changed, only moved.

MarkKoz (approved):
Had a look at it in netlify and it all looks good.

ChrisLovering (approved):

PR #899, opened 2023-03-06
Bundle workflows into a main.yaml file

shtlrs:
This bundles several workflows that we have into a main one, in order to have a better "overview" of our CI pipelines

**Changelog**
1. Merge  build & deploy workflows
2. Create a `main.yaml` workflow that calls all the rest
3. Trigger Status Embed workflow from main.
4. Trigger Static preview from main


.github/workflows/main.yaml
```diff
@@ -0,0 +1,54 @@
...
+  cancel-in-progress: true
+
+
+jobs:
+  lint-test:
+    uses: ./.github/workflows/lint-test.yaml
+
+
+  generate-sha-tag:
+    if: github.ref == 'refs/heads/main'
```
ChrisLovering:
```suggestion
```
This needs to be removed so that `publish-static-preview` can use it

.github/workflows/main.yaml
```diff
@@ -0,0 +1,54 @@
...
+          
+  publish-static-preview:
+    uses: ./.github/workflows/static-preview.yaml
+    with:
+      sha-tag: ${{ needs.generate-sha-tag.outputs.sha-tag }}
+
+
+  build-deploy:
+    if: github.ref == 'refs/heads/main'
+    uses: ./.github/worfklows/build-deploy.yaml
```
ChrisLovering:
```suggestion
    uses: ./.github/workflows/build-deploy.yaml
```

.github/workflows/main.yaml
```diff
@@ -0,0 +1,54 @@
...
+    outputs:
+      sha-tag: ${{ steps.sha-tag.outputs.sha-tag }}
+    steps:
+      - name: Create SHA Container tag
+        id: sha-tag
+        run: |
+          tag=$(cut -c 1-7 <<< $GITHUB_SHA)
+          echo "sha-tag=$tag" >> $GITHUB_OUTPUT
+          
+  publish-static-preview:
```
ChrisLovering, line 28: I think we only want to run this in the below case
```
    if: github.ref != 'refs/heads/main'
```
Doing this also means we can remove the checks inside `static-preview.yaml` that do this same check.

ChrisLovering (changes requested):

.github/workflows/main.yaml
shtlrs: True true, since we have the 'github.ref' != '..main' case 
99ca1f946a9bb88ecd54f04b8ddfae96fa0a2fdc

.github/workflows/main.yaml
shtlrs: I need to work on my typing issues 493a9486f917fc0efb6badc7caf4c803696764d3

jchristgit (approved):
Good. Thanks!

.github/workflows/main.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Create SHA Container tag
+        id: sha-tag
+        run: |
+          tag=$(cut -c 1-7 <<< $GITHUB_SHA)
+          echo "sha-tag=$tag" >> $GITHUB_OUTPUT
+
+  publish-static-preview:
+    uses: ./.github/workflows/static-preview.yaml
+    with:
+      sha-tag: ${{ needs.generate-sha-tag.outputs.sha-tag }}
```
ChrisLovering, line 33: You will need to mark this job as needing the `generate-sha-tag` job to access `needs` here

.github/workflows/main.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+  build-deploy:
+    if: github.ref == 'refs/heads/main'
+    uses: ./.github/workflows/build-deploy.yaml
+    needs:
+      - lint-test
+      - generate-sha-tag
+    with:
+      sha-tag: ${{ needs.generate-sha-tag.outputs.sha-tag }}
+      secrets: inherit
```
ChrisLovering: This is indented 1 level too much

.github/workflows/main.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+
+  sentry-release:
+    if: github.ref == 'refs/heads/main'
+    uses: ./.github/workflows/sentry_release.yml
```
ChrisLovering:
```suggestion
    uses: ./.github/workflows/sentry_release.yaml
```
Not needed for merging this, but we should consistently use either yml or yaml to avoid gotchas in the future.

(You'll need to update the file name too)

ChrisLovering (changes requested):

.github/workflows/main.yaml
shtlrs: 43e83dcfd02ae129939fc4a14b2de0d5fe2f75d6

.github/workflows/main.yaml
shtlrs: d4e2325d822fc21d0fc7548d01596364792597d9

ChrisLovering (dismissed):

jchristgit:
Can we do anything about the failing Netlify deploys - is it because of the changes here?

ChrisLovering (changes requested):
CI is failing because `lint-test.yaml` isn't updated to be triggered on workflow_call, so the static previews aren't being built

.github/workflows/main.yaml
```diff
@@ -0,0 +1,50 @@
...
+    needs:
+      - lint-test
+      - generate-sha-tag
+    with:
+      sha-tag: ${{ needs.generate-sha-tag.outputs.sha-tag }}
+    secrets: inherit
+
+  sentry-release:
+    if: github.ref == 'refs/heads/main'
+    uses: ./.github/workflows/sentry_release.yaml
```
ChrisLovering:
```suggestion
    uses: ./.github/workflows/sentry-release.yaml
```
File name is hyphen separated. We should likely make all of these consistent, either hyphen or underscores.

ChrisLovering (changes requested):

.github/workflows/build-deploy.yaml
```diff
@@ -56,3 +50,39 @@ jobs:
             ghcr.io/python-discord/site:${{ steps.sha_tag.outputs.tag }}
```
MarkKoz: The step was removed, so this seems like it will no longer work. Did you mean `inputs.sha_tag`? Same thing on line 82, and perhaps elsewhere that I missed.

MarkKoz (changes requested):

.github/workflows/build-deploy.yaml
shtlrs: Absolutely right, done in 36c9988028accc5eefaa61a630ccbb355fdd0067

ChrisLovering (approved):

PR #914, opened 2023-03-22
Drop dependency to pyfakefs

### LINKED ISSUE #679: Switch out pyfakefs to using test files in the test directory
This would also allow running `docker-compose exec web python manage.py test`, since `pyfakefs` is only installed for development.

Alternatively, we could add a separate container like `docker-compose run tests`.

"The best code is not having any code" - @lemonsaurus

jchristgit:
Create a temporary directory to manage our resource tests instead of reyling on pyfakefs to mock it away for us. This also makes the code more portable: all we need now is a way to create a temporary directory. `pathlib` mostly abstracts away the other parts for us. Since we're well-behaved, we clean up the temporary directory at the end of the Python interpreter's life using `atexit` and `shutil.rmtree`.

This PR was written and tested with Python 3.9 which required some hacks in `pyproject.toml` to make it work, it may require re-locking if CI throws up.

Closes #679.


kosayoda (approved):
Thanks!

pydis_site/apps/content/tests/helpers.py
```diff
@@ -75,29 +75,32 @@ class MockPagesTestCase(fake_filesystem_unittest.TestCase):
...
-        )
-        self.fs.create_file(f"{BASE_PATH}/category/subcategory/_info.yml", contents=CATEGORY_INFO)
-        self.fs.create_file(
-            f"{BASE_PATH}/category/subcategory/with_metadata.md", contents=MARKDOWN_WITH_METADATA
-        )
-        self.fs.create_file(
-            f"{BASE_PATH}/category/subcategory/without_metadata.md",
-            contents=MARKDOWN_WITHOUT_METADATA
-        )
+        Path(f"{BASE_PATH}/_info.yml").write_text(CATEGORY_INFO)
```
wookie184, line 79:
```suggestion
        (BASE_PATH / "_info.yml").write_text(CATEGORY_INFO)
```
potentially neater to let pathlib handle putting these together, although not sure it actually makes a different so should be fine as is.

pydis_site/apps/content/tests/helpers.py
```diff
@@ -75,29 +75,32 @@ class MockPagesTestCase(fake_filesystem_unittest.TestCase):
...
+        Path(f"{temp}/_info.yml").write_text(CATEGORY_INFO)
+        Path(f"{temp}.md").write_text(MARKDOWN_WITH_METADATA)
+        Path(f"{temp}/category").mkdir(exist_ok=True)
+        Path(f"{temp}/category/_info.yml").write_text(CATEGORY_INFO)
+        Path(f"{temp}/category/subcategory_without_info").mkdir(exist_ok=True)
+
+    @classmethod
+    def setUpTestData(cls):
+        # May get called multiple times - ignore erorrs in that case.
+        atexit.register(shutil.rmtree, BASE_PATH, ignore_errors=True)
```
wookie184: Why not put this atexit register right after the directory is created. This would ensure it is run only once, and also run even if the tests weren't run for some reason (e.g. selectively running tests possibly?).

Alternatively it may make more sense to move the tempdir to the test setup function, so it is only created if/when the tests are run. (could also use a `tempfile.TemporaryDirectory` which has a `cleanup` method instead of needing to rmtree but should be the same so doesn't matter).

wookie184 (changes requested):

pydis_site/apps/content/tests/helpers.py
jchristgit: This is a good shout. I have implemented it in 88157999.

jchristgit: As for having the tempdir in the test setup function, I do think it's a good idea, but it would require a larger refactoring due to the places BASE_PATH is referenced.

wookie184 (approved):
LGTM

Issue #674, opened 2022-02-23
Create per-directory README for our Django apps


jchristgit:
In the `apps` directory:

> This is the folder directory that I think would benefit the most from some guidance or explanation. 
> 
> Notable things to point out:
> 
> - admin - ??
> - api - I guess this is where our api resides. Probably a good point to mention DRF and dive into some api resources
> - content - This is where site content goes? Probably good to mention that it’s simple markdown, no jinja or django knowledge required
> - events - ??
> - redirect - ?? probably don’t need to dive into this
> - resources - point out how to contribute resources and the requirements for it?
> - staff - ??

jchristgit:
The final PR for this was merged yesterday :partying_face:

PR #922, opened 2023-03-28
Add a readme for the home app

jchristgit:
This commit also moves the nested structures for models and views in the home app into a single module, as they were not split up as part of the subpackage, with the goal of making this a bit more overseeable.

Part of #674.


wookie184 (approved):
Changes seems sensible, thanks 👍 👍

ChrisLovering (approved):

jchristgit:
Awesome! Thanks! King!

PR #948, opened 2023-04-21
Make unittests independent of GitHub

### LINKED ISSUE #918: Unit tests depend on external resources
At least one of our unit tests in the content apps seems to make requests to GitHub while it runs. When running multiple times, for instance when retrying tests after issues, this will cause 403 errors to appear. This also affects CI pipelines: https://github.com/python-discord/site/actions/runs/4515801023/jobs/7953504322?pr=916
When ratelimited, two other tests, will also log some form of "Could not find GitHub repository metadata from response!".
We should ideally mock away these HTTP requests or find some other way to mitigate this error.

jchristgit:
This fixes a problem where running the unit tests successively a lot would result in 403 ratelimit exceeded errors being thrown due to the GitHub API being called by the app.

Fixes #918

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -11,6 +11,8 @@ omit =
     */apps.py
     */urls.py
     pydis_site/apps/api/models/bot/metricity.py
+    # GitHub API functions are mocked away
+    pydis_site/apps/content/utils.py
     pydis_site/wsgi.py
     pydis_site/settings.py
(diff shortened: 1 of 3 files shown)


ChrisLovering (approved):

wookie184 (approved):
Nice, thanks

PR #626, opened 2021-11-19
feat: 'Edit on GitHub' for content articles

hedyhli:
## Description

Adds a link which allows users to view the page source on github for content articles.

Supersedes #623 and #506 
Closes #459 

- Shows up for all pages, including ones with sub-articles such as the popular [contributing page](https://www.pythondiscord.com/pages/guides/pydis-guides/contributing/)
- Does not show up in "folders" (ones without an actual page, such as [guides](https://www.pythondiscord.com/pages/guides/))
- Pencil icon used, similar to microsoft docs
- It's a simple link rather than a button because IMO, it is less attention-seeking and takes less visual space. (Viewers will probably not be using this link that often.) Plus, button implies an action, but this PR only adds a link to the page source (and not the actual online editor page on github) hence an action is not taken, making a link more suitable

## Screenshots

**A page with sub-articles (aka category page?) [`/pages/guides/pydis-guides/contributing/`](https://www.pythondiscord.com/pages/guides/pydis-guides/contributing/)**
pydis-editgh-1

**A page without sub-articles: [`/pages/guides/python-guides/discordpy_help_command/`](https://www.pythondiscord.com/pages/guides/python-guides/discordpy_help_command/)**
pydis-editgh-2

## Implementation

- Using a configuration in settings.py for github url, so it's not hardcoded
- Using a template filter to get the url given the request path
- The link is added on the right of the breadcrumb at the very end. I didn't feel like it makes sense to put it before the sub-articles dropdown, if any.
- The link together with the sub-articles dropdown is in a right-justified flex container, so the pull-right on sub-articles dropdown is no longer needed... I think

Note, github does have an _edit_ link that links the user to an online editor. This was not used for serveral reasons:
1. Other docs such as VS Code doesn't do this
2. It creates a fork for the user (AFAIK) which the user may not want
3. The user can chose where to go next when linked to the source page only (and not the edit link) - download, view raw, or edit.

## Design

Feedback appreciated on
- whether the pencil icon is more suitable than the github logo (as in original mock implementation in the issue by lemon)
- link rather than a button
- placing of the link in terms of semantics: should it be on the same line as the title instead? However, this will take up space for the title line for small screens, making the title wrap
- whether it's a good idea (or conventional?) to put the github source url as a variable in settings.py

### Support for smaller screens

Makes a mess when the screen is small (not able to fit the breadcrumb and the Edit on GitHub link in one line), making the link segmented.

[...]

PR #949, opened 2023-04-21
Edit on GitHub feature for content

### LINKED ISSUE #459: Edit on GitHub for content articles.
At the top of an article in the `content` app, we should show an "Edit on GitHub" button.
The button when clicked should lead to the GitHub page for that file - something like https://github.com/python-discord/site/blob/main/docs/configuration.md

The button should have a nice logo, maybe an edit pencil or a github icon.
Here's one such mock implementation:
image

Improvements on the mock design can be to reduce the visual weight of the button, or finding a way to make it smaller yet accessible.

Anyone who wants to implement this can propose alternative implementations, feel free to do so in `#dev-contrib` or `#dev-dewikification`.

jchristgit:
Pulled from https://github.com/python-discord/site/pull/626 to allow maintainer edits

Hidden in mobile:

image

Closes #459.

### DIFF
--- a/pydis_site/apps/content/templatetags/page_src.py
+++ b/pydis_site/apps/content/templatetags/page_src.py
@@ -1,0 +1,22 @@
+from django import template
+
+
+register = template.Library()
+
+
+@register.filter()
+def page_src_url(request_path: str) -> str:
+    """
+    Return the corresponding GitHub source URL for the current content article.
+
+    request_path is the relative path of an article, as returned by `request.path` in templates.
+
+    For example: /pages/rules/ would return:
+    https://github.com/python-discord/site/tree/main/pydis_site/apps/content/resources/rules.md
+    """
+    src_url = request_path.replace(
+        "/pages/",
+        "https://github.com/python-discord/site/tree/main/pydis_site/apps/content/resources/",
+    )
+    src_url = src_url[:-1] + ".md"
+    return src_url
--- a/pydis_site/apps/content/templatetags/page_src.py
+++ b/pydis_site/apps/content/templatetags/page_src.py
@@ -12,6 +12,11 @@ def page_src_url(request_path: str) -> str:
 
     request_path is the relative path of an article, as returned by `request.path` in templates.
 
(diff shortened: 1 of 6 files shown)


ChrisLovering (approved):

wookie184:
Comparing https://deploy-preview-949--pydis-static.netlify.app/pages/guides/pydis-guides/contributing/sir-lancebot/ to https://www.pythondiscord.com/pages/guides/pydis-guides/contributing/sir-lancebot/ it seems like this causes the rest of the page to drop down a bit.

Before:
image

After:
image

Any ideas on why this is?

jchristgit:
@wookie184 yes I have seen that, but I have no idea what causes it. Inspecting the layout in Firefox I concluded that it's no margin or padding from the element itself. I also concluded that it's probably not noticeable by anybody and was hoping I could sweep it under the rug. Unfortunately, you have crossed through my plan to do it.

Any other idea where we can check what's causing the surplus whitespace apart from the layout box in devtools? I also tried to use a `span` instead of `div` to no success.

ChrisLovering:
@jchristgit See @hedyhli's messages here for some solutions
https://canary.discord.com/channels/267624335836053506/635950537262759947/1100602004188958740

jchristgit:
I saw these but I don't really have the CSS-fu to drive it to the finish line with the vertical stacking :(

hedyhli:
Hey there, so I've played around with the HTML a little and it seems this patch works:
https://paste.pythondiscord.com/raw/fexodipane

Basically comparing the elements from before this PR and after, the container appears to gained a height, and the breadcrumb `<section>` also changed because of the added flex div element. So I decided to take it out of the container, then modify the CSS classes a little.

Hopefully this diff isn't a 'hacky' fix though (otherwise it might pose problems in the future).

*Note: If you compare the positioning of sub-articles and edit-on-gh before and after applying the patch, you might notice the two having moved towards the right a little, it could possibly be fixed with something like a `margin-right`. IMO it's more trivial than the entire content dropping down as wookie pointed out.*

wookie184 (approved):
Looks good, thanks @hedyhli for the help with this!

PR #968, opened 2023-05-10
Switch to ruff for linting

jchristgit:


.pre-commit-config.yaml
```diff
@@ -16,9 +16,9 @@ repos:
       - id: python-check-blanket-noqa
```
ChrisLovering: Cant comment higher, but pygrep hook can be removed from pre-commit as ruff covers it already (rule PGH)

ChrisLovering (dismissed):
Looks good, just one optional thing that might be good to do now too.

.pre-commit-config.yaml
jchristgit: Thanks! I dropped it in the amended commit. The tests have also been fixed.

ChrisLovering (approved):

jchristgit:
MASCHINNNNNNNNNNNNNNNN

PR #973, opened 2023-05-13
Python 3.11

ChrisLovering:
This bumps the site up to using Python 3.11. This includes the poetry project file, within the Docker image, in CI, and the target version for ruff.

This also fixes ruff linting errors by using the new datetime.UTC alias present in 3.11.

While running the test suite locally I noticed a test that failed on a Windows host, so updated that too.


jchristgit (approved):
Looks great, looking forward to running tests again on Debian Bookworm :muscle:

Xithrius (approved):

PR #1006, opened 2023-06-25
Remove obsolete question from FAQ

n0Oo0Oo0b:
Since the migration of the help system to a forum channel, the question "Why are the help channels named after elements/food?" is obsolete now. This PR removes it from the FAQ page.


mbaruh (approved):
Thanks!

PR #1007, opened 2023-06-25
Add new rule aliases

### LINKED ISSUE #1001: Add some more common abbreviations to the rule keyword command list
I believe it would be beneficial to add two more shorthands for invoking the 'Rules' embed: "hw" and "eng". "hw" is a common shorthand for "homework", so it should be associated with the embed for rule 8. Likewise, "eng" is a common abbreviation for "English", so it can be linked with rule 4.

jchristgit:
See suggestion in #1001.
Closes #1001.


shtlrs (approved):
Merci, omelette au fromage.

Issue #1036, opened 2023-07-16
Inconsistent Column Heights in the Contributing Section


vvzuev:
Summary: In the "Contributing" section of the  "Contributing" page, the column, which contains a link to the GitHub repository "Site", is shorter in height than the other two columns. This results in an inconsistent visual layout.

URL: https://www.pythondiscord.com/pages/guides/pydis-guides/contributing/

Environment: windows 10, display resolution: 1024x768, browsers: Chrome, Firefox

Step to reproduse: 
1.  Navigate to the "Contributing" page.
2. Scroll down to the section with the three columns linking to the GitHub repositories.
3. Observe the height of the column, which contains a link to the GitHub repository "Site", compared to the other two.

Expected Result: All columns should have the same height to maintain a consistent and visually pleasing layout.

Actual Result: The third column is shorter in height than the other two columns, resulting in an inconsistent layout.

Attachment:

![Contributing1](https://github.com/python-discord/site/assets/51853424/c560bb26-f125-452a-911b-934830beccb6)

korstep:
I can't find the file where the styles for cart and all child classes are located in the browser says that the styles are from bulma but I don't see it at all.

jchristgit:
closing via #1040

PR #1039, opened 2023-07-17
Fix Inconsistent Column Heights in the Contributing Section

korstep:
issue #1036


korstep:
forgot that .github-card in card. I'll fix it.

PR #1040, opened 2023-07-17
Fixed inconsistent column heights in the contributing section

korstep:
issue - #1036


ChrisLovering (approved):
Looking through the [bulma docs](https://bulma.io/documentation/columns/) there doesn't seem to be a native way to do this.

[This issue](https://github.com/jgthms/bulma/issues/1977) suggests doing similar to what you've done here.

I agree with this approach.

jchristgit (approved):

PR #1072, opened 2023-08-12
Block fixup commits

shtlrs:
This uses the organisation's fixup commit blocker workflow.

The purpose is to make sure all commits prefixed with `fixup` need to be squashed before merging.


MarkKoz (approved):

jchristgit (approved):
still great

PR #1108, opened 2023-09-18
Update deployment CI to pull manifest from infra repo

ChrisLovering:


.github/workflows/build-deploy.yaml
```diff
@@ -78,7 +76,7 @@ jobs:
         uses: Azure/k8s-deploy@v4
         with:
           manifests: |
-            namespaces/default/site/deployment.yaml
+            infra/kubernetes/namespaces/default/site/deployment.yaml
```
shtlrs, line 79: Shouldn't this be kubernetes/namespaces/... e.g without the infra/ as top level directory since when you checkout you're cd'd into the repo dir?

ChrisLovering: I added the `path` arg to the checkout action, so it checks out thee repo into the `infra` directory

shtlrs: Yeah, I've missed that..

shtlrs (approved):

Xithrius (approved):

PR #1111, opened 2023-09-19
Fix Dead CJ Banner Link

HassanAbouelela:
Possibly we'd want to use banner2 instead?


jchristgit (approved):

kosayoda (approved):
We can switch it to `banner2` in the future if needed, no harm going with this fix now.

PR #1165, opened 2023-12-10
Implement editing of offensive message records

### LINKED ISSUE #364: Support rescheduling of offensive messages
When a message trips the filter on the bot it is removed after a period of time. During this period it is a record in the database.

When this deletion date is reached the bot will attempt to remove the message from Discord and remove the record from the offensive message table. We currently handle for the message being not found (deleted) but if another error occurs resulting in the message not being deleted we still continue to delete the database record, causing the message to be left around.

We should allow the bot to perform a PATCH request to the deleted message endpoint to update the delete time and reschedule if something has failed (for example, a hiccup on Discord's end).

However, we must also bear in mind that permanent rescheduling could potentially leave lingering records in our database if a case is discovered where a message which cannot be deleted is rescheduled repetitively.

jchristgit:
Allow changing the deletion date of offensive message records in case the bot encounters an error during deletion attempts.

Fixes #364.
Unblocks python-discord/bot#1013.

### DIFF
--- a/pydis_site/apps/api/viewsets/bot/offensive_message.py
+++ b/pydis_site/apps/api/viewsets/bot/offensive_message.py
@@ -1,6 +1,7 @@
 from rest_framework.mixins import (
     CreateModelMixin,
     DestroyModelMixin,
+    UpdateModelMixin,
     ListModelMixin
 )
 from rest_framework.viewsets import GenericViewSet
@@ -10,7 +11,7 @@ from pydis_site.apps.api.serializers import OffensiveMessageSerializer
 
 
 class OffensiveMessageViewSet(
-    CreateModelMixin, ListModelMixin, DestroyModelMixin, GenericViewSet
+    CreateModelMixin, ListModelMixin, UpdateModelMixin, DestroyModelMixin, GenericViewSet
 ):
     """
     View providing CRUD access to offensive messages.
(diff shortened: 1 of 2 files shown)


pydis_site/apps/api/viewsets/bot/offensive_message.py
```diff
@@ -46,6 +47,16 @@ class OffensiveMessageViewSet(
     - 201: returned on success
     - 400: if the body format is invalid
 
+    ### PATCH /bot/offensive-messages/<id:int>
```
shtlrs, line 50: I just have one question, do we want to allow patching of all fields, or just the deleting date ?

jchristgit: This (currently) allows patching all fields. I can make the other fields write-once, but that will be a bit more work.

shtlrs: Sure, I was just wondering what we wanted to have. If we only want to allow date modifications, then we might block the others as well.

ChrisLovering:
I'm reading the issue that suggested this approach to fixing, however I'm not sure it's the best way.

Surely if we fail to delete the message due to a Discord issue, we just shouldn't delete it from the DB. When the bot next checks for messages to delete, the message that failed will still be there, with a date in the past, so it will try to delete again.

jchristgit:
> Surely if we fail to delete the message due to a Discord issue, we just shouldn't delete it from the DB. When the bot next checks for messages to delete, the message that failed will still be there, with a date in the past, so it will try to delete again.

I think the idea is more to have a longer, adequate retry period. As far as I remember from when we implemented this, this cleanup is more intended as a "later down the line" batch job rather than "this needs to be removed instantly", so a greater time delay before retrying is the idea here.

ChrisLovering (approved):

PR #1166, opened 2023-12-10
Add endpoint to fetch filters in JSON format

### LINKED ISSUE #710: Backend information to fetch smarter resources filters
## Summary

Following python-discord/bot#2079 it was discussed having an API allowing the bot to fetch possible filters so that implementations don't need to scrape it.

This new API could power different future bot features like having interaction-based URL builder.

## Description

Currently resources are loaded from a folder of yml, [then a number of different steps in processing are applied](https://github.com/python-discord/site/blob/main/pydis_site/apps/resources/views/resources.py#L33-L103) to sort out all filters and its topics. This is a highly dynamic system which will complicate this PR.

The API resides in [`pydis_site/apps/api`](https://github.com/python-discord/site/tree/main/pydis_site/apps/api) whilst the resources is in its own app at [`pydis_site/apps/resources`](https://github.com/python-discord/site/tree/main/pydis_site/apps/resources) (the actual view is in [`pydis_site/apps/resources/views/resources.py`](https://github.com/python-discord/site/blob/main/pydis_site/apps/resources/views/resources.py)) which complicates things. See, for the API backend to get access to that information there needs to be an import there or loading - which might be unwanted.

This means that the implementation might be a bit more complicated than one might first think. The easiest way to go about this would be placing the route under `resources` but it is still, technically, part of the API. Another option would be duplicating the processing, inside of API, but that still requires access to the folder of resources meaning that we are back at square one of having unwanted coupling.

## Assignee

@iamericfletcher has asked to implement this *(I couldn't assign them directly)*. Please note that this **is still in planning**.

jchristgit:
While this is an API endpoint consumed by the bot, keep it in the
`resources` app instead of the `api` app, as all the logic and data for
resources is contained within the `resources` app and we don't want to
start messing around with that.

The response format from the endpoint is as follows:

    {
        "topics": [
            "algorithms-and-data-structures",
            "data-science",
            "databases",
            "discord-bots",
            "game-development",
            "general",
            "microcontrollers",
            "security",
            "software-design",
            "testing",
            "tooling",
            "user-interface",
            "web-development",
            "other"
        ],
        "payment_tiers": [
            "free",
            "paid",
            "subscription"
        ],
        "type": [
            "book",
            "community",
            "course",
            "interactive",
            "podcast",
            "project-ideas",
            "tool",
            "tutorial",
            "video"
        ],
        "difficulty": [
            "beginner",
            "intermediate"
        ]
    }

Closes #710.

### DIFF
--- a/pydis_site/apps/resources/tests/test_views.py
+++ b/pydis_site/apps/resources/tests/test_views.py
@@ -27,3 +27,14 @@ class TestResourcesView(TestCase):
         url = reverse("resources:index", kwargs={"resource_type": "urinal-cake"})
         response = self.client.get(url)
         self.assertEqual(response.status_code, 404)
+
+
+class TestResourceFilterView(TestCase):
+    def test_resource_filter_response(self):
+        """Check that the filter endpoint returns JSON-formatted filters."""
+        url = reverse('resources:filters')
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        content = response.json()
+        self.assertIn('difficulty', content)
+        self.assertIsInstance(content['difficulty'], list)
--- a/pydis_site/apps/resources/urls.py
+++ b/pydis_site/apps/resources/urls.py
@@ -1,11 +1,12 @@
 from django_distill import distill_path
 
-from pydis_site.apps.resources.views import ResourceView
+from pydis_site.apps.resources.views import ResourceView, ResourceFilterView
 
 app_name = "resources"
 urlpatterns = [
     # Using `distill_path` instead of `path` allows this to be available
     # in static preview builds.
     distill_path("", ResourceView.as_view(), name="index"),
+    distill_path("filters", ResourceFilterView.as_view(), name="filters"),
     distill_path("<resource_type>/", ResourceView.as_view(), name="index"),
 ]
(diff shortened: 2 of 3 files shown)


ChrisLovering (approved):

wookie184 (changes requested):
The response format here doesn't seem ideal. My understanding is that "filters" is the display name, and "valid_filters" is what goes in the URL(?), but the response doesn't link these together.

We could link them together in the structure, along the lines of this:
```
[
    {
      "name": "Difficulty",
      "query_param": "difficulty",
      "options": [
        {"name": "Beginner", "query_value": "beginner"},
        ...
      ],
      "icon": "fas fa-brain",
      "hidden": false,
    },
    ...
]
```

Alternatively we could maybe just drop the "valid_filters" section and document the transformation to get from the name to the url.

jchristgit:
@wookie184 thanks for the review and sorry for the long delay. I've updated the response format, the "sluggified" / "kebabified" version of the names can now be found in the `slug` field for the respective category, and the `filter_slugs` field for each filter. To be specific, `filter_slugs[3] == to_kebabcase(slugs[3])`. Let me know what you think.

wookie184 (changes requested):
This seems to undo the fix for https://github.com/python-discord/site/issues/1209, I think it will cause performance issues.

jchristgit:
Looks like I managed to mess up during rebasing. It should all be fine now, and the code should be a lot simpler thanks to Rohan's changes. Please also see the new response file format.

jchristgit:
Would be happy for a re-review here, in loving memory of our brother Chris.

wookie184 (approved):
Not tested but LGTM

PR #1171, opened 2023-12-12
Add alternate accounts to the user model

jchristgit:
Introduce a way to store alternate accounts on the user, and add the `PATCH /bot/users/<id:str>/alts` endpoint, which allows updating the user's alt accounts to the alt accounts in the request..


ChrisLovering (approved):

pydis_site/apps/api/tests/test_users.py
```diff
@@ -199,13 +200,15 @@ def test_multiple_users_patch(self):
         url = reverse("api:bot:user-bulk-patch")
         data = [
             {
+                "alts": [],
```
ChrisLovering: Does this imply that setting this field is required when creating/updating? We probably don't want this, as it would mean we need to update all the places in bot that does this to also set it to `[]`.

We probably want it default to `[]`, and have a test to make sure it works if this field isn't passed in the body

ChrisLovering (changes requested):

pydis_site/apps/api/serializers.py
```diff
@@ -679,7 +679,9 @@ class Meta:
         """Metadata defined for the Django REST Framework."""
 
         model = User
-        fields = ('id', 'name', 'discriminator', 'roles', 'in_guild')
+        fields = ('id', 'name', 'discriminator', 'roles', 'in_guild', 'alts')
+        # This should be edited on a separate endpoint only
+        frozen_fields = ('alts',)
```
MarkKoz: Why are we making this design decision over allowing direct updates on the existing users endpoint?

ChrisLovering: As a client the bot would need to fetch the user, edit the list of alts, then send back the updated list to site.

Since that's not an atomic change, we could lose data if two actions were preformed on the same user at the same time.

This approach means that each action is preformed atomically by site

pydis_site/apps/api/tests/test_users.py
ChrisLovering: This isn't the case, I read wrong.

ChrisLovering (approved):

pydis_site/apps/api/models/bot/user.py
```diff
@@ -86,3 +92,34 @@ def username(self) -> str:
         For usability in read-only fields such as Django Admin.
         """
         return str(self)
+
+class UserAltRelationship(ModelReprMixin, models.Model):
```
ChrisLovering: I missed something from the original requirements when first reviewing this PR, only realised when starting to do the bot side.

We want an context text field here, along with an actor, date created and date edited.

`context` is required when marking two users as alts. This is so that mods can give proof as to _why_ they this the two users are alts.
`actor` is the api_user who created the relationship
`date_created` is insertion date
`date_edited` is when the reason what changed

This does also add a bit more complexity...as we likely want a way to edit this reason too. So both the ability to edit relationships & editing relations can be skipped for this PR, and added in the future if needed. Since if you really want to edit, you can delete and re-add

I'll start on the bot side impl of this anyway, since it's mostly going to be the same from the bot's side

ChrisLovering (changes requested):

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -284,6 +317,48 @@ def bulk_patch(self, request: Request) -> Response:
 
         return Response(serializer.data, status=status.HTTP_200_OK)
 
+    @action(detail=True, methods=['POST'], name="Add alternate account",
+            url_name='alts', url_path='alts')
+    def add_alt(self, request: Request, pk: str) -> Response:
```
ChrisLovering, line 372: This should also error if the two users are already marked as alts

pydis_site/apps/api/viewsets/bot/user.py
```diff
@@ -284,6 +317,48 @@ def bulk_patch(self, request: Request) -> Response:
...
+            raise ParseError(detail={"non_field_errors": ["Request body should be an integer"]})
+        try:
+            alt = user.alts.get(id=request.data)
+        except ObjectDoesNotExist:
+            raise ParseError(detail={
+                'non_field_errors': ["Specified account is not a known alternate account of this user"]
+            })
+
+        user.alts.remove(alt)
+        return Response('', status=status.HTTP_204_NO_CONTENT)
```
ChrisLovering:
```suggestion
        return Response(status=status.HTTP_204_NO_CONTENT)
```
The empty string here seemed to make aiohttp sad when consuming this response. Removing this seemed to solve it.

ChrisLovering (changes requested):

pydis_site/apps/api/serializers.py
MarkKoz: Can we not support a partial update? We only need to update the alts field, not the entire user.

jchristgit: Strictly speaking this line doesn't do anything at all, because from my
testing DRF already doesn't allow editing the field over the regular
patch endpoint - it just ignores whatever you give it. The
`frozen_fields` therefore serve more of a documentary purpose than
actually changing anything. I haven't found a way to enable this in
DRF...

pydis_site/apps/api/models/bot/user.py
jchristgit: I've implemented this, unfortunately we lose the symmetrical aspect of the `ManyToManyField` then:

https://docs.djangoproject.com/en/5.0/topics/db/models/#intermediary-manytomany

> For a model which has a many-to-many relationship to itself through an intermediary model, two foreign keys to the same model are permitted, **but they will be treated as the two (different) sides of the many-to-many relationship**. If there are more than two foreign keys though, you must also specify through_fields as above, or a validation error will be raised.

I've changed this such that our code (manually) adds records in both directions in a transaction, which should keep our database consistent. The only downside is that we are sort-of duplicating data. I am open to suggestions on other solutions for this.

Deletes are symmetrical automatically, I added tests to verify this works as expected for both additions and deletions in both ways.

pydis_site/apps/api/viewsets/bot/user.py
jchristgit: Added (see `test_adding_existing_alt_twice_via_{source,target}` test cases)

pydis_site/apps/api/viewsets/bot/user.py
jchristgit: Strange. I removed it.

pydis_site/apps/api/models/bot/user.py
jchristgit: Oh, also note that the field names are not what you asked exactly because I've reused our existing model timestamp mixin. So you want to look at `created_at` and `updated_at` instead.

pydis_site/apps/api/models/bot/user.py
```diff
@@ -86,3 +93,45 @@ def username(self) -> str:
...
+        help_text="The target of this user to alternate account relationship",
+    )
+    context = models.TextField(
+        help_text="The reason for why this account was associated as an alt.",
+        max_length=1900
+    )
+    actor = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        related_name='alts_marked',
```
jchristgit: Do we need this? I originally added it, but re-reading I think that some form of "look at all the alts I found!" may not be the most interesting feature.

pydis_site/apps/api/models/bot/user.py
jchristgit: > So both the ability to edit relationships **& editing relations** can be skipped for this PR

Just to make sure I understand this properly, don't we only want to be able to change the `context` of a relationship, nothing else?

ChrisLovering: Yea, that correct, clients should only be able to update the context, doing so will also update the updated_at field.

pydis_site/apps/api/models/bot/user.py
ChrisLovering: What does this do? Allow us to find all the alts that have been assigned by a single user?

Doesn't seem like we need it right now, but could be useful to have

jchristgit: Removed it

ChrisLovering (approved):

pydis_site/apps/api/serializers.py
```diff
@@ -683,6 +715,26 @@ def create(self, validated_data: dict) -> User:
...
+class UserWithAltsSerializer(FrozenFieldsMixin, UserSerializer):
+    """A class providing (de-)serialization of `User` instances, expanding their alternate accounts."""
+
+    alts = SerializerMethodField()
+
+    class Meta:
+        """Metadata defined for the Django REST Framework."""
+
+        model = User
+        fields = ('id', 'name', 'discriminator', 'roles', 'in_guild', 'alts')
```
jb3: Do we not need to include the display name field here?

jb3 (changes requested):
All looks good, only thing I've noticed is we store user display name but it looks missing from a lot of this.

It looks like the missing serialisation here was already present but maybe it should be added in now?

jb3:
Oh just noticed the PR age, the serialisation _might_ be present in main then.

pydis_site/apps/api/serializers.py
jchristgit: Yes, we do. This wasn't a thing when I opened the PR. I included it now.

jchristgit:
Migrations are merged, `display_name` was added, please re-review.

jb3 (approved):

jb3 (changes requested):
I may misunderstand Django migrations and this may be a non-issue, but do we need to update the dependencies of the 0093_user_alts migration to instead depend on 0095_user_display_name (which should be the most recent migration)

ChrisLovering (approved):

jchristgit:
@jb3 
> I may misunderstand Django migrations and this may be a non-issue, but do we need to update the dependencies of the 0093_user_alts migration to instead depend on 0095_user_display_name (which should be the most recent migration)

We do not. The merged migration takes care of resolving conflicts.

jb3 (approved):

PR #1172, opened 2023-12-12
Enable tracing and include source for slow queries

ChrisLovering:


shtlrs (approved):

jchristgit (approved):
For the future: You probably want `str.casefold`

PR #1174, opened 2023-12-13
Migrate mailing lists to their own API endpoints

jchristgit:
Add a new model for the bot to store its mailing list state in, as opposed to the current JSON blob in the BotSetting table. Migrate the existing settings from the BotSetting table into the new model.


pydis_site/apps/api/migrations/0096_mailinglistseenitem_unique_list_and_hash.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 5.0 on 2023-12-13 08:00
+
+from django.db import migrations, models
+
+
```
shtlrs: Wouldn't it be better to squash the schema migrations into one ?

pydis_site/apps/api/viewsets/bot/mailing_list.py
```diff
@@ -0,0 +1,85 @@
...
+
+    ### GET /bot/mailing-lists
+    Returns all the mailing lists and their seen items.
+
+    #### Response format
+    >>> [
+    ...     {
+    ...         'id': 1,
+    ...         'name': 'python-dev',
+    ...         'seen_items': [
```
shtlrs, line 28: Is there a reason why you're returning a list of objects instead of a list of the hashes directly ?

pydis_site/apps/api/viewsets/bot/mailing_list.py
```diff
@@ -0,0 +1,85 @@
...
+
+    lookup_field = 'name'
+    serializer_class = MailingListSerializer
+    queryset = MailingList.objects.prefetch_related('seen_items')
+
+    @action(detail=True, methods=["POST"],
+            name="Add a seen item for a mailing list", url_name='seen-items', url_path='seen-items')
+    def add_seen_item(self, request: Request, name: str) -> Response:
+        """Add a single seen item to the given mailing list."""
+        list_ = self.get_object()
```
shtlrs: We can void the unnecessary query by doing the `isinstance` check before fetching the object

shtlrs (changes requested):
Great changes ! 
This is just a code review for the timebeing, will test in a later instance.

pydis_site/apps/api/migrations/0096_mailinglistseenitem_unique_list_and_hash.py
jchristgit: I've squashed the final two, however for the previous migrations I've intentionally opted to keep them separate because one is an elidable data migration. I suggest once this and the alternate account PR is merged we make a "big squash"

pydis_site/apps/api/viewsets/bot/mailing_list.py
jchristgit: Yes. I was lazy :smiley:

I've changed and pushed this.

pydis_site/apps/api/viewsets/bot/mailing_list.py
jchristgit: Good idea, changed.

pydis_site/apps/api/migrations/0096_mailinglistseenitem_unique_list_and_hash.py
jchristgit: Hello. My lawyer has advised me to squash the migrations as you requested. This is now live.

shtlrs: Thanks !
Great lawyer btw, you should give him a raise.

pydis_site/apps/api/migrations/0094_migrate_mailing_listdata.py
```diff
@@ -0,0 +1,45 @@
...
+        (mailing_list, _created) = MailingList.objects.get_or_create(name=list_name)
+        MailingListSeenItem.objects.bulk_create(
+            MailingListSeenItem(list=mailing_list, hash=item_hash)
+            for item_hash in item_hashes
+        )
+
+    setting.delete()
+
+
+def do_nothing(*args, **kwargs):
```
shtlrs: The `reverse_code` param defaults to None, so no need to have the `do_nothing`  method def.

shtlrs (approved):
Works great, thanks !
I've marked this as do not merge untiil the bot pr is ready.

pydis_site/apps/api/migrations/0094_migrate_mailing_listdata.py
jchristgit: Oh, I actually left this in by accident. I had to add this to allow me to roll back the migration properly (if you don't specify the reverse, it will error out as being irreversible). I've removed it now for The Big Squash 2023

pydis_site/apps/api/viewsets/bot/mailing_list.py
```diff
@@ -0,0 +1,85 @@
...
+from rest_framework.mixins import ListModelMixin, RetrieveModelMixin
+from rest_framework.request import Request
+from rest_framework.response import Response
+from rest_framework.viewsets import GenericViewSet
+
+from pydis_site.apps.api.serializers import MailingListSerializer
+from pydis_site.apps.api.models import MailingList, MailingListSeenItem
+
+
+class MailingListViewSet(GenericViewSet, ListModelMixin, RetrieveModelMixin):
```
shtlrs: As I was making the bot side, I realised that the fact that we can't create mailing lists is an issue for the following reasons:
1. We don't have any migrations that create theses mailing lists, so if you delete a container that used to have th old table structure, you'll never end up with the target mailing lists.
2. We used to workaround this bot side thanks to the `sync_mailing_lists` method that creates the proper JSON that will be dumped in the db, this is no longer true
3. The bot will keep on getting `404` as it has as hardcoded mailing list that it tracks, so when it tries to post to site, it will complain saying it can't find such registered list.


We should use `ListCreateModelMixin` instead to allow the bot to create the mailing lists.

shtlrs (changes requested):

pydis_site/apps/api/viewsets/bot/mailing_list.py
jchristgit: Do you mean `CreateModelMixin`? I can't find anything about `ListCreateModelMixin`.

shtlrs: I always get confused, I was talking about the ApiView.

Yes you're right, it's the CreateModelMixin

jchristgit: Pushed this.

shtlrs (approved):
On The House

Xithrius:
Are we good to go on removing the do not merge label?

Xithrius (approved):

Issue #662, opened 2022-02-16
Improve logos on the landing page.


lemonsaurus:
Right now, there are three problems with the logos displayed in our sponsor hero and in the footer of the site.

## An alignment issue with the logos on Windows.
This problem is described in #458. We should optimize the `background-position` so that these logos will be optimally aligned on Windows machines, since our typical visitor is a Windows user.

## Hacky filtering to turn logos into white logos
Right now, we're using `filter` in CSS to turn the Django and Linode logos into something that will fit on our footer. This causes a weird display problem on mobile (seen in #458), and also poses a problem for implementing the dark theme. Let's just add staticfiles for these logos in the right size, and we will reduce the load time and make it much easier to build a dark theme around.

## Sponsor logos are arbitrarily sized
Why is the CloudFlare logo absolutely _enormous_, but other logos (like netCup) are _tiny_? It certainly doesn't represent the scope of the partnerships we have with these companies. Let's try to make these sizes more even, or choose different logos that can better fit into a smaller space for some partners.

I'll be tackling this personally.

hedyhli:
> (Re: Hacky filtering on footer logos)
> [...] and also poses a problem for implementing the dark theme

The alignment issue still exists, and it would be good if the linode logo green color is preserved, however there is no issue in dark theme caused by this 👍.

> Sponsor logos are arbitrarily sized

This issue would be fixed in my dark theme PR (#1176).

hedyhli:
A little update on this issue since the dark theme has been merged.

> Hacky filtering to turn logos into white logos

We now have dark versions of logos for django and linode in our static assets made by lemon.

> Sponsor logos are arbitrarily sized

This issue is now fixed IMO. All sponsor logos in the first two rows take up the same height and 1/3 of the total width. Netlify and Cloudflare use the same width as others but share the last row with 1/2 of the total width.

On mobile, I've made all logos fit the (slimmer) horizontal width with a dynamic height.

PR #1176, opened 2023-12-14
Dark theme

### LINKED ISSUE #362: Implement a dark theme
There's been some requests for a dark theme, and we came up with a feasible way this could be implemented.

First, let's solve the actual theming on `django-simple-bulma` instead of here. I've created an issue for this, and until this issue is solved, this issue will be blocked.
https://github.com/python-discord/django-simple-bulma/issues/46

Second, we just create an alternative theme by passing in a dict of alternate variables into `BULMA_SETTINGS`, and then ensuring we have the right bulma classes applied to the right parts of the site. We probably need to apply a class to the background of the page itself, everything is very white currently.

Third, we need some way for our users to toggle this on and off. ~~We could put a switch in the user settings, so that users would have to log in with OAuth and set their preference.~~ We could put a simple on off switch somewhere on the page and use some sort of cookie or something to store it for that session.

hedyhli:
<details>
<summary> Old PR description </summary>
I have pulled in @GDWR's work and new changes from main.

I have also went ahead and fixed all the conflicts.

This is intended to supplement the original PR #657, in case anyone working on #362 wishes to base the work on top of this one.

~~Unfortunately I did the pull from main in 2 stages, which created two commits of the same message. I couldn't find a way to rectify this without having to fix the conflicts all over again, which is rather laborious.~~

**NOTE**
As of writing, no significant changes were made to GDWR's original implementation. Only trivial ones which were necessary for fixing conflicts.

**PR Status**
For anyone looking to work on top of it, these things have already been done:
- Live theme toggle
- Set of colors in `settings.py` for dark theme
- Most of the colors throughout the site

The rest of the work, IMO, would involve mostly trivial color fixes and testing for the toggle button.
However, some decisions are to be made about the colors, such as the bright blurple, the (hardcoded-color-heavy) timeline page, and the footers.

</details>

---

This PR attempts to revive the work of previous contributors adding a dark theme for the pydis site. Hopefully, if this does not become another stale PR like the unfortunate ones before it, it should close #362.

## Note to reviewers

 Most of the files in the diff are just about adding bulma color classes so they can be displayed correctly in dark mode, you are encouraged to make use of the `Viewed` checkbox. The bulk of the work in this PR is in these files:

1. `pydis_site/settings.py`
1. `pydis_site/static/js/base/themes.js`
1. `pydis_site/templates/base/navbar.html`
1. `pydis_site/static/css/base/themes.css`
1. `pydis_site/static/css/home/index.css`
1. `pydis_site/static/css/home/index.html`

See, not so many files at all :)

I've made a number of notes throughout the diff to hopefully accompany reading of those changes, in places where changes might be questionable.

### JavaScript

Handles:
- Switch toggle click - apply theme and save preference
- Applying correct theme on page load

excalidraw export

[View on excalidraw](https://excalidraw.com/#json=qjE0H2NhmbpHBsosmPwuA,0mTLQehFoPxhOrYm4T5V3Q)

### Themes

- The task of including the actual dark theme is delegated to django-simple-bulma, using a set of colors specific for dark mode, defined in `settings.py`.
- Throughout the site, bulma's classes to control colors are applied so that they can switch accordingly to the site's theme; these are the bulk of this PR's file diffs.
- django-simple-bulma template tag is used and both the default and the dark theme stylesheet is included on every page.
- When applying themes, ~~the stylesheet that is unused for that theme is set to have attribute disabled, so the browser uses only one stylesheet~~ (this old method causes FOUC doing theme switches. Now I'm using only one stylesheet and switching its href attribute instead).

### Changes to GDWR's implementation

- Local storage is used rather than cookies to store the user preference for reasons described here 058b5623.
- Removed JQuery dependency on _every page that didn't used to use JQuery_ by rewriting some of the code that depended on JQuery.
- Some other refactors to the javascript.
- Applied bulma classes to some other parts of the sites to ensure themes display correctly.
- Added setting `data-theme="dark"` in the root html element to make CSS styling easier which targets specifically for dark mode.
- Different way to switch stylesheets which happens to eliminate FOUC

### Known limitations

- UI
  - Existing hardcoded colors throughout the site appears to be arbitrarily chosen to make it look best on the specific case, so I did the same where applicable for the dark colors (mostly drop shadows and border colors)
- UX
  - ~~There is a latency in applying the current theme on page load, because this operation is only done after the page is loaded.~~ (this is now fixed thanks to @wookie184)
  - ~~Sometimes, not always, there is a flash of unstyled content (FOUC) effect when switching themes.~~ This is now fixed with the tradeoff of slightly more brittle javascript code.
  -  FOUC when switching system preference while the site is live. See [this comment](https://github.com/python-discord/site/pull/1176#pullrequestreview-1834627048).
- Code
  - Bulma supports neither CSS variables (see [jgthms/bulma#1837], and [[this branch](https://github.com/jgthms/bulma/tree/bulma-v1-css-variables)]) nor any good theming API to make this easier (the `scheme*` variables, due to limitation of django-simple-bulma requiring all variables to be predefined as bulma SCSS aren't imported first, makes defining them in `settings.py` verbose, even so, using these `scheme` variables leads to inconsistency due to limited support for the full range of bulma color variables (only `black*` and `white*` colors are supported, no `grey*`)).

    What I ended up doing is copying bulma CSS attributes where applicable and modifying only the colors.

## Demo

<details><summary>Live system color scheme preference change</summary>

https://github.com/python-discord/site/assets/50042066/867c2ee7-9f49-43f6-88a5-dfeb60f6b104

</details>

## Thanks

This PR's work stands on the shoulder of giants. Notably, with thanks to:

- Lemon for the beautiful switch design and implementation ideas in the original issue,
- GDWR for initial implementation and bulk of the work on applying the bulma classes,
- Contributors to django-simple-bulma adding the theme feature.

## Todo

- [x] Fix homepage wave
- [x] Fix homepage hardcoded colors
- [x] Fix the hardcoded colors on the timeline page
- [x] Fix colors on contributing page
- [x] Fix colors on resources page
- [x] Look for hardcoded colors elsewhere
- [x] Possibly reconsider how the dark theme is applied so that it is the most minimal, performant, and maintainable (like using [localstorage](https://caniuse.com/?search=localstorage) [rather than cookies](https://web.dev/patterns/theming/theme-switch#js)?)
- [x] Respect `@media` query if user preference not already recorded

--- WIP ➡️  ready for review ---

- [x] Device & browser testing for the switch


jchristgit:
> Unfortunately I did the pull from main in 2 stages, which created two commits of the same message. I couldn't find a way to rectify this without having to fix the conflicts all over again, which is rather laborious.

You could try to squash the commits in an interactive rebase. Or use `git rerere` (reuse recorded resolution)

hedyhli:
@jchristgit thanks for these ideas. Unfortunately interactive rebase didn't seem to work as they were merge commits, and AIUI rerere only takes effect after it is enabled. I have made it into a single commit and force pushed, so all good now. (Made use of `git checkout --theirs <file>` which sped things up.)

Lint check is also fixed.

pydis_site/static/images/waves/wave_white.svg
GDWR, line 1: Oops lol, thats a hilarious commit

pydis_site/context_processors.py
```diff
@@ -1,7 +1,7 @@
 from django.conf import settings
-from django.template import RequestContext
+from django.http import HttpRequest
 
 
-def git_sha_processor(_: RequestContext) -> dict:
+def git_sha_processor(_: HttpRequest) -> dict:
```
hedyhli, line 5: Hi @GDWR, it's totally fine if you don't remember anymore, but would you happen to know why this change was required in your original PR?

pydis_site/static/images/waves/wave_black.svg
```diff
@@ -0,0 +1,77 @@
...
+         offset="100%"
+         id="stop4" />
+    </linearGradient>
+  </defs>
+  <path
+     fill="url(#a)"
+     fill-rule="evenodd"
+     d="M 1599.995,17.566918 C 1289,17.566918 1190.102,-0.03623696 789,5.6042811e-5 389,5.6042811e-5 289,17.566918 0,17.566918 v 11.178914 h 1600 c 0,0 -0.01,-6.968673 -0.01,-11.178914 z"
+     id="path9"
+     style="fill:#2C2F33;fill-opacity:1;stroke-width:0.381026" />
```
hedyhli: This file was copied as-is from `wave_white.svg`, only changing the fill color to `#2C2F33`, the primary site background color defined in `settings.py`.

pydis_site/static/css/home/index.css
```diff
@@ -107,6 +111,10 @@ h1 {
     margin: 0 1em;
 }
 
+[data-theme="dark"] #showcase .box {
+    box-shadow: 0 0.5em 1em -0.125em rgb(10 10 10 / 10%), 0 0px 0 1px rgb(10 10 10 / 2%);
```
hedyhli: This was copied as-is from bulma's own code, only changing the shadow colors to match those of the project cards.

hedyhli (commented):
Leaving some notes for (future) reviewers.

pydis_site/context_processors.py
GDWR: its been a while, but iirc the type was incorrect. And just thought it would be nice to correct it incase anybody ever tries to use the param.

hedyhli: Oh I see. Thanks!

hedyhli:
~~Most~~ All parts of the site that I visited have been tested to look acceptable in dark mode, I'm marking this ready for review.

pydis_site/static/images/resources/kivy.png
hedyhli, line 1: Elements in this image were placed in such a way so it looks the same as before (in light theme before this PR).

pydis_site/apps/resources/resources/automate_the_boring_stuff_book.yaml
hedyhli, line 1: I changed all "black" icons in resources to use dark so they can show up correctly in dark theme. For some reason I couldn't make black icons show as white without writing the CSS manually.

The tradeoff is that they're all blurple now due to being links.

pydis_site/apps/resources/resources/kivy.yaml
hedyhli, line 1: The existing kivy title image is dark and transparent making it hard to read in dark mode.

I've merged the icon and title image together into one image so it appears more consistent with other title images.

pydis_site/settings.py
hedyhli, line 1: These colors were adjusted with accessibility in mind.

django-simple-bulma does not support including the default variables into per-theme variables, so those non-color related variables have to be repeated for dark theme.

It also doesn't support reusing of default Bulma variables, so we have to define it ourselves before using it (in which case I've copied Bulma's colors for it).

pydis_site/static/css/home/index.css
```diff
@@ -218,21 +241,72 @@ h1 {
     text-align: center;
 }
 
-#sponsors .columns {
-    display: block;
-    justify-content: center;
-    margin: auto;
-    max-width: 80%;
+[data-theme="dark"] #sponsors {
+    background-color: #2C2F33 !important;
```
hedyhli: The sponsors section have been revamped so they can be displayed nicely readable in dark mode. This also happens to make their sizes more consistent for light mode (see #662).

Also see [relevant discord discussion](https://discord.com/channels/267624335836053506/635950537262759947/1193170173574774855).

pydis_site/templates/home/index.html
```diff
@@ -181,35 +181,39 @@ <h1 class="is-size-1">Projects</h1>
...
 
   <!-- Sponsors -->
-  <section id="sponsors" class="hero is-light">
+  <section id="sponsors" class="hero has-background-white-ter">
     <div class="hero-body">
       <div class="container">
         <h1 class="title is-6 has-text-grey">
           Sponsors
         </h1>
-        <div class="columns is-mobile is-multiline">
```
hedyhli, line 190: Read together with [this](https://github.com/python-discord/site/pull/1176#discussion_r1454365099) comment.

hedyhli (commented):
Some more notes.

wookie184 (changes requested):
This is really good! This review is just from me playing around with the site in firefox, on desktop. I'll try an have a check in chrome and on mobile when I get a chance, and have a better look at the code changes, but honestly i'm not sure i'll have much else to say since this seems great to me already!

-----
Regarding
> There is a latency in applying the current theme on page load, because this operation is only done after the page is loaded.

which causes a small flash on the page in some cases, I think we can safely set the style sheets before the page is loaded by moving this code out of the `DOMContentLoaded` listener which should prevent flashing.
```js
let theme;
const systemPrefersDark = window.matchMedia("(prefers-color-scheme: dark)");

if (systemPrefersDark.matches)
    theme = getCurrentTheme("dark");
else
    theme = getCurrentTheme();

setStyleSheets(theme);
```
As far as my HTML/browser understanding goes, this should work fine as long as the script tag is below the stylesheet tags it modifies, which it is currently.

-----
The titles for projects in light mode are now lighter than they were before. I think this makes them harder to read as the contrast is lower.

Before:
image

After:
image

----

The [panda3d logo on the resources page](https://deploy-preview-1176--pydis-static.netlify.app/resources/?search=panda3d) doesn't display on dark mode very well:

image

We could use a non transparent image, or only use the panda face icon and put the "panda3d" part as the title text. All the other ones seem good.

hedyhli:
@wookie184 
> I think we can safely set the style sheets before the page is loaded by moving this code out of the DOMContentLoaded listener which should prevent flashing.

This is a great idea, I've applied the change, thanks!

> The titles for projects in light mode are now lighter than they were before. I think this makes them harder to read as the contrast is lower.

This was actually the case for not just the project cards but almost all other `box` components throughout the site. I was initially planning to have it go unnoticed, however since you mentioned contrast which I hadn't considered, I've decided to fix all of them.

> The [panda3d logo on the resources page](https://deploy-preview-1176--pydis-static.netlify.app/resources/?search=panda3d) doesn't display on dark mode very well

Darn, that was a subtle one. I'm quite impressed you managed to find it! It's fixed now.

pydis_site/static/js/base/themes.js
```diff
@@ -0,0 +1,92 @@
...
+        // Let the new system preference take precedence over toggle preference
+        // on page reloads.
+        localStorage.removeItem("theme");
+        setStyleSheets(newTheme);
+        toggleThemeSwitch(newTheme);
+    })
+
+    const switchToggle = document.getElementById("theme-switch");
+    switchToggle.addEventListener("click", () => {
+        const newTheme = getCurrentTheme() === "light" ? "dark" : "light";
```
wookie184: If you have nothing set in localstorage and are on a system that prefers dark mode, the site will display in dark mode, but you have to click twice to change to light mode.

I believe this is because `getCurrentTheme` will return `light` here if there is nothing in local storage even if the current theme is dark mode because of system preference. Passing the `systemPrefers` argument here should fix it.

wookie184 (changes requested):
Tested this out on firefox/chrome on desktop and on firefox on my phone and all seems good. Just one minor thing I noticed then this should be good as far as i'm concerned.

pydis_site/static/js/base/themes.js
hedyhli: This is a good catch.

Your suggested solution does fix it. However after implementing it I decided that it made the logic and integration with `getCurrentTheme` it little complicated.

In the switch toggle our intention is to toggle the current theme. Since the `data-theme` attribute is set for all of system preference changes, switch clicks, and page loads, why not check for that instead? It makes the code cleaner and more explicit IMO. And it cannot, ever, contradict what the user actually sees the theme to be on the site.

The localStorage would then only be used when the page is first loaded to restore preference from last time.
The system preference would serve only two purposes: obtain correct theme if no localStorage preference stored, apply new theme if the system preference changes.

That's just my opinion. I can force-push any alternative implementations if you prefer.

Thanks for your thorough testing!

hedyhli (commented):
By the way, I am observing flash of unstyled content when I change the system preference and have the site automatically adapt to the new theme. It might have been caused by me fixing FOUC for switch toggle clicks, and/or FOUC on page loads.

I will probably not be fixing this for two reasons:
1. No idea how to fix for now
2. This feature is used much less often than the switch toggle, I only implemented it for completeness

If anyone has a fix though, I would gladly apply it.

pydis_site/static/js/base/themes.js
wookie184: Yeah, this is nicer, good stuff 👌

wookie184 (approved):
Perfect 🚀

lemonsaurus (approved):
Hi @hedyhli. This PR is absolutely _beautiful_ work, and it's maybe one of the best PR descriptions I've ever seen in my entire career of managing software developers. Bravo. 👏🏼 👏🏼 

The code is well documented, well implemented, and you've been extremely thorough. I have notes, but they are all cosmetic notes and everything in this PR is more or less perfectly implemented already.

# The colors
My main complaint is simply that I'm not a big fan of the colors. 

### Darker background
I think the background should be darker, and there should be more contrast between the background and the cards, for example, this is with the background set to `#1e2329` - note the contrast now between background and cards, I quite like it.
image

### Mismatched switcher purple
I don't really think this purple works well with the navbar purple:
image

### Navbar is too damn bright 😎 
Perhaps most of all, and this may be controversial, I think the navbar is far too bright when dark mode is enabled. I'd propose that we change the navbar color to be darker. For example, here it is using `#3b4774`:
image

### Darker discord component
I also think the discord component in the top right should change along with it. This would require altering the underlaying svg to maybe the same as the html background, and then changing the background-color of the <a> component to something like `#485580`:
image

# Other small stuff
I also have a few minor things that annoy me, these are not very important but I'll list them anyway.

### Sponsor logos
I get why we did this, but having a bunch of bright white boxes on the front page during dark mode irks me. Some of these logos do have dark-mode friendly alternatives available in their press kits and branding pages. For the others, I would suggest we just make some. Happy to help with this.
image

### Resource logos
Many of the resources have icky dark theme logos as well, this would also be great to fix by just creating a set of dark theme logos for each resource. 
image

### Timeline icons
Not a huge fan of the timeline icons having white backgrounds, as shown here:
image

Here's an example where the icon is darkened a bit and the circle around it is the same color as the line between the icons:
image

### Various buttons
There are a few buttons that also use the very bright navbar color, like the `"Check it out!"` button on the main page in the timeline widget. Here's an example where that one is darker:
image

# But should we merge this?
I mean, **maybe**. It's functional and these changes could all be in a separate PR, and would be easier to review and discuss if they were. It might be better for us to just merge this, migrate this review into a new issue, and tackle those changes separately. I certainly don't want to be the reason this PR goes stale.

I think you've done some tremendous work here, @hedyhli, and I'd love to see it in production very soon. On the other hand, if you are looking at these suggestions and thinking "oh okay that's easy, most of this will take me 10 minutes" and you agree with all my feedback, feel free to sneak a few of them in before we merge. It's your call. I'm approving this and you can merge if you'd like.

hedyhli:
Really appreciate this @lemonsaurus. ❤️

I agree on some of your points. I didn't have much confidence in the colors in this PR; partly due to it being my first time designing for dark mode and also due to the worry of the code being a bit too messy if I customize for dark mode too much (because of Bulma's limitations). However, I am glad you included suggested colors.

> I think the background should be darker, and there should be more contrast between the background and the cards, for example, this is with the background set to #1e2329 - note the contrast now between background and cards, I quite like it.

I spent quite some time adjusting the box shadows to try and make the boxes & cards get more contrast. Totally did not think of changing the background color instead! I will use the suggested background color.

---
> I don't really think this purple works well with the navbar purple

It looked fine to me. Is there a better color? Using the normal navbar background colors here makes the switch a bit muted and less noticeable. Maybe this could be discussed in another issue.

---
> Perhaps most of all, and this may be controversial, I think the navbar is far too bright when dark mode is enabled. I'd propose that we change the navbar color to be darker. For example, here it is using #3b4774

That hero in the screenshot looks spectacular. It absolutely feels like our ducky is browsing pydis at night. I think I'll also apply this suggestion.

---
> I also think the discord component in the top right should change along with it. This would require altering the underlaying svg to maybe the same as the html background, and then changing the background-color of the component to something like #485580

Yes, if the darker background is used, the current discord button is a bit too bright. However your proposed color seems to make the discord button not as prominent as it should be IMO. I'll see if I can find a good color for this one.

---
> Sponsor [and resources] logos

It's currently not very pretty and it can definitely be better. However, knowing how much time it took when I added those white backgrounds and given that a similar amount of time needs to be taken, I don't think I'll be fixing it in this PR.

Note to future readers: the issue is also applicable to the sponsor logos in our code jam event pages.

---
> Timeline icons

I did want to discuss this in `#dev-contrib` but the message got buried and I forgot about it, my bad! I'll probably just fix the background.

---
> Various buttons

I can update the button primary color to use the (darker) link color, it's fairly easy to change. But for an even better (and darker) button color, this could be left in a future issue.

---
**Summary**

Will fix before merge:
- [x] Background
- [x] Navbar & hero
- [x] (maybe) discord button
- [x] Timeline icon background (outline thing)
- [x] Slightly darker buttons

For another issue/PR:
- Switch button, light state background color
- [x] Homepage sponsors, resources, code jam sponsors logos (thanks to lemon for editing _all_ the images!)
- Timeline icon pastel colors
- ~A lot darker buttons~

hedyhli:
All done now!

I won't be accepting any more suggestions unless they are blatantly obvious and uncontroversial. Should be ready to merge.

Xithrius (approved):
Approval for good luck

:shipit:

PR #1181, opened 2023-12-15
Insert deleted messages as a Batch

shtlrs:
Sentry is sending an N+1 query [alert](https://python-discord.sentry.io/issues/4727001467/?referrer=slack) because of the way we were inserting messages (one by one)

### DIFF
--- a/pydis_site/apps/api/serializers.py
+++ b/pydis_site/apps/api/serializers.py
@@ -151,12 +151,9 @@ class MessageDeletionContextSerializer(ModelSerializer):
         """
         messages = validated_data.pop('deletedmessage_set')
         deletion_context = MessageDeletionContext.objects.create(**validated_data)
-        for message in messages:
-            DeletedMessage.objects.create(
-                deletion_context=deletion_context,
-                **message
-            )
-
+        DeletedMessage.objects.bulk_create(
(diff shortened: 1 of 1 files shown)


pydis_site/apps/api/serializers.py
```diff
@@ -151,12 +151,10 @@ def create(self, validated_data: dict) -> MessageDeletionContext:
...
-        for message in messages:
-            DeletedMessage.objects.create(
-                deletion_context=deletion_context,
-                **message
-            )
-
+        deleted_messages = [DeletedMessage(deletion_context=deletion_context, **message) for message in messages]
+        DeletedMessage.objects.bulk_create(
+            deleted_messages
+        )
```
jchristgit, line 156: Wow, who wrote the original, super inefficient code? Crazy.

I think it would be nice if: 1. `deleted_messages` was a generator (even though Django casts it to a list anyhow, if they ever implement some form of batching, we will benefit from it) and 2. this would be a single line.

shtlrs: Alright, fixed in the force push 2128dcf63088f10d8613a345ee4c028ec06df2c7

jchristgit (approved):

PR #1228, opened 2024-02-11
Resources config ready

### LINKED ISSUE #1209: Only read resource yaml files once on startup
Currently resources are loaded and parsed etc. in the `__init__` of the resources view here:
https://github.com/python-discord/site/blob/4ce61bddff9fd459b769305f302dcd9eb8ded395/pydis_site/apps/resources/views.py#L28-L36

I assume the intention here is for this to be run once on application startup, but in reality it seems this is run every time the `/resources` view is accessed. This leads to non-ideal performance. From sentry:
image

Cloudflare caching makes this issue less noticeable on the actual website, but it will still be an issue sometimes so is probably worth improving.

RohanJnr:
Closes #1209 

## Changes
Loading the resources file in the app config which runs only once per startup instead of doing it on very request send to the resources view.

### DIFF
--- a/pydis_site/apps/resources/apps.py
+++ b/pydis_site/apps/resources/apps.py
@@ -1,7 +1,99 @@
+from pathlib import Path
+
+import yaml
 from django.apps import AppConfig
+
+from pydis_site import settings
+from pydis_site.apps.resources.templatetags.to_kebabcase import to_kebabcase
+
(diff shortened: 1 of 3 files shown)


ChrisLovering (approved):

shtlrs (approved):
Quite nice, thank you !

PR #1232, opened 2024-02-13
Move from Font Awesome Kit to serving our own FA static files

### LINKED ISSUE #1213: Show font awesome icons in development
<details><summary> Old issue description</summary>

The site uses premium icons, but not all icons we use are premium.

Sometimes it is helpful for us to get a sense of how the icons look like, for free icons, when developing locally. This is especially so in PRs that involve colors or adding icons, such as #1210. My workaround is to change the `fontawesome_token` setting myself, but this is annoying when having to commit multiple changes together (especially when a change in settings.py is involved as well) -- to stage by-patch.

I propose that we add a `FONTAWESOME_TOKEN` to the envs, and then use that value in `settings.py` with the fallback being our actual font awesome token for production.

Files to be modified:
- `settings.py`
- Docs for contributing to site
  - Example `.env`
  - Instructions to how to create your kit on font awesome

---

</details>

**Problem**: We've surpassed the usage limits for the font awesome kit we've set in settings.py, and we need a way for icons to show up consistently in both production and local environments.

**Solutions**: Starting from lemon's first comment.

Further explanation on wookie's suggestion to use the font awesome django plugin:
https://discord.com/channels/267624335836053506/635950537262759947/1202436500856512583
> I meant <https://fontawesome.com/docs/web/use-with/python-django>, which would be a drop in replacement. It just bundles all the js/css/fonts into Django's static files so all that stuff would be served by us and we don't need to worry about kits/tokens/their cdn limits etc. (though the django plugin isn't really necessary for that, it could be better to just copy the files we need into the repo ourself)
> 
> Trying to manage each icon we want individually sounds like a bit more of a pain, the current way the icons are packaged using css and fonts is quite convenient, so I'm not sure if it would be worth adding complication just for a bit of a performance increase. I don't know exactly how that would work though so I can't really tell.

hedyhli:
Closes #1213.

Motivation:
- Show icons in development
- Drop dependency on lemon's fa kit
- Ensure icons are *always* shown in production (due to having surpassed kit views limits)
- More control over the static assets for font awesome

Notes:
- No significant changes in load time of the icons for netlify preview before and after this change
- [FontAwesome: Webfonts + CSS or SVG + JS?](https://fontawesome.com/docs/web/dig-deeper/webfont-vs-svg)


wookie184:
This seems to make collecting static files quite slow for me. I think this is because a bunch of stuff is being bundled that we don't need.

Before:
- `python manage.py collectstatic --no-input` with no staticfiles directory: ~3 seconds
- `python manage.py collectstatic --no-input` with existing staticfiles directory: ~3 seconds
- Size of staticfiles directory: 12MB from ~300 files

After:
- `python manage.py collectstatic --no-input` with no staticfiles directory: ~12 seconds
- `python manage.py collectstatic --no-input` with existing staticfiles directory: ~5 seconds
- Size of staticfiles directory: 55MB from ~9800 files

Because of this, starting site locally is pretty slow, especially since it seems we clear staticfiles whenever starting up site with `poetry run task start`/`python manage.py run --debug`:
https://github.com/python-discord/site/blob/88f2c27e8fd760f12e7b94854cf60c12047ea1d4/manage.py#L111-L116

I wonder if we'd be better off just copying the files we need into the repo instead of using the plugin, since I think we're only actually using about 5 of those 9500 files, and I don't see a way of configuring the plugin to only include what we want.

hedyhli:
Hmm, I hadn't noticed such a huge difference in startup for me. But having thousands of files that we don't need this way should probably be fixed - most of them are individual SVG files of the icons.

Looks like the only advantage of using this plugin as compared to copying files ourselves is to have automatic upgrades of Font Awesome through poetry/dependabot.

It would've been best if the plugin allows specifying which files to copy, unless there is a way to run scripts automatically every time poetry upgrades the dependency...?

So let's say we just copy the files ourselves without using the plugin. Looking at the [changelog](https://fontawesome.com/docs/changelog/), falling behind on versions probably won't affect us too much. Bumping the minor versions periodically doesn't seem too much of a maintenance burden either, but we'll have to remember to do it when needed. It's possible to achieve the same level of automation as dependabot when using the Font Awesome PyPI package, such as adding a script to the CI etc, though I guess keeping it up-to-date is not such a huge priority if it currently already "works".

For now I'm inclined to go for copying the five-ish fives to our own `static/` and include the license file there. Perhaps with a readme on instructions of how to upgrade when needed. What do we think?

wookie184:
Copying the files into the repo sounds good to me.

An alternative would be to use the the plugin but by extracting just the files we want manually rather than adding the plugin as an app, but I'm not sure exactly how that would work and still download/store lots of files from installing the module, so I think copying into the repo is probably the simpler and most efficient option.

hedyhli:
Thanks wookie.

I opted for `all.min.css` because I can't tell whether we would need the `regular` style as well in the future, and `all.css` would just include everything we are allowed to use for the free plan.

The latest FA release was added with only the required CSS/font files kept before force-pushing. I _think_ it's safe to just include everything in `webfonts` since it's just a few files and we'll never know when we might use icons that need them.

---

([Chris](https://discord.com/channels/267624335836053506/635950537262759947/1210548050406416404) triggered another netlify static build which then succeeded after the first one)

wookie184 (approved):
Looks good, thanks!

jchristgit (approved):
This is a good change. Thank you!

PR #1240, opened 2024-02-20
Ensure messages returned for metricity data do not match all excluded channels

ChrisLovering:
The logic here was wrong due to the `!=` requiring us to use negative logic. This meant that messages in excluded channels were still being used for message count.

Tests with my user:
```sql
SELECT
    COUNT(*)
FROM messages
WHERE
    author_id = '126811506632294400'
    AND NOT is_deleted
```
\>>> 111,045

```sql
SELECT
    COUNT(*)
FROM messages
WHERE
    author_id = '126811506632294400'
    AND NOT is_deleted
    AND channel_id != ANY('{267659945086812160,607247579608121354}')
```
\>>> 111,045

```sql
SELECT
    COUNT(*)
FROM messages
WHERE
    author_id = '126811506632294400'
    AND NOT is_deleted
    AND channel_id != ALL('{267659945086812160,607247579608121354}')
```
\>>> 110,402

### DIFF
--- a/pydis_site/apps/api/models/bot/metricity.py
+++ b/pydis_site/apps/api/models/bot/metricity.py
@@ -49,7 +49,7 @@ class Metricity:
             WHERE
                 author_id = '%s'
                 AND NOT is_deleted
-                AND channel_id != ANY(%s)
+                AND channel_id != ALL(%s)
             """,
             [user_id, EXCLUDE_CHANNELS]
         )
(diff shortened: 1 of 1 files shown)


wookie184 (approved):
Nice find

jchristgit (approved):
Sorry for the delay here, good catch!

PR #1255, opened 2024-03-11
Fix legacy Font Awesome icon classes

hedyhli:
We were using some Font Awesome 4 classes, eg:
- https://fontawesome.com/v4/icon/snowflake-o
- https://fontawesome.com/v4/icon/github

Which have since been removed. They don't show up correctly after switching to Font Awesome 6 (in #1232).

This is because the Kit we were previously using included some JS files that handled the backwards compatibility. (Thanks to @lemonsaurus for finding this out)

This fix replaces these font awesome classes to the most sensible replacements I could find, which are the new class names and should all display correctly.

### Timeline

https://pythondiscord.com/timeline/
https://deploy-preview-1255--pydis-static.netlify.app/timeline/

1

image
image

2

image
image

### Resources 

https://pythondiscord.com/resources/?type=project-ideas
https://deploy-preview-1255--pydis-static.netlify.app/resources/?type=project-ideas

image
image

### Code Jam results

https://www.pythondiscord.com/events/code-jams/
https://deploy-preview-1255--pydis-static.netlify.app/events/code-jams/

For CJ10, 9, 8.

image
image

### DIFF
--- a/pydis_site/apps/resources/templatetags/get_category_icon.py
+++ b/pydis_site/apps/resources/templatetags/get_category_icon.py
@@ -20,7 +20,7 @@ _ICONS = {
     "Other": "fa-question-circle",
     "Paid": "fa-dollar-sign",
     "Podcast": "fa-microphone-alt",
-    "Project Ideas": "fa-lightbulb-o",
+    "Project Ideas": "fa-lightbulb",
     "Security": "fa-solid fa-lock",
     "Software Design": "fa-paint-brush",
     "Subscription": "fa-credit-card",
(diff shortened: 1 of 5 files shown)


hedyhli:
I noticed there are a few more [FA icons](https://www.pythondiscord.com/events/code-jams/10/) on our site that needs fixing, from looking at a [deploy preview](https://deploy-preview-1210--pydis-static.netlify.app/events/code-jams/10/) from before #1232 the issue might be from upgrading FA, from which incompatible classes we might've missed.

I think I'll just fix those as well in this PR together since I found them already.

jchristgit (approved):
It's kind of annoying that the "lazy load" animation on the timeline doesn't work properly if you try to search for anything. Oh well. Thank you for the fixes.

hedyhli:
There's a one-liner fix for that which just changes hidden items from `visibility: hidden` to `opacity: 0`. Further testing might be needed though.

Xithrius:
I'm bad at requesting reviews, apologies

ChrisLovering (approved):

PR #1261, opened 2024-03-19
Timeline: Fix Liechtenstein postcard image

hedyhli:
The discord link which we've previously used stopped working for me, I've switched it to the raw github link from our branding repo instead.


ChrisLovering (approved):
Looks good! With Discord's new security and expirations for CDN links, we should look at changing all CDN links we use to something else.

PR #1266, opened 2024-03-22
Fix broken Bulma logo in the footer

hedyhli:
It looks like Bulma updated their links to their brand assets.


Xithrius (approved):
brilliant

wookie184 (approved):
👍

PR #1268, opened 2024-03-22
Use proper logos and better positioning in the footers

### LINKED ISSUE #458: Django logo is not aligned with baseline of text in the footer
bild

Here is a close up

bild

bild

The Bulma logo is also a bit off, but it's close enough to not be visible at small sizes I think (I couldn't even notice until I zoomed in to take the screenshots above).

### LINKED ISSUE #662: Improve logos on the landing page.

hedyhli:
Closes #662, and as part of it, also closes #458.

Relevant `#dev-contrib` discussion: https://discord.com/channels/267624335836053506/635950537262759947/1220627130644434996

### Changes

Footer logos and positioning:
- Switch from CSS filtering to actual logos with white text.
- Adjust positioning to align all text in the logos with baseline and ensure they're center aligned with a uniform side spacing.


wookie184 (commented):
The text alignment still isn't perfect for me (firefox on windows), is it the same for you?
Before:
image
After:
image

The other changes seem good though.

hedyhli:
It should be alright in both windows and android now.

On iOS, it's pretty misaligned vertically if I zoom in, but it's probably somewhat subtle otherwise.

Since this isn't such a major issue (I've made sure no part of the image is cut off on all platforms) we'll just make sure it looks good for the majority of our users for now, as lemon suggested in the linked discussion. Considering that this is web dev, I wonder if a cross-platform solution even exists.

[Descenders](https://en.wikipedia.org/wiki/Descender) in Django and Netcup which is what caused the issue can't really be fixed without hardcoding the vertical offsets, but I've cropped the linode icon and switched to the text-only Bulma logo (both of which has no descenders) so it will show up aligned on all platforms.

wookie184 (approved):
It looks really good for me now
image

I agree with that it's not massively important so whatever balance you think seems best between different platforms is good with me.

hedyhli:
Of course, wookie, I hacked into your computer to test it! ;P

But yes, since they're fairly small logos in the footer I don't think the offsets on iOS are _that_ noticeable. We could see if anyone complains.

jchristgit (approved):
Thanks!

PR #1276, opened 2024-03-29
Support running tests with --keepdb option

jchristgit:
Allow faster local tests by skipping the database re-setup. Only two cases were dependent on hardcoded IDs and already existing data, they have been adjusted appropriately.


wookie184 (approved):
Looks good 👍

Xithrius (approved):

PR #1277, opened 2024-03-31
Work around DRF 3.15 regression

jchristgit:
Manually define the fields on the FilterSerializer due to a bug in DRF 3.15 reported upstream at
https://github.com/encode/django-rest-framework/issues/9345. Once that issue is fixed, we can remove the custom field code again. A new test case is also added to ensure that we don't run into this issue again.


jchristgit:
This doesn't work as expected. Let's roll back the version until an upstream fix is available.

PR #1280, opened 2024-03-31
Set elidable on RunPython operations in migrations

jchristgit:
Any RunPython operation that only migrated existing data (that is, did not create data) can be safely elided by Django when we run `squashmigrations`.


ChrisLovering (approved):

wookie184 (approved):
lgtm

PR #1281, opened 2024-04-01
Don't propagate the status received from Discord to GitHub Webhook

jb3:
When we received a bad request exception from Discord (which is not actually representative of an error here), we were propagating this back to GitHub.

Instead, we should always return a success request if we have been successful in the proxy attempt, and return the downstream status code as a part of the response body.

This prevents GitHub Webhook Bad Requests being logged to site logs.

### DIFF
--- a/pydis_site/apps/api/views.py
+++ b/pydis_site/apps/api/views.py
@@ -305,7 +305,14 @@ class GitHubWebhookFilterView(APIView):
         )
         headers.pop('Connection', None)
         headers.pop('Content-Length', None)
-        return Response(data=body, headers=headers, status=response_status)
+
+        response_body = {
+            "original_status": response_status,
+            "data": body.decode("utf-8"),
+            "headers": headers,
+        }
+
+        return Response(response_body)
 
     def send_webhook(
         self,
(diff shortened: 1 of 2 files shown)


pydis_site/apps/api/views.py
```diff
@@ -305,7 +305,14 @@ def post(self, request: Request, *, webhook_id: str, webhook_token: str) -> Resp
         )
         headers.pop('Connection', None)
         headers.pop('Content-Length', None)
```
MarkKoz: We don't need to do this anymore, right?

pydis_site/apps/api/views.py
```diff
@@ -305,7 +305,14 @@ def post(self, request: Request, *, webhook_id: str, webhook_token: str) -> Resp
...
-        return Response(data=body, headers=headers, status=response_status)
+
+        response_body = {
+            "original_status": response_status,
+            "data": body.decode("utf-8"),
```
MarkKoz: How does GH know it has to look in this nested structure now instead of the root?

MarkKoz (approved):

pydis_site/apps/api/views.py
jb3: Good spot, removed in 346d570

pydis_site/apps/api/views.py
jb3: It doesn't -- but it doesn't need to, GitHub stores all response details for us to browse, it doesn't do any processing of this response. I'm going to add a sanity log on our side though to mark when we get a 400 from downstream (but without logging sensitive info).

jchristgit (approved):
great

PR #1286, opened 2024-04-01
Prefetch the filter_list field in the FilterListSerializer

jb3:
We were experiencing an issue in production where the `api:bot:filterlist-list` view was taking several seconds to process, upon enabling query logs and monitoring for this we found it was making a query for every filter in the database.

Prefetching the `filter_list` field solves this and greatly reduces the query count, optimising the route.


ChrisLovering (approved):

jchristgit (approved):
great

PR #1287, opened 2024-04-01
Upsert tags in two queries

jchristgit:
Before, we had an N+1 query, where each tag would be separately fetched and updated. Instead, first delete any tags we no longer care about, then perform an upsert of all fields except for `last_commit`, which we want to keep from the existing tag.

This comes with one caveat: on the first update of the tags, the `tags` list is no longer modified in-place, because mutation is bad, however, the database state is updated accordingly.


ChrisLovering (approved):

jb3 (approved):
Great!

PR #1290, opened 2024-04-02
Replace `docker-compose` with `docker compose` in action

wookie184:
Assuming it's just a drop in change

Done because our [lint workflow is failing](https://github.com/python-discord/site/actions/runs/8527879039/job/23360258497) since `docker-compose` was removed from the image we're using https://github.com/actions/runner-images/issues/9557


jchristgit (approved):
Thanks!

Xithrius (approved):
yep

PR #1321, opened 2024-05-19
Update status embed workflow

wookie184:
Removes redundant and incorrect condition, and improves consistency with other repos

### DIFF
--- a/.github/workflows/status-embed.yaml
+++ b/.github/workflows/status-embed.yaml
@@ -7,24 +7,12 @@ on:
     types:
       - completed
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   status_embed:
-    # We need to send a status embed whenever the workflow
-    # sequence we're running terminates. There are a number
-    # of situations in which that happens:
-    #
-    # 1. We reach the end of the Deploy workflow, without
-    #    it being skipped.
-    #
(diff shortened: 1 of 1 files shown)


Xithrius (approved):

.github/workflows/status-embed.yaml
```diff
@@ -7,24 +7,12 @@ on:
     types:
       - completed
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
MarkKoz, line 13: What does this do and why?

wookie184: It means that if a status-embed workflow is triggered while one is already running it will kill the old one before running the new one. So if you quickly but separately push multiple commits in sequence it will only post the status of the newest CI run*.

I mainly added for consistency as most other repos had it already.

*now that I think about it, it's a bit more complicated than this, since the CI workflow takes longer than the status_embed workflow. If CI is replaced with a new run, the old one's status embed (saying it was cancelled) will finish before the new CI finishes, so both status embeds will be sent. However if the CI is rerun again, then it might cancel the old status embed and send the two newest... The behaviour is somewhat questionable but I don't think it's too bad...

MarkKoz: Thanks. If we already use it successfully in other repos then I'm not concerned with keeping it.

MarkKoz (approved):

PR #1337, opened 2024-06-07
Add a management command to close all connections

shtlrs:
We are experiencing SSL errors in site which we think are due to the abrupt shut down of the django app.

A potential solution is to close all the connections before shutting the server down.

This will work in conjunction with [this infra PR](https://github.com/python-discord/infra/pull/350)

### DIFF
--- a/pydis_site/apps/api/management/commands/close_db_connections.py
+++ b/pydis_site/apps/api/management/commands/close_db_connections.py
@@ -1,0 +1,19 @@
+from django.core.management.base import BaseCommand
+from django.db import connections
+import logging
+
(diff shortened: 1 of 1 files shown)


jchristgit (changes requested):
This doesn't, and cannot work, see my comment in infra.

I think we should close this. Sorry.

shtlrs:
Yeah i noticed that once i had tested it,  and was planning on changing the approach once you look at my suggestion in [devoops](https://discord.com/channels/267624335836053506/675756741417369640/1248605523981500506)

PR #1338, opened 2024-06-07
Update production deployment configuration

jb3:
- Change `manage.py` to only be used for debug running of the site
- Update Dockerfile to use the `gunicorn` command to run site in production
- Update docker-compose for new local debug running

Hopefully, this will rectify some odd database issues we've been having related
to other application and database code being loaded before the first request is
hit, leading to errors on first request served by site after boot.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,8 @@ RUN if [ $STATIC_BUILD = "TRUE" ] ; \
   then SECRET_KEY=dummy_value poetry run python manage.py distill-local build --traceback --force ; \
 fi
 
-# Run web server through custom manager
-ENTRYPOINT ["poetry", "run", "python", "manage.py"]
-CMD ["run"]
+ENTRYPOINT ["poetry", "run"]
+CMD ["gunicorn", "--preload", "-b", "0.0.0.0:8000", \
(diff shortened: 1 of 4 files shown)


jchristgit (approved):

jb3:
I will write and update the site with an init container in the deployment manifest at python-discord/infra once we merge this PR.

Dockerfile
```diff
@@ -39,6 +39,7 @@ RUN if [ $STATIC_BUILD = "TRUE" ] ; \
   then SECRET_KEY=dummy_value poetry run python manage.py distill-local build --traceback --force ; \
 fi
 
-# Run web server through custom manager
-ENTRYPOINT ["poetry", "run", "python", "manage.py"]
-CMD ["run"]
+CMD ["poetry", "run", "gunicorn", "--preload", "-b", "0.0.0.0:8000", \
```
MarkKoz: Would it be more convenient to have `poetry run` be the entrypoint?

MarkKoz (approved):

Dockerfile
jb3: That's a good call, updated in force push.

MarkKoz (approved):

PR #1373, opened 2024-08-02
Use kebab-case for keys of `valid_filters` in resources

hedyhli:
The YAML side uses snake case, whereas the HTML template and JavaScript use kebab case. This difference affects only `Payment tiers` right now.

Currently if the page has a payment tier filter active, reloading the page would lose the filter. A simple fix would be to use kebab case in `valid_filters` which will be used in JavaScript.

Demo:
https://deploy-preview-1373--pydis-static.netlify.app/resources/?payment-tiers=free&difficulty=beginner
https://www.pythondiscord.com/resources/?payment-tiers=free&difficulty=beginner


jchristgit (approved):
Thanks!

PR #1374, opened 2024-08-02
Fix image for People, Postgres, Data resource

hedyhli:
The discord link no longer works, I've pulled the image for light mode from: https://postgresconf.org/conferences and updated the image for dark mode to be similar to what it looked like before this change.


jchristgit (approved):
Thanks!

PR #1437, opened 2024-11-03
Fix typo in `git remote add` in the Sir Lancebot guide

RundownRhino:
`git add remote` isn't valid, and most likely was a typo.


ChrisLovering (approved):
Nice catch, thanks

PR #1444, opened 2024-12-03
Heroku is no longer free

kkrypt0nn:
Heroku [no longer](https://help.heroku.com/RSBRUH58/removal-of-heroku-free-product-plans-faq) offers a [free plan](https://www.heroku.com/pricing).

### DIFF
--- a/pydis_site/apps/content/resources/guides/python-guides/vps-services.md
+++ b/pydis_site/apps/content/resources/guides/python-guides/vps-services.md
@@ -42,17 +42,3 @@ While these may seem like nice and free services, it has a lot more caveats than
 
 - They use a shared IP for everything running on the service.
 This one is important - if someone is running a user bot on their service and gets banned, everyone on that IP will be banned. Including you.
-
-### Heroku
-- Bots are not what the platform is designed for. Heroku is designed to provide web servers (like Django, Flask, etc). This is why they give you a domain name and open a port on their local emulator.
-
-- Heroku's environment is heavily containerized, making it significantly underpowered for a standard use case.
-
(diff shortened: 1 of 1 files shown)


shenanigansd (approved):
LGTM
Though all these criticisms are still valid on the paid tiers, aren't they? As a possible alternative, we could change the header from "why free hosts are bad" to "why shared hosts are bad", and leave the content intact.

kkrypt0nn:
"Shared hosts" is maybe not the right term, any VPS is technically shared as long as you don't have a dedicated server.

Though open to the idea to merge from replit & heroku everything under a subtitle

jchristgit:
I'm a bit 50/50 on this, because as said above, these criticisms still stand.
I'm not sure we should merge this as is.

kkrypt0nn:
I do believe that no points of Heroku are valid for a title about shared hosts - just the last point of Replit. Could move the last point of Replit to a subtitle like "Why shared IP hosts are bad". It would just contain a single point though. (Maybe can be written as a small note?)

Alternatively remove the entire "Replit" subtitle, and just put in the bullet points for free hosts, listing Replit as an example. The bullet points would contain the ones of Heroku as well. Though those bullet points won't apply for *all* free hosting solutions.

wookie184 (approved):
I think this change is fine as-is. If someone wanted to rewrite the section to just include a list of common issues instead of doing it per service, or to add in more current free services, I think that would be fine, but that seems separate from the issue this PR is fixing.

kkrypt0nn:
> I think this change is fine as-is. If someone wanted to rewrite the section to just include a list of common issues instead of doing it per service, or to add in more current free services, I think that would be fine, but that seems separate from the issue this PR is fixing.

I definitely could do that if wished, also renaming the PR

wookie184:
Sure, sounds good.

jchristgit (approved):

PR #1478, opened 2025-02-25
Clarify embed characters limit

kkrypt0nn:
The guide didn't mention how many embeds are allowed per message nor that the total characters is shared across all embeds of the message.

Also removed the `disputils` library from pagination options as it's not supported for discord.py 2.x -> https://github.com/LiBa001/disputils/blob/master/setup.py#L21


jchristgit (approved):
Thank you for the PR!

PR #1495, opened 2025-03-25
Remove sololearn from resources

L3viathan:


swfarnsworth (approved):

jchristgit:
Thank you for the PR. It would be nice to have the context of these change in the commit somehow in the future. I figured it's based on the messages in #community-meta, but those might be hard to find in the future when somebody might ask why we don't link to it.

PR #1532, opened 2025-06-02
Update Timeline (400k members, #coveragepy, 3.13 release stream)

L3viathan:
It's been a while since the last update of the timeline, I've added three more items:

- The 3.13 release stream
- Adding the `#coveragepy` channel
- Hitting the 400,000 member mark


ChrisLovering (approved):
Looks good!

{280D7E03-E635-4EC6-96FE-4D2059D4A773}

jb3 (approved):

PR #1547, opened 2025-07-13
Resources Categories Update

janine9vn:
As a part of going through our current open issues for resources, I've identified a new category we could benefit from (Conference Talks) and also renaming another one (Project Ideas --> Practice).

I adjusted the sole resources that was using the category that was renamed as well.


jb3 (approved):

jchristgit (approved):
Thanks!

PR #1556, opened 2025-07-22
Initial CJ12 update

janine9vn:
- upload banner and logo
- new cj12 pages
- update home page


pydis_site/templates/events/pages/code-jams/12/rules.html
```diff
@@ -0,0 +1,85 @@
...
+    <li>
+        <p>
+            <strong>The core of your project must focus on and incorporate "Python in the Browser".</strong>
+        </p>
+    </li>
+    <li><p>Your solution should be platform agnostic. For example, if you use filepaths in your submission, use <code>pathlib</code> to create platform agnostic Path objects instead of hardcoding the paths.</p></li>
+    <li>
+        <p>
+            You must document precisely how to install and run your project.
+            This should be as easy as possible, which means you should consider using dependency managers like <code>pipenv</code> or <code>poetry</code>.
```
jb3:
```suggestion
            This should be as easy as possible, which means you should consider using dependency managers like <code>uv</code> or <code>poetry</code>.
```
With our recent move to `uv` and our existing use of `poetry`, there aren't any situations where I'd recommend pipenv now. For us as reviewers, this is going to make things so much easier when initialising environments.

jb3 (approved):
Minor one, otherwise all looks good to me on the static preview.

PR #1562, opened 2025-09-02
Address some issues from the accessibility milestone

### LINKED ISSUE #1381: No alt text for images on events page
As pointed out in #1363:

Many images on the events page lack a definition for alt text.

All images need alt text defined so that screen readers can present the information to the user. If an image is not necessary for a screen reader user, it should define alt="".

### LINKED ISSUE #1382: Non-descriptive alt text on thumbnail on home page
As pointed out in #1363:

It is not clear for screen reader users that the welcome video thumbnail is a playable video.

The alt text should inform the user that the content is a video, such as "Welcome video, click to play".

### LINKED ISSUE #1406: Alt text on decorative images on timeline is useless
As pointed out in #1363:

Python logos and other icons in the center line of the timeline are read as “logo image” and other generic names.

While descriptive text is imperative to describe images to those who can’t see them, these images are purely decorative and add no value to someone not seeing the page. Use alt=”” instead so the experience with a screen reader is less cluttered.

DragonSenseiGuy:
Change log:
- Added alt-text for the Welcome Video in the homepage
- Added alt-texts of "" for images that does not affect people with screen-reader in the events page as well as the timeline page
- Changes alt-texts in the events page form "AoC" to "Advent of Code" so, people who don't have knowledge about it can still know what it is.
- Change date in the 3.13 release from `2022` to `2024`

Fixes
- #1382
- #1381
- #1406

### DIFF
--- a/pydis_site/templates/events/scheduled_events.html
+++ b/pydis_site/templates/events/scheduled_events.html
@@ -3,7 +3,7 @@
     <article class="media">
         <div class="media-left">
             <p class="image is-64x64">
-                <a href="{% url "events:page" path="code-jams/12" %}"><img class="is-rounded" src="/static/images/events/summer_code_jam_2025/logo.png" alt=""></a>
+                <a href="{% url "events:page" path="code-jams/12" %}"><img class="is-rounded" src="/static/images/events/summer_code_jam_2025/logo.png" alt="Summer Code Jam 2025"></a>
             </p>
         </div>
         <div class="media-content">
@@ -26,7 +26,7 @@
     <article class="media">
         <div class="media-left">
             <p class="image is-64x64">
-                <a href="{% url "events:page" path="pyweek" %}"><img class="is-rounded" src="/static/images/events/icons/pyweek.png" alt=""></a>
+                <a href="{% url "events:page" path="pyweek" %}"><img class="is-rounded" src="/static/images/events/icons/pyweek.png" alt="PyWeek 38"></a>
             </p>
         </div>
         <div class="media-content">
@@ -49,7 +49,7 @@
     <article class="media">
         <div class="media-left">
             <p class="image is-64x64">
-                <a href="{% url "events:page" path="adventofcode" %}"><img class="is-rounded" src="/static/images/events/icons/adventofcode.png" alt=""></a>
+                <a href="{% url "events:page" path="adventofcode" %}"><img class="is-rounded" src="/static/images/events/icons/adventofcode.png" alt="Advent of Code"></a>
             </p>
         </div>
         <div class="media-content">
--- a/pydis_site/templates/events/scheduled_events.html
+++ b/pydis_site/templates/events/scheduled_events.html
@@ -3,7 +3,7 @@
     <article class="media">
         <div class="media-left">
             <p class="image is-64x64">
-                <a href="{% url "events:page" path="code-jams/12" %}"><img class="is-rounded" src="/static/images/events/summer_code_jam_2025/logo.png"></a>
+                <a href="{% url "events:page" path="code-jams/12" %}"><img class="is-rounded" src="/static/images/events/summer_code_jam_2025/logo.png" alt=""></a>
             </p>
         </div>
         <div class="media-content">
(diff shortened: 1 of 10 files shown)


jchristgit:
Thank you for the PR! FYI it'll take me a while to review this, I'm a bit swamped with work right now :-)

DragonSenseiGuy:
> Thank you for the PR! FYI it'll take me a while to review this, I'm a bit swamped with work right now :-)

NP take your time

MarkKoz (approved):

pydis_site/templates/events/scheduled_events.html
```diff
@@ -3,7 +3,7 @@ <h2 class="title is-5" style="margin-bottom: 2rem;">Scheduled events</h2>
     <article class="media">
         <div class="media-left">
             <p class="image is-64x64">
-                <a href="{% url "events:page" path="code-jams/12" %}"><img class="is-rounded" src="/static/images/events/summer_code_jam_2025/logo.png"></a>
+                <a href="{% url "events:page" path="code-jams/12" %}"><img class="is-rounded" src="/static/images/events/summer_code_jam_2025/logo.png" alt=""></a>
```
L3viathan: Why do these three have `alt=""`? As proper links they should be clickable and with empty alt text it will be ignored by screen readers, right?

DragonSenseiGuy: https://github.com/python-discord/site/issues/1406#issue-2446336451

>"While descriptive text is imperative to describe images to those who can’t see them, these images are purely decorative and add no value to someone not seeing the page. Use alt=”” instead so the experience with a screen reader is less cluttered."

jb3: These images are not decorative. If they are the primary (and in this case sole) component within an `<a>` tag then they are by definition not decorative, they are serving a purpose as a hyperlink to the code jam pages and as such must have appropriate alt text.

DragonSenseiGuy: > These images are not decorative. If they are the primary (and in this case sole) component within an `<a>` tag then they are by definition not decorative, they are serving a purpose as a hyperlink to the code jam pages and as such must have appropriate alt text.

I'm just following what the issue says, but I can add it if needed.

jb3: > I'm just following what the issue says, but I can add it if needed.

Yes, the issue generally covers the situations where that *is* a problem, but this is not one of them.

Hypothetically for the following content:

```html
<div id="main">
  <!-- where this is a decorative graphic that holds no useful user information -->
  <img src="/some/graphic.png" />
</div>
```

The image there is decorative, is not clickable, serves no purpose and thus is not worth having alt text for, adding alt text here would just be noise for a screen reader. So, we'd add `alt=""`:

```html
<div id="main">
  <!-- where this is a decorative graphic that holds no useful user information -->
  <img src="/some/graphic.png" alt="" />
</div>
```

However, if the `` is within an `<a>` then that image is clickable, and if there is nothing else in that `<a>` then there is nothing else to indicate to a user _what_ they are clicking.

For example, basing things off your current PR, we have something like the following:

```html
<a href="/login">
  <img src="/login_graphic.png" alt="" />
</a>
```

There is no descriptive text inside the `<a>`, so a screenreader cannot look at this and say anything to the user along the lines of "This is a login button", it has no idea how to describe the content because we have explicitly disabled the alt text.

It's very important to keep the descriptive text where the object is not decorative, obviously an image that is acting as a button/link is not decorative since clicking it takes you somewhere.

Break down your elements into three categories: informational, decorative or actions.

- You can have buttons that are actions, they take you somewhere when you click them. These obviously must be eithered labelled with text elements or be an image element with `alt` text.
- You can have images that may contain information or may add information to content surrounding them (even if accessibility tools cannot read the image content), these also obviously need `alt` text to symbolise to a screenreader/user using accessibility tooling what the content of that image is.
- Decorative objects are objects where you could remove them from the document and the content and actions from the document would remain the same, you can gain the same information from the document and use the same actions to continue browsing to other areas of the webpage.

Decorative objects do not need accessibility text because it is not important that tools to help with visual impairment describe elements of a webpage that have no use to users who are visually impaired (for example, some fancy SVG background doesn't need explaining to a user who has no way to see the background).

DragonSenseiGuy: On it, probably will have a new commit in like 1-2 days.

DragonSenseiGuy: Fixed.

L3viathan (approved):

jb3 (approved):

DragonSenseiGuy:
@jb3 Thanks for the merge!

PR #1566, opened 2025-10-21
fix: update bot's required version and contrib deadlinks

onerandomusername:
Planned to merge these with a larger update to the contributing documentation, but wanted to get these out of the way so they don't get held up by my time constraints.

### DIFF
--- a/pydis_site/apps/content/resources/guides/pydis-guides/contributing/bot.md
+++ b/pydis_site/apps/content/resources/guides/pydis-guides/contributing/bot.md
@@ -15,7 +15,7 @@ This page will focus on the quickest steps one can take, with mentions of altern
 ### Setup Project Dependencies
 Below are the dependencies you **must** have installed to get started with the bot.
 
-1. Make sure you have [Python 3.13](https://www.python.org/downloads/) installed. It helps if it is your system's default Python version.
+1. Make sure you have [Python 3.13](https://www.python.org/downloads/) installed. uv [can also be used to](https://docs.astral.sh/uv/guides/install-python/#installing-python) install Python, if desired.
 2. [Install uv](https://github.com/astral-sh/uv#installation).
 3. [Install the project's dependencies](../installing-project-dependencies).
 4. Docker.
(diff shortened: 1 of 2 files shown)


pydis_site/apps/content/resources/guides/pydis-guides/contributing/bot.md
```diff
@@ -15,7 +15,7 @@ This page will focus on the quickest steps one can take, with mentions of altern
 ### Setup Project Dependencies
 Below are the dependencies you **must** have installed to get started with the bot.
 
-1. Make sure you have [Python 3.12](https://www.python.org/downloads/) installed. It helps if it is your system's default Python version.
+1. Make sure you have [Python 3.13](https://www.python.org/downloads/) installed. It helps if it is your system's default Python version.
```
wookie184: IIRC `uv` installs the Python version you need for you so this step probably isn't necessary any more I guess?

onerandomusername: It may be useful to keep somewhat documented, as uv installed from a distro-packaged uv may result in a uv with automatic downloads to be disabled without the user knowing.

Reference: https://github.com/astral-sh/uv/issues/15625
https://src.fedoraproject.org/rpms/uv/blob/rawhide/f/uv.toml

onerandomusername: This also matches the existing contributing guides for sir-lancebot and site which don't mention that uv can install python. That feels out of scope for this pull.

jb3: I'd keep it but probably remove the note about it being system Python. UV is good at discovering it even if it's non-system (for example on macOS I use Homebrew versions of Python, but UV figures that out).

onerandomusername: [`80ee7f3` (#1566)](https://github.com/python-discord/site/pull/1566/commits/80ee7f3a356eb39f39d7c0f2c9e9326f36ef76af)

wookie184 (approved):
Thanks

jchristgit (approved):

jchristgit:
Thanks!

PR #1596, opened 2026-03-05
if __name__ == '__main__':

nedbat:
A page for the new shorter embed to link to. The shorter embed is https://github.com/python-discord/bot/pull/3475


L3viathan:
:+1:

graingert:
Should probably link to https://docs.python.org/3/library/__main__.html#idiomatic-usage

nedbat:
That's a good idea. I was looking for other resources and didn't find this one.

PR #1620, opened 2026-06-12
Fix status embed to use single file artifact

wookie184:
Was broken by 7f343b28d98c0307286c63872664a747751518a1 switching to a non-zip artifact without updating the usage.


MarkKoz (approved):

jchristgit (approved):
Thanks, BROTHER! I did not test it, I will not, because I TRUST IN YOUR ABILITY!

PR #1633, opened 2026-07-06
Fix n+1 queries

ChrisLovering:


pydis_site/apps/api/serializers.py
```diff
@@ -705,6 +714,18 @@ class Meta:
         depth = 1
         list_serializer_class = UserListSerializer
 
+    def validate_roles(self, roles: list[int]) -> list[int]:
```
MarkKoz, line 717: Is this covered by tests?

ChrisLovering: The existing user test will cover this, even tough it's moved from model to serialiser.
https://github.com/python-discord/site/blob/main/pydis_site/apps/api/tests/test_users.py#L110-L122

However, I have noticed it doens't assert the response detail, nor do we have coverage for PATCH (single, nor bulk), so have added all of that in https://github.com/python-discord/site/pull/1633/commits/f7b8d0c4f191d24658c9f2fece3c4eaa17889d09

MarkKoz (approved):

PR #1634, opened 2026-07-09
Fix n+1 query issue in message deleteion context

ChrisLovering:
Declare `id` and `author` as plain integers to drop DRF's per-row UniqueValidator and FK lookup, and validate all authors and IDs with one batched query each. Query count is now constant instead of 2 per message.

### DIFF
--- a/pydis_site/apps/api/serializers.py
+++ b/pydis_site/apps/api/serializers.py
@@ -111,9 +111,10 @@ class DeletedMessageSerializer(ModelSerializer):
     model for more information.
     """
 
-    author = PrimaryKeyRelatedField(
-        queryset=User.objects.all()
-    )
+    # Plain integers to avoid per-row DB validators (id's UniqueValidator, author's
+    # FK lookup) that cause N+1 queries. Validation is done in the `MessageDeletionContextSerializer` instead.
+    id = IntegerField()
+    author = IntegerField(source='author_id', allow_null=True)
     deletion_context = PrimaryKeyRelatedField(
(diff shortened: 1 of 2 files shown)


jb3:
deferring to @jchristgit for a review 🧌

jchristgit (changes requested):
> Declare uid=502(chris) gid=20(staff) groups=20(staff),12(everyone),61(localaccounts),702(com.apple.sharepoint.group.2),100(_lpoperator),701(com.apple.sharepoint.group.1) and as plain integers to drop DRF's per-row UniqueValidator and FK lookup, and validate all authors and IDs with one batched query each. Query count is now constant instead of 2 per message.

I can't help but feel that this isn't a good solution, but it also appears that there is no better solution.

There is a ticket for DRF at https://github.com/encode/django-rest-framework/issues/9607. I suppose this is a nice demonstration of where it falls short.

I guess n+1s are problematic from a performance perspective, but we're not exactly chasing top of the shelf response time here - does this solve an actual problem or just some noise in Sentry? Because if not, I'm inclined to say we leave it as is, because this code would be annoying to maintain. I would rather think about how to fix the issue in DRF or perhaps even how to stop using it altogether, but it's not like anyone has time for that anyways.

jchristgit:
FWIW, I also tried various solutions with `select_related`, but those don't have any effect, because we're trying to create something new here - those would probably help on the GET endpoints though.

jb3:
After deferring to my Vice-Best Man, I am in agreement with the conclusions he has drawn. THANKS!

PR #1638, opened 2026-07-25
Fix more n plus 1 queries

ChrisLovering:


jchristgit (changes requested):
The select related change in `apps/staff/views.py` looks good to me.

The cache functionality does not seem like a good idea to me, because the invalidation isn't bulletproof. If multiple HTTP workers cache the roles, and one of them creates or deletes a new role, then only that HTTP worker will have its cache cleared, the rest will keep running around with a stale cache.

The filter list test also seems unrelated. I do not mind, but why was this added?

ChrisLovering:
> The select related change in `apps/staff/views.py` looks good to me.
> 
> The cache functionality does not seem like a good idea to me, because the invalidation isn't bulletproof. If multiple HTTP workers cache the roles, and one of them creates or deletes a new role, then only that HTTP worker will have its cache cleared, the rest will keep running around with a stale cache.
> 
> The filter list test also seems unrelated. I do not mind, but why was this added?

jc, thank you for the thorough review.

I have dropped the role cache, as it breaks the bullet proof nature of oiur site deployment. I am sorry.

I have also moved the filter list test to the first commit, since it's related to the code change in there, to retain 100% code coverage.

jchristgit (approved):
LGTM, but note the queryset probably won't do anything on POST, right?

